PR #6, opened 2017-02-22
Add some exception handling around the config files.

blueandgold:
This helps to gracefully handle errors when the config files are not created nor have proper values.


blueandgold:

PR #8, opened 2017-02-23
Add some error handling around the config files.

blueandgold:
This is to gracefully handle if the config files are not created or set properly.


blueandgold:

PR #9, opened 2017-02-23
Add some exception handlings around the config files.

blueandgold:
This is to make sure we gracefully handle if the config files are not created or set properly.


blueandgold:

PR #10, opened 2017-02-23
Fixing issues with the inventory deployment templates

mirons-google:
- Applying the property of database-name on the right template
- Properly applying dependsOn in the template, not the YAML


blueandgold:

PR #17, opened 2017-02-24
Implement emailing service.

blueandgold:
I debated whether to send an email on fatal exceptions.  I leaned toward not doing it, since it's easy to tell something went wrong if no email for successful run is received.  Let me know if we prefer to explicitly receive email for failures.

Please focus the review on the mechanics & structure.  Any issues with email formatting & language is extra-scope as I will need guidance what they should be.

Left some TODOs around how to handle secrets, and setting of the config values (e.g. email recipient & sender) that I will follow-up.  This also includes updating the README.


blueandgold:

PR #18, opened 2017-02-24
Use deployment templates and commandline flags

carise:
Currently spins up all the instances and does the initial install. Runs the CLI tools.


blueandgold:

PR #19, opened 2017-02-24
Single quote

mirons-google:
Fixing the deployment templates to stick to single quotes.


blueandgold:

PR #20, opened 2017-02-27
Move the retry to a common util.

blueandgold:
There is a couple of ways to do this, but I'm not sure what would work
best.  Either we can group all the exceptions together, or split up the
exceptions by smaller groups.  Such as email_exceptions vs http
exceptions.  This PR does the one big grouping as that seems easier to
manage.  But will be happy to hear of any feedback for the alternative.


blueandgold:

PR #21, opened 2017-02-27
Remove a TODO that is already resolved.

blueandgold:


blueandgold:

PR #22, opened 2017-02-28
Removing the need for replication

mirons-google:
It resolves the issue with deployment templates. We can always put this back once the issue is resolved.


blueandgold:

PR #24, opened 2017-03-01
Update run script to use full path of forseti_*

carise:


blueandgold:

PR #27, opened 2017-03-02
Scan organization IAM policies

carise:
Scan organization IAM policies. Also, I tweaked the email util so that we don't send an email if there is no sender/recipient provided.


blueandgold:

PR #30, opened 2017-03-02
Update docstrings after move to 4 space standard.

ahoying:
Migrate retry on mismatched firewall policy code to project_enforcer.


blueandgold:

PR #35, opened 2017-03-03
Update readme

mirons-google:
This simplifies the landing page experience for new users and clearly delinates the two paths a user can take to installing and using Forseti Security.


blueandgold:

PR #42, opened 2017-03-03
Inventory: don't halt pipeline runner if there are errors

carise:
Also, send an email after the inventory loader is done.

Some DM template changes for scope.


google/cloud/security/inventory/inventory_loader.py
```diff
@@ -219,21 +219,37 @@ def main(unused_argv=None):
...
+        try:
+            pipeline['pipeline'].run(
+                dao, cycle_timestamp, configs, crm_rate_limiter)
+            pipeline['status'] = 'SUCCESS'
+        except LoadDataPipelineError as e:
+            LOGGER.error(
+                'Encountered error to load data.\n{0}'.format(e))
+            pipeline['status'] = 'FAILURE'
+
+    succeeded = filter(lambda p: p['status'] == 'SUCCESS', pipelines)
```
ahoying: I think something like this would be more readable:

succeeded = [p['status'] == 'SUCCESS' for p in pipelines]

if all(succeeded):
    snapshot_cycle_status == 'SUCCESS'
elif any(succeeded):
    snapshot_cycle_status == 'PARTIAL_SUCCESS'
else:
    snapshot_cycle_status == 'FAILURE'

blueandgold: +1 for not using the lambda function :)

carise:
PTAL: https://github.com/GoogleCloudPlatform/forseti-security/pull/42/files#diff-f3010219d1f4657ac39c0159e3fac4d9R241

blueandgold:

PR #46, opened 2017-03-04
Update README.md

mirons-google:
Minor cleanup


blueandgold:

PR #47, opened 2017-03-05
Change makefile to use argparse instead of gflags

carise:
When running the makefile.py in subprocess during the setup.py installation, the makefile can't find gflags or apputils and fails the entire setup. I tried to include the apputils and gflags .eggs in the PYTHONPATH env variable and passed it to subprocess.Popen(), which worked. However, the rest of the setup.py installation didn't include apputils or gflags in the installed packages. As a result, running any of the forseti_* or even the scripts directly could not find gflags or apputils. (pip list installed also didn't show them.)

Given the time constraint we have, I felt that it would make sense just to use argparse since 1) argparse is in the standard library and 2) the makefile is essentially just a script and doesn't really do anything fancy with the flags or apputils. We can revisit this after next week if it will be an issue.

### DIFF
--- a/cleanup.sh
+++ b/cleanup.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+rm -rf build dist .eggs *.egg-info
+find . -type f -name "*pb2*" -delete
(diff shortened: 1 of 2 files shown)


blueandgold:

PR #49, opened 2017-03-06
Improved local installation README

mirons-google:
- Create a TOC
- Clearer heading levels for key parts


blueandgold:

PR #83, opened 2017-03-06
Update README.md -- fix the role name

carise:
Fix the role name


blueandgold:

PR #87, opened 2017-03-07
Update README.md -- instructions about cloudsql instance name

carise:


deployment-templates/README.md
```diff
@@ -73,14 +73,16 @@ $ gcloud organizations add-iam-policy-binding ORGANIZATION_ID \
 The provided DM templates are samples for you to use. Make a copy of `deploy-forseti.yaml.sample` as `deploy-forseti.yaml` and update the following variables:
 
 * CLOUDSQL_INSTANCE_NAME
+  * Instance name must be lowercase letters, numbers, and hyphens; must start with a letter. See [naming guidelines](https://cloud.google.com/sql/docs/mysql/instance-settings#settings-2ndgen) for more information.
```
blueandgold: nit: Can we explicitly call out no underscore?  Something like this below, which will also help us to remember in the future?
"For example, underscore is not allowed in the instance name."

carise: I think the instructions are very clear that the characters allowed are [a-z0-9-], which excludes underscores.

blueandgold:

PR #114, opened 2017-03-11
Travis CI Integration

mirons-google:
A first pass at an implementation of Travis integration.
- This adds a helper script to download protoc and install it correctly.
- This adds a badge to the main README

#14

### DIFF
--- a/scripts/travis_install.sh
+++ b/scripts/travis_install.sh
@@ -17,11 +17,12 @@
 
 set -e
 
-if [ ! -d "/tmp/protoc" ]; then
-  wget https://github.com/google/protobuf/releases/download/v3.2.0/protoc-3.2.0-linux-x86_64.zip
-  unzip -d /tmp/protoc protoc-3.2.0-linux-x86_64.zip
-  sudo mv /tmp/protoc/bin/protoc $PROTOC_PATH
-  sudo chmod 755 $PROTOC
+if [ ! -d $PROTOC_DOWNLOAD_PATH ]; then
+  wget -P /tmp \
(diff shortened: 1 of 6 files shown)


scripts/travis_install.sh
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+
+# A script to perform the download and installation of protobuf.
+
+set -e
+
+echo "Downloading protoc."
+mkdir -p $PROTOC_DOWNLOAD_PATH
+cd $PROTOC_DOWNLOAD_PATH
+wget https://github.com/google/protobuf/releases/download/v3.2.0/protoc-3.2.0-linux-x86_64.zip
```
blueandgold: nit: Please feel free to pass on this, or do it later.  My nit is that we have 2 places where we are specifying which protoc to download.  So, we might update one, but forget to update here.  So, is there anyway that we can specify the protoc once, and both this and forseti-instance.py would be updated?  Again, not critical or blocking.

https://github.com/GoogleCloudPlatform/forseti-security/blob/91621f3109d3339b1da9a1eb08e7fdbe36ecc692/deployment-templates/py/forseti-instance.py#L121

blueandgold: At the least, can we put a comment near L121 in forseti-instance.py to say, "if you change the protoc version, please also update in travis_install.sh"?

mirons-google: Good point. One way might be to check-in a file to the repo containing the version/URL that these two scripts fetch.

carise: Yeah, that would be a good idea -- a cfg file that lists the required third party binaries/urls

mirons-google: I added #115

blueandgold:

PR #117, opened 2017-03-13
Updating pylint and implementing it into Travis

mirons-google:
This PR does quite a bit
- Either disables or fixes pylint (1.6.5) changes to the code-base
- Improvements/changes to pylintrc
- Updates to .travis.yaml along with a new travis script and updates to the previous one

One special thing to note is that for pylint messages I couldn't easily resolve or didn't make sense I added disable's and thus TODOs to come back and revisit them.

Resolves:
#114, #75


blueandgold:
Hmmm, I feel that this is a case where it would be beneficial to separate out the changes (if possible), i.e. one PR for the pylint/travis integration, and another PR for the changes to the app code.  I am just worried that this is a pretty extensive change on the app code.  So in case the app gets borked and we have to do a rollback, the pylint change won't be rolled back as well.

Or does doing the pylint change here, forces us to also make the update the app code?

mirons-google:
Yes, essentially by checking in the pylint test we need the code to pass at the same time, otherwise the build breaks. So the code changes are are limited to pylint required changes.

mirons-google:
@carise I've confirmed the tests pass (check travis) and I've successfully deployed and the tool ran successfully on the hour.

google/cloud/security/common/data_access/dao.py
```diff
@@ -42,6 +38,7 @@
     'org_iam_policies': create_tables.CREATE_ORG_IAM_POLICIES_TABLE,
     'projects': create_tables.CREATE_PROJECT_TABLE,
     'project_iam_policies': create_tables.CREATE_PROJECT_IAM_POLICIES_TABLE,
+    # pylint: disable=line-too-long
```
ahoying: Can you add TODOs on the common pylint errors that you don't correct? They should be fixed in a followup PR.

google/cloud/security/common/data_access/organization_dao.py
```diff
@@ -59,8 +58,8 @@ def get_org_iam_policies(self, timestamp):
                     org = Organization(organization_id=row[0])
                     iam_policy = json.loads(row[1])
                     org_iam_policies[org] = iam_policy
-                except ValueError as json_error:
-                    self.logging.warn('Error parsing json {}'.format(row[2]))
+                except ValueError as _:
```
ahoying: You can just say 'except ValueError' without assigning a variable if you don't use it.

google/cloud/security/enforcer/enforcer.py
```diff
@@ -120,21 +119,23 @@ def enforce_single_project(enforcer, project_id, policy_filename):
...
+        FLAGS.concurrent_threads, FLAGS.maximum_project_writer_threads,
+        FLAGS.maximum_firewall_write_operations, FLAGS.dry_run)
 
     if FLAGS.enforce_project and FLAGS.policy_file:
         enforcer_results = enforce_single_project(enforcer,
                                                   FLAGS.enforce_project,
                                                   FLAGS.policy_file)
 
-        print(enforcer_results)
+        print enforcer_results
```
ahoying: We should be adding:

from __future__ import absolute_import
from __future__ import division
from __future__ import print_function

to our imports and using print() to make the code easier to move to python3 support.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -198,13 +197,13 @@ def _send_email(cycle_timestamp, status, sendgrid_api_key,
     except EmailSendError:
         LOGGER.error('Unable to send email that inventory snapshot completed.')
 
-def main(unused_argv=None):
+def main():
```
ahoying: Better to accept argv and then delete it when not used, for compatibility with apputils and gflags.

def main(argv):
    """..."""
    del argv

google/cloud/security/scanner/scanner.py
```diff
@@ -333,12 +336,12 @@ def _build_scan_summary(all_violations, total_resources):
...
         # Keep track of # of violations per resource id.
         if (violation.resource_id not in
-            resource_summaries[resource_type]['violations']):
+                resource_summaries[resource_type]['violations']):
             resource_summaries[resource_type][
                 'violations'][violation.resource_id] = 0
 
         # Make sure to count each member violation as a separate one.
-        for member in violation.members:
+        for _ in violation.members:
```
ahoying: Unrelated to this change, but this should be refactored to remove the loop.

         members = len(violation.members)    
         resource_summaries[resource_type][
              'violations'][violation.resource_id] += members
  
         total_violations += members

pylintrc
```diff
@@ -216,7 +217,7 @@ single-line-if-stmt=no
 no-space-check=trailing-comma,dict-separator
 
 # Maximum number of lines in a module
-max-module-lines=1000
+max-module-lines=1100
```
ahoying: This seems an arbitrary increase. We should reason about this and some of the other changes rather than just forcing the changes to minimize our alerts.

google/cloud/security/common/data_access/organization_dao.py
mirons-google: Done!

pylintrc
mirons-google: It is. Most of the code that broke this rule is the enforcement code. I can open an issue to track shrinking it's size.

google/cloud/security/scanner/scanner.py
mirons-google: Ack.

google/cloud/security/enforcer/enforcer.py
mirons-google: Ack.

google/cloud/security/common/data_access/dao.py
mirons-google: done.

pylintrc
mirons-google: Created #119

mirons-google:
Addressed/done. PTAL @ahoying.

mirons-google:
Pinging with a comment to see if this re-starts the webhook request for CLA

blueandgold:

PR #125, opened 2017-03-14
Try to split up some functions and remove the unnecessary for loop

carise:


google/cloud/security/scanner/scanner.py
```diff
@@ -340,13 +352,9 @@ def _build_scan_summary(all_violations, total_resources):
...
                 'violations'][violation.resource_id] = 0
 
-        # Make sure to count each member violation as a separate one.
-        # TODO: Refactor as the loop is not required.
-        for _ in violation.members:
-            resource_summaries[resource_type][
-                'violations'][violation.resource_id] += 1
-
-            total_violations += 1
+        resource_summaries[resource_type][
```
mirons-google: Not looking, but do we have a test to make sure these counts come out equal? If not it might be good to add one.

google/cloud/security/scanner/scanner.py
```diff
@@ -240,15 +224,43 @@ def _write_violations_output(logger, violations):
                 'member': '{}:{}'.format(member.type, member.name)
             }
 
-def _upload_csv_to_gcs(logger, output_path, output_filename, csv_name):
+def _output_results(logger, all_violations, **kwargs):
+    """Send the output results.
+
+    Args:
+        logger: The logger.
```
mirons-google: Do you need to pass the logging instance? Shouldn't it already be available to the method?

carise: I chose not to make it global. Currently working on extending the log utility to look for commandline flags and things get really ugly (because of gflags) when flags are not parsed before accessing them -- that's what would happen if I had the logger.setup_logging() called on the global scope.

google/cloud/security/scanner/scanner.py
carise: no, there are currently no unit tests for the runner script.

blueandgold:

PR #129, opened 2017-03-20
Don't fail pipeline on an error with IAM policy.

### LINKED ISSUE #128: Better Error Handling for Snapshot Tables
A project (or any resource) in a problematic state should not bork the entire snapshot table.  Handle any per-resource errors so the inventory pipeline can continue.

This might involve a schema change to flag 'iam_binding_error' on projects with IAM errors, so we can track them.

blueandgold:
Previously, we had a stance where we failed the entire pipeline if there was even a single error with an IAM policy.  We want to change this, so that if there is problem with the IAM policy of one project, we flag it but keep going with getting the IAM policies for other projects.

For this PR, we are flagging by logging.  There will be some more work to do, if we want to flag it in the database.

Fixes #128.

### DIFF
--- a/google/cloud/security/common/gcp_api/cloud_resource_manager.py
+++ b/google/cloud/security/common/gcp_api/cloud_resource_manager.py
@@ -114,16 +114,17 @@ class CloudResourceManagerClient(_BaseClient):
         """
         projects_stub = self.service.projects()
 
-        try:
-            for project_number in project_numbers:
+        for project_number in project_numbers:
+            try:
                 with self.rate_limiter:
                     request = projects_stub.getIamPolicy(
                         resource=project_number, body={})
                     response = self._execute(request)
                     yield {'project_number': project_number,
                            'iam_policy': response}
-        except (HttpError, HttpLib2Error) as e:
-            raise ApiExecutionError(resource_name, e)
+            except (HttpError, HttpLib2Error) as e:
+                LOGGER.error('Unable to get IAM policies for project %s:\n%s',
+                             project_number, e)
 
     def get_organization(self, org_name):
         """Get organizations.
--- a/google/cloud/security/common/gcp_api/cloud_resource_manager.py
+++ b/google/cloud/security/common/gcp_api/cloud_resource_manager.py
@@ -117,7 +117,7 @@ class CloudResourceManagerClient(_BaseClient):
                     resource=project_number, body={})
                 response = self._execute(request)
                 return {'project_number': project_number,
-                       'iam_policy': response}
+                        'iam_policy': response}
         except (HttpError, HttpLib2Error) as e:
             raise ApiExecutionError(resource_name, e)
 
--- a/google/cloud/security/common/gcp_api/cloud_resource_manager.py
+++ b/google/cloud/security/common/gcp_api/cloud_resource_manager.py
@@ -97,31 +97,29 @@ class CloudResourceManagerClient(_BaseClient):
         except (HttpError, HttpLib2Error) as e:
             raise ApiExecutionError(resource_name, e)
 
-    def get_project_iam_policies(self, resource_name, project_numbers):
+    def get_project_iam_policies(self, resource_name, project_number):
         """Get all the iam policies of given project numbers.
 
         Args:
             resource_name: String of the resource's name.
-            project_numbers: MySQLdb cursor object of project numbers.
+            project_number: Integer of project number.
 
-        Yields:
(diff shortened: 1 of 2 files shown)


google/cloud/security/common/gcp_api/cloud_resource_manager.py
```diff
@@ -114,16 +114,17 @@ def get_project_iam_policies(self, resource_name, project_numbers):
         """
         projects_stub = self.service.projects()
 
-        try:
-            for project_number in project_numbers:
+        for project_number in project_numbers:
```
ahoying: The logic to loop through project numbers should be handled in the calling function, not here. This function should just try to get the IAM policy for a single project, either by project_id or project_number. Either works when used as the resource in getIamPolicy, so make sure the docstring and arg name is generic. The function can just return the response then if there are no errors, no need to have it be a generator or create a dictionary for the response.

This function should still raise APIExecutionError on exception, and then the calling function can decide what to do when an exception is raised. (Log an error, set a flag in the database record, etc).

Otherwise if when you mask the exception inside of this loop, the caller has no way of knowing it, or you need to create a new key in your response dict, which is messy and not very Pythonic.

google/cloud/security/inventory/pipelines/load_projects_iam_policies_pipeline.py
```diff
@@ -53,21 +53,18 @@ def run(dao=None, cycle_timestamp=None, configs=None, crm_rate_limiter=None):
...
-        # pylint: disable=redefined-variable-type
-        iam_policies_map = list(iam_policies_map)
+    # Retrieve data from GCP.
+    # Flatten the iterator since we will use it twice, and it is faster
+    # than cloning to 2 iterators.
+    iam_policies_map = crm_client.get_project_iam_policies(
+        RESOURCE_NAME, project_numbers)
+    # TODO: Investigate improving so the pylint disable isn't needed.
+    # pylint: disable=redefined-variable-type
+    iam_policies_map = list(iam_policies_map)
```
ahoying: Be sure to fix this pylint error in this PR.

blueandgold: Done.

google/cloud/security/common/gcp_api/cloud_resource_manager.py
blueandgold: Very good points.  I have addressed the core of this comment by moving the loop to the caller, and retain the APIExecutionError.

The only item that's tricky to do, is to genericize the incoming project_number parameter.  Because the project_number is used specifically as the key in the project_iam_policies database table.
https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/data_access/sql_queries/create_tables.py#L38
How do you think?

ahoying: I think this works fine for now. My comment on accepting either project_number or project_id was more to encourage code reuse if some other bit of code wanted to call get_project_iam_policies and only had the project_id, for example. Like if a future enforcer update wants to check the current policy before enforcing changes on it.

blueandgold: Got it, I made the parameter generic now.  This makes the api cleaner.  It will just be a bit more work on the caller to handle the output.

google/cloud/security/inventory/pipelines/load_projects_iam_policies_pipeline.py
```diff
@@ -53,21 +56,24 @@ def run(dao=None, cycle_timestamp=None, configs=None, crm_rate_limiter=None):
...
-        raise LoadDataPipelineError(e)
+
+    # Retrieve data from GCP.
+    # Not using iterator since we will use the iam_policy_maps twice.
+    iam_policy_maps = []
+    for project_number in project_numbers:
+        try:
+            iam_policy = crm_client.get_project_iam_policies(
+                RESOURCE_NAME, project_number)
+            iam_policy_map = {'project_number': project_number,
```
ahoying: This looks like it would be better as a namedtuple, then the keys/attributes are explicitly defined ahead of time, which is cleaner than a dict with potentially variable keys, since any new keys have to be defined in the namedtuple definition before use: https://docs.python.org/2/library/collections.html#collections.namedtuple. Can be done in a future CL.

blueandgold: Ack.

blueandgold:

PR #130, opened 2017-03-20
Some instructions and tweaks for using user-defined release for Forseti

carise:
- Rename the variable to "release-version"
- Some instructions on how to configure

### DIFF
--- a/deployment-templates/deploy-forseti.yaml.sample
+++ b/deployment-templates/deploy-forseti.yaml.sample
@@ -44,7 +44,6 @@ resources:
 - name: forseti-instance
   type: forseti-instance.py
   properties:
-    forseti-version: "1.0"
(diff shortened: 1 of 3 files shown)


deployment-templates/deploy-forseti.yaml.sample
```diff
@@ -55,7 +54,8 @@ resources:
     scanner-bucket: SCANNER_BUCKET
     database-name: forseti_security
     organization-id: YOUR_ORG_ID
-    src-path: https://github.com/GoogleCloudPlatform/forseti-security/archive/v1.0.tar.gz
+    release-version: "master"
```
blueandgold: nit: I would rather think that the default should point at a stable release version, with getting the bleeding edge "master" being an option.

blueandgold:

PR #131, opened 2017-03-21
Improve the inventory loading email.

### LINKED ISSUE #106: Improve email notification for inventory.
Right now, we have a skeleton for inventory email notification.  For improvements:

(1) We should add more data to this, such as listing each resources that is queued for loading, and what is the final status (success, failed, not loaded, etc).

(2) Use email template.

blueandgold:
Make the inventory loading email more interesting with details of what pipelines have run, how many resources have been processed, and the status.

Example.
image

Fixes #106

### DIFF
--- a/google/cloud/security/common/data_access/dao.py
+++ b/google/cloud/security/common/data_access/dao.py
@@ -97,6 +97,30 @@ class Dao(_DbConnector):
                 OperationalError, ProgrammingError) as e:
             raise MySQLError(resource_name, e)
 
+    def select_record_count(self, resource_name, timestamp):
+        """Select the record count from a snapshot table.
+
+        Args:
+            resource_name: String of the resource name, which is embedded in
+                the table name.
+            timestamp: String of timestamp, formatted as YYYYMMDDTHHMMSSZ.
+
+        Returns:
+             Integer of the record count in a snapshot table.
+
+        Raises:
+            MySQLError: An error with MySQL has occurred.
+        """
+        try:
+            record_count_sql = select_data.RECORD_COUNT.format(
+                resource_name, timestamp)
+            cursor = self.conn.cursor()
+            cursor.execute(record_count_sql)
+            return cursor.fetchone()[0]
+        except (DataError, IntegrityError, InternalError, NotSupportedError,
+                OperationalError, ProgrammingError) as e:
+            raise MySQLError(resource_name, e)
+
     def select_project_numbers(self, resource_name, timestamp):
         """Select the project numbers from a snapshot table.
 
--- a/google/cloud/security/common/email_templates/inventory_snapshot_summary.jinja
+++ b/google/cloud/security/common/email_templates/inventory_snapshot_summary.jinja
@@ -16,14 +16,10 @@ limitations under the License.
 
 <!doctype html>
 <html>
-
-<head>
-</head>
-
 <body>
     Organization ID: {{ organization_id }}<br>
-    Timestamp: {{ cycle_timestamp }}<br>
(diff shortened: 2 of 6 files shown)


google/cloud/security/common/data_access/dao.py
```diff
@@ -97,6 +97,30 @@ def load_data(self, resource_name, timestamp, data):
...
+             Integer of the record count in a snapshot table.
+
+        Raises:
+            MySQLError: An error with MySQL has occurred.
+        """
+        try:
+            record_count_sql = select_data.RECORD_COUNT.format(
+                resource_name, timestamp)
+            cursor = self.conn.cursor()
+            cursor.execute(record_count_sql)
```
carise: It's suggested to pass the parameters as a tuple to execute instead of using .format().

http://stackoverflow.com/questions/775296/python-mysql-parameterized-queries

google/cloud/security/common/email_templates/inventory_snapshot_summary.jinja
```diff
@@ -0,0 +1,46 @@
...
+
+<!doctype html>
+<html>
+
+<head>
+</head>
+
+<body>
+    Organization ID: {{ organization_id }}<br>
+    Timestamp: {{ cycle_timestamp }}<br>
```
carise: Just a question -- Would it make sense to include a human readable timestamp as well?

google/cloud/security/common/email_templates/inventory_snapshot_summary.jinja
```diff
@@ -0,0 +1,46 @@
...
+    Status Summary: {{ status_summary}}<br>
```
carise: space between status_summary and }}

google/cloud/security/common/email_templates/inventory_snapshot_summary.jinja
```diff
@@ -0,0 +1,46 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<!doctype html>
+<html>
+
+<head>
```
carise: Is it possible to remove the <head> element?

google/cloud/security/common/data_access/dao.py
blueandgold: Great catch.  I tried it, however not working.  When I look up the error, it seems that parameterized query only works for column value, and not anything else.
http://stackoverflow.com/a/12831519/2830207

google/cloud/security/common/email_templates/inventory_snapshot_summary.jinja
blueandgold: Done.

google/cloud/security/common/email_templates/inventory_snapshot_summary.jinja
blueandgold: Done.

google/cloud/security/common/email_templates/inventory_snapshot_summary.jinja
blueandgold: Done.

carise:
You're right. I forgot that format() must be used for column names.

google/cloud/security/common/data_access/dao.py
carise: You're right. I forgot that format() must be used for column names.

blueandgold:

PR #132, opened 2017-03-21
Organization and consolidation of our docs

mirons-google:
This helps to resolve #97

PR #133, opened 2017-03-21
Add documentataion about sendgrid tracking pixel.

blueandgold:
This fixes this problem in gmail.
image


blueandgold:
This is incorporated in #132

blueandgold:

Issue #102, opened 2017-03-09
Handle exception if the firewall json is not properly formatted.


blueandgold:
Address the problem if the json file contains a single rule, either raise an exception early or properly format it as a single item list transparently.

mirons-google:
@blueandgold can you provide an example of a problematic rule?

carise:
This is because the enforcer is expecting a list of rules and bombs out when you pass it a json object.

blueandgold:
Yup, if the firewall rules are not provided as a list, as in this example.
https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/enforcer/samples/default_allow_policy.json

So, if someone just have a single rule, they still need to put into a list.  i.e. they can not just have this.

```
    {
        "sourceRanges": ["0.0.0.0/0"],
        "description": "Allow ICMP from anywhere",
        "allowed": [
            {
                "IPProtocol": "icmp"
            }
        ],
        "name": "allow-icmp"
    }
```

PR #138, opened 2017-03-22
Adding checks for ensuring properly parsed enforcer policies (including tests)

mirons-google:
Resolves issue #102

### DIFF
--- a/google/cloud/security/enforcer/gce_firewall_enforcer.py
+++ b/google/cloud/security/enforcer/gce_firewall_enforcer.py
@@ -486,9 +486,6 @@ class FirewallRules(object):
           InvalidFirewallRuleError: One or more rules failed validation.
         """
         for rule in rules:
-            if not isinstance(rule, dict):
-                raise InvalidFirewallRuleError(
-                    'Invalid rule type. Found %s expected %s', type(rule), dict)
             self.add_rule(rule, network_name=network_name)
 
     def add_rule(self, rule, network_name=None):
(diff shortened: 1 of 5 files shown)


google/cloud/security/enforcer/gce_firewall_enforcer.py
```diff
@@ -486,6 +486,9 @@ def add_rules(self, rules, network_name=None):
           InvalidFirewallRuleError: One or more rules failed validation.
         """
         for rule in rules:
```
blueandgold: I believe the check should be before this for loop, i.e. make sure we are getting a list of rules.

ahoying: Actually I think the check should go in add_rule not here.

mirons-google: If I put this in add_rule I'd still check this on line one, so why not check here? This would prevent entering the function unnecessarily.

blueandgold: This is the correct check to make sure it is a valid dictionary.  So, LGTM, and addressing andrew's comment.

ahoying: Ahh, I missed that this was to check for an invalid policy.

Okay, so two things. One, add_rule is what cares that rule is a dict, so the isinstance check belongs there, in case some other code path is calling add_rule directly. The add_rules methods cares that rules is an iterable of rule dicts, so slightly different flavor, and different check.

It also makes sense to verify that the loaded json policy in enforcer.py is in the expected format and return a meaningful error to the user at that point, rather than waiting for it to error out while building the rules object.

mirons-google:
Done and I've added tests. PTAL.

blueandgold:

PR #139, opened 2017-03-22
Fix rule inheritance

### LINKED ISSUE #136: Rules engine does not properly respect "applies_to"
When there are multiple definitions for the same resource (e.g. org) but different "applies_to", the Rules engine only respects the first applies_to that it sees.

e.g.

org 1234, applies_to=SELF, whitelist, allow roles/owner, user@company.com
org 1234, applies_to=CHILDREN, whitelist, allow roles/* *@company.com

This will only look at org 1234 "SELF".

Reported by @lukwam

carise:
resolves #136 

For an example rule:
1. org 1234, applies_to=SELF, whitelist, allow roles/owner, user@company.com
2. org 1234, applies_to=CHILDREN, whitelist, allow roles/* *@company.com
3. org 1234, applies_to=SELF_AND_CHILDREN, blacklist, don't allow roles/* *@anothercompany.com

This fixes the bug where the scan ignores (i.e. doesn't take into consideration) the rules for org1234 after the first rule. In other words, in the above example, rule 2 and 3 get dropped.


google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -58,17 +58,17 @@ def find_policy_violations(self, resource, policy, force_rebuild=False):
...
         Returns:
-            A list of tuples containing the policy diffs.
+            A generator of rule violations.
         """
         if self.rule_book is None or force_rebuild:
             self.build_rule_book()
 
         violations = []
-
+        # pylint: disable=redefined-variable-type
```
mirons-google: Is there a way to do this w/o having to add a pylint disable here?

mirons-google:
Given what's in #136 can you detail the new result here?

google/cloud/security/scanner/audit/org_rules_engine.py
carise: I'll see if I can

carise:
@mirons-google I've updated the PR description to explain what the fix does.

mirons-google:
@carise thanks! Does this also warrant some documentation changes too?

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -459,10 +480,8 @@ def _dispatch_rule_mode_check(self, mode, rule_members=None,
             rule_members=rule_members,
             policy_members=policy_members)
 
-    # TODO: Investigate making a function, not a method.
-    # pylint: disable=no-self-use
-    def _whitelist_member_check(self, rule_members=None,
-                                policy_members=None):
+    @staticmethod
```
ahoying: I'm not sure the best solution is protected static methods, it looks like trying to use a class as a namespace, which is not really the intent.

Since they can be called and tested independent of their class, I would move these out of the class(es) they are defined in, to top level protected functions. If you want them in a different namespace for readability, then I'd move them into their own class that only contains static methods, or a new .py module, but a better solution is to make sure they are named appropriately (which I think they already are).

google/cloud/security/scanner/audit/org_rules_engine.py
ahoying: You can initially set violations to an empty itertools.chain() object.

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -58,17 +58,17 @@ def find_policy_violations(self, resource, policy, force_rebuild=False):
...
-
+        # pylint: disable=redefined-variable-type
         for binding in policy.get('bindings', []):
-            binding_violations = (
+            violations = itertools.chain(
+                violations,
                 self.rule_book.find_violations(resource, binding))
-            violations.extend(binding_violations)
 
         return violations
```
ahoying: To make this back into a list I would return list(itertools.chain.from_iterable(violations)).

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -269,22 +269,45 @@ def add_rule(self, rule_def, rule_index):
...
+        """Get all the resource rules for (resource, RuleAppliesTo.*).
+
+        Args:
+            resource: The resource to find in the ResourceRules map.
+
+        Returns:
+            A list of ResourceRules.
+        """
+        resource_rules = []
+        resource_rules.append(self.resource_rules_map.get(
```
ahoying: Maybe

for rule_appies_to in RuleAppliesTo.apply_types:
  if (resource, rule_applies_to) in self.resource_rules_map:
    resource_rules.append(self.resource_rules_map[(resource, rule_applies_to)])

return resource_rules

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -300,33 +323,32 @@ def find_violations(self, resource, policy_binding):
         """
         violations = []
```
ahoying: Same comment as earlier on making this a chain, and then returning it as a flattened list at the end.

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -300,33 +323,32 @@ def find_violations(self, resource, policy_binding):
...
-                violations,
-                resource_rules.find_mismatches(resource, policy_binding))
+            resource_rules = self._get_resource_rules(curr_resource)
+
+            for resource_rule in resource_rules:
+                # Check whether rules match if the applies_to condition is met:
+                # SELF: check rules if the starting resource == current resource
+                # CHILDREN: check rules if starting resource != current resource
+                # SELF_AND_CHILDREN: always check rules
+                do_rules_check = (
```
ahoying: I clearer name might be rule_applies_to_resource.

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -269,22 +269,45 @@ def add_rule(self, rule_def, rule_index):
...
+        """
+        resource_rules = []
+        resource_rules.append(self.resource_rules_map.get(
+            (resource, RuleAppliesTo.SELF)))
+        resource_rules.append(self.resource_rules_map.get(
+            (resource, RuleAppliesTo.SELF_AND_CHILDREN)))
+        resource_rules.append(self.resource_rules_map.get(
+            (resource, RuleAppliesTo.CHILDREN)))
+
+        return [r for r in resource_rules if r]
```
blueandgold: nit: if you decide not to do what andrew mention above, then the more pythonic way  to write this is:
```python
if r is not None
```
http://stackoverflow.com/a/3965129/2830207

google/cloud/security/scanner/audit/org_rules_engine.py
carise: Done

google/cloud/security/scanner/audit/org_rules_engine.py
carise: Would you prefer that find_policy_violations() returns a list (or probably a set) or a generator?

ahoying: Since it is a small number, and maybe used or transformed by the caller, I would return a list or set object. Probably a list.

google/cloud/security/scanner/audit/org_rules_engine.py
carise: got it

google/cloud/security/scanner/audit/org_rules_engine.py
carise: Let me do address this in a bit -- going to submit some changes first.

google/cloud/security/scanner/audit/org_rules_engine.py
carise: I'll make a followup PR to refactor.

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -298,35 +391,34 @@ def find_violations(self, resource, policy_binding):
...
+
+                if not rule_applies_to_resource:
+                    continue
+
+                violations = itertools.chain(
+                    violations,
+                    resource_rule.find_mismatches(resource, policy_binding))
+
+                # If the rule does not inherit the parents' rules, stop.
+                if not resource_rule.inherit_from_parents:
```
ahoying: Can you verify this is working as expected? The logic doesn't seem correct. Make sure there are tests for breaking inheritance with multiple rule types applying at the current resource level (self, children, self_and_children). It may be the change I suggested for _get_resource_rules broke the assumption of rule order implicitly made by this check.

carise: ack. Working on tests for those cases.

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -23,18 +23,90 @@
...
+
+    Return:
+        A list of the violating members: rule members not found in the
+        policy (required-whitelist).
+    """
+    violating_members = []
+    for rule_member in rule_members:
+        found_in_rules = False
+        for policy_member in policy_members:
+            found_in_rules = rule_member.matches(policy_member)
```
mirons-google: You can drop the need for found_in_rules

``
if rule_member.matches(policy_member):
 break

violating_members.append(rule_member)
``

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -269,22 +340,44 @@ def add_rule(self, rule_def, rule_index):
...
-                    resource_rules = self.resource_rules_map.get(gcp_resource)
+                    resource_rules = self.resource_rules_map.get(
+                        (gcp_resource, applies_to))
 
                     # If no mapping exists, create it.
                     if not resource_rules:
                         resource_rules = ResourceRules(
                             resource=gcp_resource,
-                            applies_to=resource.get('applies_to'),
+                            applies_to=applies_to,
```
mirons-google: This applies_to=applies_to is not very clear to read/understand. Can you look at improving either the naming or code block to make this more readable?

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -269,22 +340,44 @@ def add_rule(self, rule_def, rule_index):
...
                             inherit_from_parents=rule_def.get(
                                 'inherit_from_parents', False))
-                        self.resource_rules_map[gcp_resource] = resource_rules
+                        self.resource_rules_map[
+                            (gcp_resource, applies_to)] = resource_rules
```
mirons-google: nit: awkward line break

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -562,7 +570,7 @@ def __init__(self, rule_name, rule_index, bindings, mode=None):
         """
         self.rule_name = rule_name
         self.rule_index = rule_index
-        self.bindings = self._get_bindings(bindings)
+        self.bindings = [IamPolicyBinding.create_from(b) for b in bindings]
```
mirons-google: Can we leave __init__ for variable setting and not executing code? Arguably if it's needed it should be done prior to using the class and then passed in.

google/cloud/security/scanner/audit/org_rules_engine.py
mirons-google: +1

google/cloud/security/scanner/audit/org_rules_engine.py
mirons-google: +1

google/cloud/security/scanner/audit/org_rules_engine.py
carise: any suggestions? don't know how to make this readable.

ahoying: Above, around line 346, you could set rules_map_key = (gcp_resource, applies_to), then for the get on line 346 and here you can use that variable, which is shorter. Another option is use () to move resource_rules to the next line.

self.resource_rules_map[<blah>] = (
    resource_rules)

ahoying: Or even shorter, rule_key = (...)

google/cloud/security/scanner/audit/org_rules_engine.py
carise: I think maybe it's the fact that it's mid afternoon but I am missing how this decides whether to append the rule member or not, if it is not found in the policy (but required to be in the policy).

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -31,6 +31,81 @@
...
+        policy_members: A list of IamPolicyMembers in the policy.
+
+    Return:
+        A list of the violating members: rule members not found in the
+        policy (required-whitelist).
+    """
+    violating_members = []
+    for rule_member in rule_members:
+        found_required_member = False
+        for policy_member in policy_members:
```
ahoying: Maybe change to:

found_required_member = any(rule_member.matches(m) for m in policy_members)

google/cloud/security/scanner/audit/org_rules_engine.py
```diff
@@ -31,6 +31,81 @@
...
+        policy_members: A list of IamPolicyMembers in the policy.
+
+    Return:
+        A list of the violating members: policy members NOT found in
+        the whitelist (rule members).
+    """
+    violating_members = []
+    for policy_member in policy_members:
+        is_whitelisted = False
+        for rule_member in rule_members:
```
ahoying: See comment below on use of any(), would also work for this function.

blueandgold:

PR #147, opened 2017-03-24
Make email sending optional.

### LINKED ISSUE #146: Sending email should be optional
If the recipient is specified, then send an email, otherwise don't.

blueandgold:
Based on user feedback, we want to make the email to be optional, especially when running initially for evaluation purposes.

The only other place where the email comes up is in the DM template.  Will we assume that email notifications will be required there?  If no, i can test it to make sure the optional case is handled correctly.
https://github.com/GoogleCloudPlatform/forseti-security/blob/master/deployment-templates/py/forseti-instance.py

Fixes #146

### DIFF
--- a/deployment-templates/deploy-forseti.yaml.sample
+++ b/deployment-templates/deploy-forseti.yaml.sample
@@ -58,6 +58,7 @@ resources:
     src-path: https://github.com/GoogleCloudPlatform/forseti-security/archive/v1.0.tar.gz
     cloudsqlproxy-os-arch: linux.amd64
     db-port: 3306
+    # Remove or comment out the email properties if you do not want to send email.
     sendgrid-api-key: YOUR_SENDGRID_API_KEY
     email-sender: EMAIL_ADDRESS_OF_YOUR_SENDER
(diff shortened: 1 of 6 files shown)


carise:
re: your question about DM -- is it possible to test whether the email address properties are present in the context.properties and conditionally add them as commandline flags in the startup script?

blueandgold:
I was concerned about that, too.  Let me verify how it behaves.

blueandgold:
PTAL, there's a bit more change to make sure we are handling the case when someone does not want to send email in the VM environment.  In this case, I am assuming the email properties are not specified.  In which case, a string 'None' will need to be handled since it None will be formatted to 'None'.

Let me know if there's a better way this can be handled.

carise:
I'm ok with the current changes, but maybe what you could also do for the DM is create a separate string for the email flags (instead of --sendgrid-api-key {} --email-sender {} etc. just have a {} and string replace the value with either a blank or the full list of flags/values), perhaps?

blueandgold:
Okay, I shuffled the strings around a bit in the DM template.  How does it look now?

blueandgold:

PR #152, opened 2017-03-24
Filter out non-ACTIVE lifecycle state projects

### LINKED ISSUE #141: Option to Retrieve Only ACTIVE Projects
Add an option to the get_project() to support filtering, such as a filter for only active projects.

https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/gcp_api/cloud_resource_manager.py

carise:
There is currently no ListProjects filter for lifecycleState. While we wait for that, we'll have to do the filtering externally from the API call (i.e. in the pipeline that loads the projects into inventory).

I added an arbitrary arguments option to CRM API's get_projects() so that the call to get_projects() from inventory will behave similarly to when the underlying CRM (v2?) API implements the lifecycleState filter. The get_projects() method does a manual filter for now until we can use the lifecycleState filter directly in the API call.

resolves #141


google/cloud/security/inventory/transform_util.py
```diff
@@ -39,6 +39,16 @@ def flatten_projects(projects):
     for project in (project for d in projects \
                     for project in d.get('projects', [])):
         project_json = json.dumps(project)
+
+        project_id = project.get('projectId')
+        lifecycle_state = project.get('lifecycleState')
```
mirons-google: nit: project_lifecycle_state

google/cloud/security/inventory/transform_util.py
```diff
@@ -39,6 +39,16 @@ def flatten_projects(projects):
...
+        parent = project.get('parent', {})
+
+        if lifecycle_state != 'ACTIVE':
```
blueandgold: qq: Instead of filtering here, would we want to do the filter when we actually make the crm api call?  i.e. only retrieve active projects?  Is there any reason why we would want to retrieve non-active projects?
https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/gcp_api/cloud_resource_manager.py#L86

carise: as noted above, the CRM API does not allow filtering on lifecycleState at the moment.

google/cloud/security/inventory/transform_util.py
carise: I was trying to avoid long variable names -- thoughts?

google/cloud/security/inventory/transform_util.py
blueandgold: Ah, I should've read the description.  :)  I feel that this filtering should be done upstream in the get_projects(), so that if anything else calls get_projects(), they would not have to do the filtering again.

carise:
Updated the filtering -- ptal

carise:
Thanks for the review!

blueandgold:

PR #157, opened 2017-03-27
Remove usage of db password for command options.

### LINKED ISSUE #154: Use config file for growing number of commandline settings
- [x] Avoid passing db_password as a commandline flag, as that could show up in logs or process list.
- [ ] Create a config file instead of using a lot of commandline flags.

blueandgold:
- Cloud SQL Proxy will be required for access.

- Cloud SQL Proxy permissioning will be:
  (1) service account on GCP env
  (2) developer on local env 

- This change only affects the local env, as the GCP env does not use db password.

Addresses #154


blueandgold:
Correct.  This solves the specific issue (as mentioned in issue #154) of passing in the db password via the commandline flag, which shouldn't be done anyways as a best practice.

blueandgold:

PR #159, opened 2017-03-28
Adding initial table schema for GSuite groups.

mirons-google:
One of many PRs to add Google Group expansion in IAM policies.

### DIFF
--- a/google/cloud/security/common/data_access/sql_queries/create_tables.py
+++ b/google/cloud/security/common/data_access/sql_queries/create_tables.py
@@ -73,3 +73,12 @@ CREATE_RAW_ORG_IAM_POLICIES_TABLE = """
         PRIMARY KEY (`id`)
     ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 """
+
+CREATE_GSUITE_ACCOUNT_GROUPS_TABLE = """
+    CREATE TABLE `{0}` (
+        `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+        `group_id` varchar(255) DEFAULT NULL,
+        `group_email` varchar(255) DEFAULT NULL,
+        PRIMARY KEY (`id`)
+    ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+"""


google/cloud/security/common/data_access/sql_queries/create_tables.py
```diff
@@ -73,3 +73,12 @@
         PRIMARY KEY (`id`)
     ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 """
+
+CREATE_GSUITE_ACCOUNT_GROUPS_TABLE = """
```
blueandgold: a couple of nits here:
(1) can we avoid using "ACCOUNT" here?  and just leave this as CREATE_GSUITE_GROUPS_TABLE ?
(2) can we make this more generic, i.e not use "GSUITE"?  as to avoid any fallout from name changes in the future?  and also to make this more platform neutral?

google/cloud/security/common/data_access/sql_queries/create_tables.py
```diff
@@ -73,3 +73,12 @@
...
+    CREATE TABLE `{0}` (
+        `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
```
blueandgold: Are there other attributes we like to save?  i.e. "name"

BTW, what is the resource description for the group response?  is there a link?

mirons-google: [Response](https://developers.google.com/admin-sdk/directory/v1/guides/manage-groups#get_all_domain_groups)

Right now I'm just collect group_id and group_email.

google/cloud/security/common/data_access/sql_queries/create_tables.py
mirons-google: Agreed. I put some thought into this too.

Currently the API is defined here as the [Directory API](https://developers.google.com/admin-sdk/directory/v1/guides/manage-groups#get_all_domain_groups). When fetching groups you can scope the request to an account, a domain, or a sub-domain. Since it's my intention to specify account in the request I felt it appropriate to name the table with the same scope.

As it relates to GSuite I'm open to other ideas.

google/cloud/security/common/data_access/sql_queries/create_tables.py
blueandgold: Looking at the groups resource definition, the "alias" might be interesting:
https://developers.google.com/admin-sdk/directory/v1/reference/groups#resource

Do you think that we would need aliases?  Could aliases be used somehow in permissioning? If so, it would help to collect it.

mirons-google:
We might, but this is intended to be a base implementation so I'm keeping it easy until we find out more.

blueandgold:

Issue #121, opened 2017-03-13
Modularize the Inventory Pipelines


blueandgold:
As reported by ahoying:

The pipelines ran by the inventory loader should be configurable and loadable as plugins, which should be self contained with their database schema, loaders, fetches, transformers, etc all bundled along with dependencies. 

Then the inventory_runner can get the list of enabled pipelines, figure out the order to process based on dependencies, raise exceptions if dependencies are missing, optionally launch different pipelines in parallel if they are not dependent on each other, etc.

Which would help migrate to dataflow (if we decide that's a good idea) down the road.

PR #160, opened 2017-03-28
Make the pipeline into actual objects.

blueandgold:
This PR address PR #121, to modularize the pipelines, and is the first step to build a dependency tree for all the pipelines and make them configurable.  By modular, I am aiming to make the pipeline more self-contained, isolated, and to be classes.

To make this PR easier to review, I am only including one pipeline.  I am open to any suggestions.  Once this first pipeline looks good, will revise the other pipelines and include them into this PR.


google/cloud/security/inventory/pipelines/_base_pipeline.py
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Base pipeline to load data into inventory."""
+
+
+# pylint: disable=too-few-public-methods
+# TODO: Look into improving to prevent using the disable.
```
ahoying: Switch to Abstract Base Class (ABC)

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -24,58 +24,78 @@
 from google.cloud.security.common.gcp_api.cloud_resource_manager import CloudResourceManagerClient
 from google.cloud.security.inventory import transform_util
 from google.cloud.security.inventory.errors import LoadDataPipelineError
+from google.cloud.security.inventory.pipelines._base_pipeline import _BasePipeline
```
ahoying: Re-enable line-too-long lint check after the imports.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -24,58 +24,78 @@
 from google.cloud.security.common.gcp_api.cloud_resource_manager import CloudResourceManagerClient
```
ahoying: Modules should be imported, not Classes, see https://google.github.io/styleguide/pyguide.html#Imports.
Example:

from google.cloud.security.common.data_access import errors as data_access_errors

...

except (data_access_errors.ApiExecutionError, data_access_errors.MySQLError) as e:
...

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -24,58 +24,78 @@
...
+        if org_id == '<organization id>':
+            raise LoadDataPipelineError('No organization id is specified.')
+
+        try:
+            # Retrieve data from GCP.
+            # Flatten the iterator since we will use it twice, and it is faster
+            # than cloning to 2 iterators.
+            iam_policies_map = self.gcp_api_client.get_org_iam_policies(
+                self.name, org_id)
+            # TODO: Investigate improving so the pylint disable isn't needed.
```
ahoying: Please fix pylint error.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -24,58 +24,78 @@
...
-
-        for i in iam_policies_map:
-            i['iam_policy'] = json.dumps(i['iam_policy'])
-        dao.load_data(RAW_ORG_IAM_POLICIES, cycle_timestamp, iam_policies_map)
-    except (CSVFileError, MySQLError) as e:
-        raise LoadDataPipelineError(e)
+class LoadOrgIamPoliciesPipeline(_BasePipeline):
+    """Pipeline to load org IAM policies data into Inventory."""
+
+    raw_org_iam_policies = 'raw_org_iam_policies'
```
ahoying: Class constants should be in CAPS.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -24,58 +24,78 @@
...
+            # than cloning to 2 iterators.
+            iam_policies_map = self.gcp_api_client.get_org_iam_policies(
+                self.name, org_id)
+            # TODO: Investigate improving so the pylint disable isn't needed.
+            # pylint: disable=redefined-variable-type
+            iam_policies_map = list(iam_policies_map)
+
+            # Flatten and relationalize data for upload to cloud sql.
+            flattened_iam_policies = (
+                self.transform_util.flatten_iam_policies(iam_policies_map))
```
ahoying: I'm not sure I understand the reason for setting transform_util to an attribute if each pipeline will have a different interface? It makes sense if each one implements a flatten() method that is specific to that pipeline.

Another use for an ABC.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -250,25 +248,22 @@ def main(argv):
...
-        {'pipeline': load_org_iam_policies_pipeline,
-         'status': ''},
+        LoadOrgIamPoliciesPipeline(
+            cycle_timestamp, configs, crm_rate_limiter, dao),
+        # TODO: add load projects pipeline
+        # TODO: add load project policies pipeline
     ]
 
+    succeeded = []
     for pipeline in pipelines:
```
ahoying: There is a lot happening in main() which makes testing hard. It would be worth just doing the bare initialization required in main then calling out to one or more functions that build, run and report the pipeline status, which can be tested. (Note for a future PR)

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -194,15 +192,15 @@ def _send_email(organization_id, cycle_time, cycle_timestamp, status, pipelines,
 
     for pipeline in pipelines:
         try:
-            pipeline['count'] = dao.select_record_count(
-                pipeline['pipeline'].RESOURCE_NAME,
+            pipeline.count = dao.select_record_count(
```
ahoying: Is it possible to set this from pipeline.run() without having the email function have to call back out to SQL?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -194,15 +192,15 @@ def _send_email(organization_id, cycle_time, cycle_timestamp, status, pipelines,
...
+                pipeline.name,
                 cycle_timestamp)
         except MySQLError as e:
             LOGGER.error('Unable to retrieve record count for %s_%s:\n%s',
-                         pipeline['pipeline'].RESOURCE_NAME,
+                         pipeline.name,
                          cycle_timestamp,
                          e)
-            pipeline['count'] = 'N/A'
+            pipeline.count = 'N/A'
```
ahoying: If some new code expects pipeline.count to be an int() this will blow up. I think set to None to show there was an error and the result is unknown.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
blueandgold: Done. Good suggestion.

blueandgold:
I will update the tests if the structure here looks good.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -49,53 +48,101 @@ def __init__(self, cycle_timestamp, configs, crm_rate_limiter, dao):
...
+            raise inventory_errors.LoadDataPipelineError(
+                'No organization id is specified.')
+
+        iam_policies_map = self._retrieve(org_id)
+
+        flattened_iam_policies = self._flatten(iam_policies_map)
+
+        self._load(iam_policies_map, flattened_iam_policies)
+
+        super(LoadOrgIamPoliciesPipeline, self)._get_loaded_count()
```
ahoying: You can directly call self._get_loaded_count(), you shouldn't need to call up to super, that's only needed if you override _get_loaded_count() in the child class.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -250,25 +235,22 @@ def main(argv):
     crm_rate_limiter = RateLimiter(max_crm_calls, 100)
 
     pipelines = [
-        {'pipeline': load_projects_pipeline,
-         'status': ''},
-        {'pipeline': load_projects_iam_policies_pipeline,
-         'status': ''},
-        {'pipeline': load_org_iam_policies_pipeline,
-         'status': ''},
+        LoadOrgIamPoliciesPipeline(
```
ahoying: Since this will break anyone relying on master/HEAD instead of a specific release, this shouldn't be merged into master until all the existing pipelines are ported over.

ahoying: Or you can do a hybrid where the un-ported pipelines are still run using the older code.

google/cloud/security/inventory/pipelines/base_pipeline.py
```diff
@@ -0,0 +1,99 @@
...
+        try:
+            self.count = self.dao.select_record_count(
+                self.name,
+                self.cycle_timestamp)
+        except data_access_errors.MySQLError as e:
+            self.logger.error('Unable to retrieve record count for %s_%s:\n%s',
+                self.name, self.cycle_timestamp, e)
+            self.count = None
+
+    def _parse_member_info(self, member):
```
ahoying: Is this a common enough function that it should be in the base class? Is it useful outside of the pipeline, where adding it to a common/util module would be better?

ahoying: In either case, it shouldn't be duplicated between transform_util.py and here.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -16,66 +16,133 @@
...
-        raise LoadDataPipelineError(e)
+from google.cloud.security.common.data_access import errors as data_access_errors
+from google.cloud.security.common.gcp_api._base_client import ApiExecutionError
+from google.cloud.security.common.gcp_api import cloud_resource_manager as crm
+from google.cloud.security.inventory import errors as inventory_errors
+from google.cloud.security.inventory.pipelines import base_pipeline
+# pylint: enable=line-too-long
+
+
+class LoadOrgIamPoliciesPipeline(base_pipeline._BasePipeline):
```
ahoying: There should be tests for this pipeline created at the same time.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Ack, I am planning to make a sweeping change to all the other pipelines and include them in this PR for merging.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
blueandgold: Ack, will work on testing now.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
blueandgold: Great, thanks!!

google/cloud/security/inventory/pipelines/base_pipeline.py
blueandgold: Done, moved to common/util.  Will probably delete the whole transform_util when I finish reworking all the pipelines.

blueandgold:
PTAL, some things to note:

- extended the original work to the other data pipelines (projects, project iam policies, groups).
- added unit tests for all the pipelines, except groups which I will circle back after this PR to do (don't want to hold this PR up any longer)
- merged changes.... hope nothing too much got lost :P 
- will keep looking for things to clean-up, as this PR is really big

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
blueandgold: Done.

mirons-google:
Would you also mind updating the branch with master so we can pickup the recent coveralls integration and see the increase in test coverage?

blueandgold:
Merged master into branch again.  I think I already did that in the beginning of this submit cycle as well.

google/cloud/security/inventory/pipelines/base_pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+        """Retrieve resource data from source."""
+        pass
+
+    @abc.abstractmethod
+    def _transform(self, resource_from_api):
+        """Transform api resource data into loadable format."""
+        pass
+
+    @abc.abstractmethod
+    def _load(self):
```
carise: Should we make the _load() signature

_load(self, **kwargs)

And then in the respective overloaded _load() method do the necessary handling?

blueandgold: This actually pointed out a better way to use this method, where it should be called once for each resource to be loaded, instead of passing multiple resources into it.

Example:
regular data and raw_json_data
```
_load(regular data)
_load(raw_json_data)
```
And once I did that, I can make this a concrete method, which saves a lot of code elsewhere.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -256,36 +247,47 @@ def main(argv):
     # Otherwise, there is a gap where the ratelimiter from one pipeline
     # is not used for the next pipeline using the same API. This could
     # lead to unnecessary quota errors.
-    crm_rate_limiter = CloudResourceManagerClient.get_rate_limiter()
-    admin_directory_rate_limiter = AdminDirectoryClient.get_rate_limiter()
+    # TODO: Provide the api client from the api client module instead of here.
+    max_crm_calls = configs.get('max_crm_api_calls_per_100_seconds', 400)
+    crm_rate_limiter = RateLimiter(max_crm_calls, 100)
+    crm_api_client = crm.CloudResourceManagerClient(
```
mirons-google: nit: I thought we had decided to push rate limit definitions/creation into the modules that support that code.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -34,21 +34,23 @@
...
 from google.cloud.security.common.data_access import db_schema_version
+from google.cloud.security.common.data_access import errors as data_access_errors
 from google.cloud.security.common.data_access.dao import Dao
-from google.cloud.security.common.data_access.errors import MySQLError
-# TODO: Investigate improving so we can avoid the pylint disable.
-# pylint: disable=line-too-long
 from google.cloud.security.common.data_access.sql_queries import snapshot_cycles_sql
-from google.cloud.security.common.gcp_api.admin_directory import AdminDirectoryClient
-from google.cloud.security.common.gcp_api.cloud_resource_manager import CloudResourceManagerClient
+from google.cloud.security.common.gcp_api import admin_directory
```
mirons-google: ... admin_directory as ad

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -256,36 +247,47 @@ def main(argv):
...
-    crm_rate_limiter = CloudResourceManagerClient.get_rate_limiter()
-    admin_directory_rate_limiter = AdminDirectoryClient.get_rate_limiter()
+    # TODO: Provide the api client from the api client module instead of here.
+    max_crm_calls = configs.get('max_crm_api_calls_per_100_seconds', 400)
+    crm_rate_limiter = RateLimiter(max_crm_calls, 100)
+    crm_api_client = crm.CloudResourceManagerClient(
+        rate_limiter=crm_rate_limiter)
+
+    # TODO: Make this configurable.
+    try:
```
mirons-google: We should try and standardize the client setup as much as possible.
- rate limit getting should be from the module
- rate limit setting should be passed into the creation of the client
- credentials should be built inside the client and never exposed here

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -14,112 +14,125 @@
...
-    flattened_groups = transform_util.flatten_groups(groups_map)
-
-    try:
-        dao.load_data(RESOURCE_NAME, cycle_timestamp, flattened_groups)
-    except (data_errors.CSVFileError, data_errors.MySQLError) as e:
-        raise inventory_errors.LoadDataPipelineError(e)
+from google.cloud.security.inventory.pipelines import base_pipeline
+
+
+class LoadGroupsPipeline(base_pipeline.BasePipeline):
```
mirons-google: Why did we remove the credential building from this class?

mirons-google: ...inventory.pipelines.base_pipeline import BasePipeLine

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -16,64 +16,139 @@
...
-        for i in iam_policies_map:
-            i['iam_policy'] = json.dumps(i['iam_policy'])
-        dao.load_data(RAW_ORG_IAM_POLICIES, cycle_timestamp, iam_policies_map)
-    except (data_errors.CSVFileError, data_errors.MySQLError) as e:
-        raise inventory_errors.LoadDataPipelineError(e)
+from google.cloud.security.inventory.pipelines import base_pipeline
+# pylint: enable=line-too-long
+
+
+class LoadOrgIamPoliciesPipeline(base_pipeline.BasePipeline):
```
mirons-google: See above about importing just the class.

google/cloud/security/common/util/parser.py
```diff
@@ -0,0 +1,38 @@
...
+        member_type: String of the member type.
+        member_name: String of the name portion of the member.
+        member_domain: String of the domain of the member.
+    """
+    member_type, email = member.split(":", 1)
+
+    if '@' in email:
+        member_name, member_domain = email.split('@', 1)
+    else:
+        # member is really something like domain:google.com
```
blueandgold: Done.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -34,21 +34,23 @@
...
 
+from datetime import datetime
 import sys
 
-from datetime import datetime
 import gflags as flags
+from ratelimiter import RateLimiter
 
-
+# TODO: Investigate improving so we can avoid the pylint disable.
```
blueandgold: Done.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Cool, that tidies things up a bit!

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -71,6 +73,8 @@
                     'The file with credentials for the service account.'
                     'NOTE: This is only required when running locally.')
 flags.DEFINE_string('organization_id', None, 'Organization ID.')
+flags.DEFINE_integer('max_crm_api_calls_per_100_seconds', 400,
```
blueandgold: Yes, I think this flag is needed for when people requests quota increases and there is a way to configure the increased quota.  What might work better is to move the pipeline-specific flags to each of the pipeline or api modules.  Will look into that.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -256,36 +247,47 @@ def main(argv):
     # Otherwise, there is a gap where the ratelimiter from one pipeline
     # is not used for the next pipeline using the same API. This could
     # lead to unnecessary quota errors.
-    crm_rate_limiter = CloudResourceManagerClient.get_rate_limiter()
-    admin_directory_rate_limiter = AdminDirectoryClient.get_rate_limiter()
+    # TODO: Provide the api client from the api client module instead of here.
```
blueandgold: Done.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Yup, clarifying the TODO from above to be clear about this.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Great points, adding these to the TODO descriptions.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -256,36 +247,47 @@ def main(argv):
...
-            pipeline['pipeline'].run(
-                dao, cycle_timestamp, configs, pipeline['rate_limiter'])
-            pipeline['status'] = 'SUCCESS'
+            pipeline.run()
+            pipeline.status = 'SUCCESS'
         except LoadDataPipelineError as e:
             LOGGER.error(
-                'Error loading data (%s).\n%s', pipeline, e)
-            pipeline['status'] = 'FAILURE'
+                'Encountered error to load data.\n%s', e)
```
blueandgold: Done.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -256,36 +247,47 @@ def main(argv):
...
+            pipeline.status = 'FAILURE'
```
blueandgold: Done.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -14,112 +14,109 @@
...
-        config.get('service_account_credentials_file'),
-        config.get('domain_super_admin_email')]
+    def _can_inventory_google_groups(self):
+        """A simple function that validates required inputs to inventory groups.
 
-    if _is_our_environment_gce():
-        required_execution_config = required_gcp_execution_config
-    else:
-        required_execution_config = required_local_execution_config
+        Args:
```
ahoying: Nit: You don't need an Args section if there are no args. Same below where you've added an Args: None section.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -14,112 +14,109 @@
...
-def _build_proper_credentials(config):
-    """Build proper credentials required for accessing the directory API.
+        required_local_execution_config = [
+            self.configs.get('service_account_email'),
+            self.configs.get('service_account_credentials_file'),
+            self.configs.get('domain_super_admin_email')]
 
-    Args:
-        config: Dictionary of configurations.
+        if metadata_server.can_reach_metadata_server():
```
ahoying: Unrelated to this change, but is the result of this call cached? We should use memoize or similar so that after the first check the cached result is always returned, regardless of how often it is called.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -16,64 +16,113 @@
...
+
+    def __init__(self, cycle_timestamp, configs, crm_client, dao, parser):
+        """Constructor for the data pipeline.
+
+        Args:
+            cycle_timestamp: String of timestamp, formatted as YYYYMMDDTHHMMSSZ.
+            configs: Dictionary of configurations.
+            crm_client: CRM API client.
+            dao: Data access object.
+            parser: Forseti parser object.
```
ahoying: It's odd to have this passed in and set to an object attribute. Will you ever point this to something other than google.cloud.security.common.util.parser? The import should just be local to this pipeline module and used directly, rather than as self.parser. Same for the other pipelines in this PR.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -16,64 +16,113 @@
...
+
+        Returns:
+            None
+
+        Raises:
+            LoadDataPipelineException: An error with loading data has occurred.
+        """
+        org_id = self.configs.get('organization_id')
+        # Check if the placeholder is replaced in the config/flag.
+        if org_id == '<organization id>':
```
ahoying: Another option, to catch more classes or configuration error, would be to do something like:

try:
  org_id = int(self.configs.get('organization_id'))
except (TypeError, ValueError):
  raise ...

Same below in the other pipelines that use org id.

Or this should be a validator on the flag itself and raised earlier.

google/cloud/security/inventory/pipelines/load_org_iam_policies_pipeline.py
```diff
@@ -16,64 +16,113 @@
...
+            raise inventory_errors.LoadDataPipelineError(
+                'No organization id is specified.')
+
+        iam_policies_map = self._retrieve()
+
+        loadable_iam_policies = self._transform(iam_policies_map)
+
+        self._load(self.name, loadable_iam_policies)
```
ahoying: Nit: I think it would be more readable here and elsewhere in the pipeline code to just use the class constant (self.RESOURCE_NAME) instead of the initialized in the base class attribute self.name, which is harder to trace back when reviewing the code. Particularly since you go on to use self.RAW_RESOURCE_NAME below.

tests/inventory/pipelines/base_pipeline_test.py
```diff
@@ -0,0 +1,103 @@
...
+            self.pipeline.cycle_timestamp,
+            EXPECTED_LOADABLE_PROJECTS)       
+
+    def test_load_errors_are_handled(self):
+        """Test that errors are handled when loading."""
+
+        self.pipeline.dao.load_data.side_effect = (
+            data_access_errors.MySQLError('error error', mock.MagicMock()))
+
+        self.assertRaises(inventory_errors.LoadDataPipelineError,
```
ahoying: Nit: The context manager version of assertRaises is easier to read for tests like this one.

with self.assertRaises(...):
  self.pipeline._load(...)

tests/inventory/pipelines/base_pipeline_test.py
```diff
@@ -0,0 +1,103 @@
...
+    
+    Because base pipeline is an abstract class, there is no way to instantiate
+    it for testing directly.  So, will test it by a pipeline that implements it.
+    """
+
+    def setUp(self):
+        """Set up."""
+
+        self.cycle_timestamp = '20001225T120000Z'
+        self.configs = {'organization_id': '66666',
```
ahoying: You could choose to move the config dict into your test_data as well instead of copying it into each test class.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -47,7 +47,7 @@
 from google.cloud.security.common.data_access import errors as data_access_errors
 from google.cloud.security.common.data_access.dao import Dao
 from google.cloud.security.common.data_access.sql_queries import snapshot_cycles_sql
-from google.cloud.security.common.gcp_api import admin_directory
+from google.cloud.security.common.gcp_api import admin_directory as ad
```
carise: @ahoying has requested that we make the import statements consistent and import the module instead of the class, so that's the approach that @blueandgold and I have been taking -- I can't seem to find his comment about why though

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -256,36 +249,52 @@ def main(argv):
...
-    crm_rate_limiter = CloudResourceManagerClient.get_rate_limiter()
-    admin_directory_rate_limiter = AdminDirectoryClient.get_rate_limiter()
+    #
+    # TODO: Move the building of the rate limiter and credential
+    # to the api client:
+    # rate limit getting should be from the module
+    # rate limit setting should be passed into the creation of the client
+    # credentials should be built inside the client and never exposed here
+    max_crm_calls = configs.get('max_crm_api_calls_per_100_seconds', 400)
+    crm_rate_limiter = RateLimiter(max_crm_calls, 100)
```
carise: @mirons-google do you mean CloudResourceManagerClient.get_rate_limiter()?

google/cloud/security/inventory/inventory_loader.py
carise: update -- here's the link from the style guide https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports

"Use imports for packages and modules only"

google/cloud/security/inventory/inventory_loader.py
mirons-google: Yes, mirror the functionality of AdminDirectoryClient.get_rate_limiter() in CloudResourceManagerClient

tests/inventory/pipelines/base_pipeline_test.py
blueandgold: Done

tests/inventory/pipelines/base_pipeline_test.py
blueandgold: Done. Fixed everywhere.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -234,14 +225,16 @@ def _send_email(organization_id, cycle_time, cycle_timestamp, status, pipelines,
     except EmailSendError:
         LOGGER.error('Unable to send email that inventory snapshot completed.')
 
+# TODO: Investigate improving so the pylint disable isn't needed.
+# pylint: disable=too-many-locals
 def main(argv):
```
blueandgold: Adding a TODO here, and will follow up in a separate PR, with the proper testing.

blueandgold:

PR #165, opened 2017-03-29
Adding fetching and storing of groups

mirons-google:
This builds on #159 

A couple caveats:
- This only fetches group_emails and group_ids
- This only works running locally, DM execution will be in a follow-up CL
- A follow-up CL will be done to update the documentation given the extra requirements for using DWD

### DIFF
--- a/google/cloud/security/common/data_access/load_data_sql_provider.py
+++ b/google/cloud/security/common/data_access/load_data_sql_provider.py
@@ -14,25 +14,17 @@
 
 """Provides the load data sql for resources."""
 
-# TODO: Investigate improving so we can avoid the pylint disable.
-# pylint: disable=line-too-long
-from google.cloud.security.common.data_access.csv_writer import GROUPS_FIELDNAMES
-from google.cloud.security.common.data_access.csv_writer import ORG_IAM_POLICIES_FIELDNAMES
-from google.cloud.security.common.data_access.csv_writer import PROJECTS_FIELDNAMES
-from google.cloud.security.common.data_access.csv_writer import PROJECT_IAM_POLICIES_FIELDNAMES
-from google.cloud.security.common.data_access.csv_writer import RAW_ORG_IAM_POLICIES_FIELDNAMES
-from google.cloud.security.common.data_access.csv_writer import RAW_PROJECT_IAM_POLICIES_FIELDNAMES
+from google.cloud.security.common.data_access import csv_writer
 from google.cloud.security.common.data_access.sql_queries import load_data
-# pylint: enable=line-too-long
 
 
 FIELDNAME_MAP = {
-    'groups': GROUPS_FIELDNAMES,
-    'org_iam_policies': ORG_IAM_POLICIES_FIELDNAMES,
-    'projects': PROJECTS_FIELDNAMES,
-    'project_iam_policies': PROJECT_IAM_POLICIES_FIELDNAMES,
-    'raw_org_iam_policies': RAW_ORG_IAM_POLICIES_FIELDNAMES,
-    'raw_project_iam_policies': RAW_PROJECT_IAM_POLICIES_FIELDNAMES,
+    'groups': csv_writer.GROUPS_FIELDNAMES,
+    'org_iam_policies': csv_writer.ORG_IAM_POLICIES_FIELDNAMES,
+    'projects': csv_writer.PROJECTS_FIELDNAMES,
+    'project_iam_policies': csv_writer.PROJECT_IAM_POLICIES_FIELDNAMES,
+    'raw_org_iam_policies': csv_writer.RAW_ORG_IAM_POLICIES_FIELDNAMES,
+    'raw_project_iam_policies': csv_writer.RAW_PROJECT_IAM_POLICIES_FIELDNAMES,
 }
 
 
--- a/google/cloud/security/common/data_access/load_data_sql_provider.py
+++ b/google/cloud/security/common/data_access/load_data_sql_provider.py
@@ -16,6 +16,7 @@
 
 # TODO: Investigate improving so we can avoid the pylint disable.
 # pylint: disable=line-too-long
+from google.cloud.security.common.data_access.csv_writer import GROUPS_FIELDNAMES
 from google.cloud.security.common.data_access.csv_writer import ORG_IAM_POLICIES_FIELDNAMES
 from google.cloud.security.common.data_access.csv_writer import PROJECTS_FIELDNAMES
 from google.cloud.security.common.data_access.csv_writer import PROJECT_IAM_POLICIES_FIELDNAMES
@@ -26,6 +27,7 @@ from google.cloud.security.common.data_access.sql_queries import load_data
 
 
 FIELDNAME_MAP = {
+    'groups': GROUPS_FIELDNAMES,
     'org_iam_policies': ORG_IAM_POLICIES_FIELDNAMES,
     'projects': PROJECTS_FIELDNAMES,
     'project_iam_policies': PROJECT_IAM_POLICIES_FIELDNAMES,
--- a/google/cloud/security/common/gcp_api/_base_client.py
+++ b/google/cloud/security/common/gcp_api/_base_client.py
@@ -16,6 +16,7 @@
 
 from apiclient import discovery
 from oauth2client.client import GoogleCredentials
+
 from retrying import retry
 
 from google.cloud.security.common.gcp_api._supported_apis import SUPPORTED_APIS
--- a/google/cloud/security/common/gcp_api/_base_client.py
+++ b/google/cloud/security/common/gcp_api/_base_client.py
@@ -16,7 +16,6 @@
 
 from apiclient import discovery
 from oauth2client.client import GoogleCredentials
-
 from retrying import retry
 
 from google.cloud.security.common.gcp_api._supported_apis import SUPPORTED_APIS
--- a/google/cloud/security/common/gcp_api/admin_directory.py
+++ b/google/cloud/security/common/gcp_api/admin_directory.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Wrapper for Resource Manager API client."""
+"""Wrapper for Admin Directory  API client."""
 
 from googleapiclient.errors import HttpError
 from httplib2 import HttpLib2Error
@@ -32,8 +32,8 @@ class AdminDirectoryClient(_BaseClient):
     API_NAME = 'admin'
     DEFAULT_MAX_QUERIES = 400
 
-    def __init__(sefl, credentials=None, rate_limiter=None):
-      super(AdminDirectoryManagerClient,self).__init__(
+    def __init__(self, credentials=None, rate_limiter=None):
+      super(AdminDirectoryClient, self).__init__(
           credentials=credentials, api_name=self.API_NAME)
       if rate_limiter:
           self.rate_limiter = rate_limiter
@@ -60,9 +60,9 @@ class AdminDirectoryClient(_BaseClient):
             try:
                 with self.rate_limiter:
                     response = self._execute(request)
-                    results.extend(response.items('groups', [])
+                    results.extend(response.items('groups', []))
                     request = groups_stub.list_next(request, response)
             except (HttpError, HttpLib2Error) as e:
-                LOGGER.error(ApiExecutionError(project_id, e))
+                raise ApiExecutionError(groups_stub, e)
 
         return results
--- a/google/cloud/security/common/gcp_api/admin_directory.py
+++ b/google/cloud/security/common/gcp_api/admin_directory.py
@@ -71,6 +71,7 @@ class AdminDirectoryClient(_BaseClient):
         request = groups_stub.list(customer=customer_id)
         results = []
 
+        # TODO: Investigate yeilding results to handle large group lists.
         while request is not None:
             try:
                 with self.rate_limiter:
--- a/google/cloud/security/common/gcp_api/admin_directory.py
+++ b/google/cloud/security/common/gcp_api/admin_directory.py
@@ -25,23 +25,35 @@ from google.cloud.security.common.util.log_util import LogUtil
 
 LOGGER = LogUtil.setup_logging(__name__)
 
+def get_rate_limiter():
+    DEFAULT_MAX_QUERIES = 150000
+    DEFAULT_RATE_BUCKET_SECONDS = 86400
+
+    return RateLimiter(
+        DEFAULT_DAILY_MAX_QUERIES,
+        DEFAULT_RATE_BUCKET_SECONDS)
+
 
 class AdminDirectoryClient(_BaseClient):
     """GSuite Admin Directory API Client."""
 
     API_NAME = 'admin'
-    DEFAULT_MAX_QUERIES = 400
 
-    def __init__(self, credentials=None, rate_limiter=None):
+    def __init__(self, credentials=None, rate_limiter):
         super(AdminDirectoryClient, self).__init__(
             credentials=credentials, api_name=self.API_NAME)
         if rate_limiter:
             self.rate_limiter = rate_limiter
         else:
-            self.rate_limiter = RateLimiter(self.DEFAULT_MAX_QUERIES, 100)
+            self.rate_limiter = get_rate_limiter()
 
     def get_groups(self, customer_id='my_customer'):
         """Get all the groups for a given customer_id.
+
+        A note on customer_id='my_customer'. This is a magic string instead
+        of using the real customer id.
+
+        See: https://developers.google.com/admin-sdk/directory/v1/guides/manage-groups#get_all_domain_groups
 
         Args:
             customer_id: The customer id to scope the request to
--- a/google/cloud/security/common/gcp_api/admin_directory.py
+++ b/google/cloud/security/common/gcp_api/admin_directory.py
@@ -1,0 +1,68 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Wrapper for Resource Manager API client."""
+
+from googleapiclient.errors import HttpError
+from httplib2 import HttpLib2Error
+from ratelimiter import RateLimiter
+
+from google.cloud.security.common.gcp_api._base_client import _BaseClient
+from google.cloud.security.common.gcp_api._base_client import ApiExecutionError
+from google.cloud.security.common.util.log_util import LogUtil
+
+
+LOGGER = LogUtil.setup_logging(__name__)
+
+
+class AdminDirectoryClient(_BaseClient):
+    """GSuite Admin Directory API Client."""
+
+    API_NAME = 'admin'
+    DEFAULT_MAX_QUERIES = 400
+
+    def __init__(sefl, credentials=None, rate_limiter=None):
+      super(AdminDirectoryManagerClient,self).__init__(
+          credentials=credentials, api_name=self.API_NAME)
+      if rate_limiter:
+          self.rate_limiter = rate_limiter
+      else:
+          self.rate_limiter = RateLimiter(self.DEFAULT_MAX_QUERIES, 100)
+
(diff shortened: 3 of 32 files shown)


google/cloud/security/common/gcp_api/_base_client.py
```diff
@@ -16,6 +16,7 @@
 
 from apiclient import discovery
 from oauth2client.client import GoogleCredentials
+
```
blueandgold: nit: don't think an extra line is needed here; as retrying is an installed library and should be grouped with the rest of the installed library.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,88 @@
...
+
+    credentials = ServiceAccountCredentials.from_json_keyfile_name(
+        configs.get('service_account_credentials_file'),
+        scopes=REQUIRED_SCOPES)
+
+    return credentials.create_delegated(
+        configs.get('domain_super_admin_email'))
+
+
+def run(dao=None, cycle_timestamp=None, configs=None, crm_rate_limiter=None):
```
blueandgold: nit: probably should not use "crm_rate_limiter", and just call it "admin_rate_limiter" or equivalent

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -257,15 +298,27 @@ def main(argv):
          'status': ''},
     ]
 
+    if configs.get('inventory_groups'):
+        if _should_inventory_google_groups(configs):
```
blueandgold: nit: the check is more about "can" instead of "should", so maybe change the name to _can_inventory_google_groups?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -223,6 +237,33 @@ def _send_email(organization_id, cycle_time, cycle_timestamp, status, pipelines,
...
+def _should_inventory_google_groups(configs):
+    """A simple function that validates required inputs for inventorying groups.
+
+    Args:
+        configs: Dictionary of configurations built from our flags.
+
+    Returns:
+        Boolean
+    """
+    if not configs.get('service_account_email'):
```
blueandgold: nit: You probably can consolidate these checks into something simplier:

```python
required_info = [configs.get('service_account_email'),
                 configs.get('service_account_credentials_file'),
                 configs.get('domain_super_admin_email')]

if False in required_info:
  LOGGER.error('Unable to inventory groups without required info: %s', required_info).
  return False
```

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -223,6 +237,33 @@ def _send_email(organization_id, cycle_time, cycle_timestamp, status, pipelines,
...
+        LOGGER.error('Unable to inventory groups with a service accoutn email.')
```
blueandgold: nit: without, and spelling on account

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -223,6 +237,33 @@ def _send_email(organization_id, cycle_time, cycle_timestamp, status, pipelines,
...
+        return False
+
+    if metadata_server.can_reach_metadata_server():
```
blueandgold: nit: Can this metadata server check be moved up, as the first condition?

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+LOGGER = LogUtil.setup_logging(__name__)
+
+
+class AdminDirectoryClient(_BaseClient):
+    """GSuite Admin Directory API Client."""
+
+    API_NAME = 'admin'
+    DEFAULT_MAX_QUERIES = 400
```
carise: I noticed in the IAM & Admin -> Quotas dashboard, under "All Quotas", that there doesn't seem to be a ratelimit for the Admin SDK, but just the total number of queries that can be made to it in a day. Do you happen to know what would happen if someone should exceed that quota? It looks like the default is 150,000 queries per day.

google/cloud/security/inventory/inventory_loader.py
blueandgold: nit: I feel a better place for this validation might be inside the load_groups_pipeline module, and raise LoadDataPipelineError if it can not run.

Here in the loader, it should be checking if the pipeline needs to run or not.  Whether it can or not should be inside the load_groups_pipeline module.

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __init__(self, credentials=None, rate_limiter=None):
+        super(AdminDirectoryClient, self).__init__(
+            credentials=credentials, api_name=self.API_NAME)
+        if rate_limiter:
+            self.rate_limiter = rate_limiter
+        else:
+            self.rate_limiter = RateLimiter(self.DEFAULT_MAX_QUERIES, 100)
+
+    def get_groups(self, customer_id='my_customer'):
```
carise: is "my_customer" the default or should you just make it None?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -45,17 +49,27 @@
...
 from google.cloud.security.inventory.errors import LoadDataPipelineError
+from google.cloud.security.inventory.pipelines import load_groups_pipeline
 from google.cloud.security.inventory.pipelines import load_org_iam_policies_pipeline
 from google.cloud.security.inventory.pipelines import load_projects_iam_policies_pipeline
 from google.cloud.security.inventory.pipelines import load_projects_pipeline
 # pylint: enable=line-too-long
 
 FLAGS = flags.FLAGS
 
+flags.DEFINE_bool('inventory_groups', False,
```
carise: for now, the flags are ok. But what happens when we have 20 different things that we want to inventory? Scanner will also be affected.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
carise: Could we just call it "rate_limiter"?

google/cloud/security/common/gcp_api/admin_directory.py
blueandgold: Nice catch, if this is the case, the ratelimiter needs to be defined like:

self.DEFAULT_MAX_QUERIES = 150000
RateLimiter(self.DEFAULT_MAX_QUERIES, 86400)

google/cloud/security/common/gcp_api/admin_directory.py
mirons-google: Great question. This is magic string for this API for fetching all groups w/o having to know the account id.

See 'my_customer':
https://developers.google.com/admin-sdk/directory/v1/guides/manage-groups#get_all_domain_groups

google/cloud/security/inventory/inventory_loader.py
mirons-google: Yes, I noticed this too. I was thinking we should open an issue and move to a configuration file (YAML) for these things that way either module can read from it.

google/cloud/security/common/gcp_api/admin_directory.py
carise: Got it. A short explanation/link would be a nice clarification here then.

google/cloud/security/common/gcp_api/_base_client.py
mirons-google: done

google/cloud/security/common/gcp_api/admin_directory.py
mirons-google: done

google/cloud/security/common/gcp_api/admin_directory.py
mirons-google: done

google/cloud/security/inventory/inventory_loader.py
mirons-google: done

google/cloud/security/inventory/inventory_loader.py
mirons-google: done

google/cloud/security/inventory/inventory_loader.py
mirons-google: done

google/cloud/security/inventory/inventory_loader.py
mirons-google: done

mirons-google: done

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
mirons-google: done.

I also improved how rate_limiters are created and fetched for inventory_loader.py

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    required_gcp_execution_config = [config.get('service_account_email'),
+                                     config.get('domain_super_admin_email')]
+
+    required_local_execution_config = [
+        config.get('service_account_email'),
+        config.get('service_account_credentials_file'),
+        config.get('domain_super_admin_email')]
+
+    if metadata_server.can_reach_metadata_server():
```
blueandgold: nit: Is there any reason why both the local and gcp execution config needs to be checked?  Maybe we can determine the environment separately, and check the execution config once?

Do you think this might be better and easier to read??

```python
    if is_gce():
        required_execution_config = required_gcp_execution_config
    else:
        required_execution_config = required_local_execution_config

    if metadata_server.can_reach_metadata_server():
        if False in required_execution_config:
            return False
        return True
```

google/cloud/security/common/data_access/load_data_sql_provider.py
```diff
@@ -16,6 +16,7 @@
 
 # TODO: Investigate improving so we can avoid the pylint disable.
 # pylint: disable=line-too-long
+from google.cloud.security.common.data_access.csv_writer import GROUPS_FIELDNAMES
```
carise: I think Andrew has been suggesting importing the module instead of the class or variable.

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -0,0 +1,83 @@
...
+        super(AdminDirectoryClient, self).__init__(
+            credentials=credentials, api_name=self.API_NAME)
+        if rate_limiter:
+            self.rate_limiter = rate_limiter
+        else:
+            self.rate_limiter = self.get_rate_limiter()
+
+    @staticmethod
+    def get_rate_limiter():
+        """Return an appriopriate rate limiter."""
```
carise: Is there a reason this is static vs instance method, or is it possible for us to generalize it for all the API clients (we can do this in a future PR)?

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
mirons-google: I implemented a variation of this. PTAL.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+
+    if _is_our_environment_gce():
+        required_execution_config = required_gcp_execution_config
+    else:
+        required_execution_config = required_local_execution_config
+
+    if False in required_execution_config:
+      return False
+
+    return True
```
blueandgold: qq: do we not need to know the metadata server anymore, i.e. include it as part of the check?

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+    Args:
+        config: Dictionary of configurations built from our config.
+
+    Returns:
+        Boolean
+    """
+    required_gcp_execution_config = [config.get('service_account_email'),
+                                     config.get('domain_super_admin_email')]
+
+    required_local_execution_config = [
```
blueandgold: nit: i just realized that you can do something like this.  totally optional.
```python
required_local_execution_config = required_gcp_execution_config.extend(config.get('service_account_credentials_file'))
```

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
carise: Would False be found in the required_execution_config if config.get('missing_key') can't find that key? I thought config.get('missing_key') returns None.

What about

return metadata_server.can_reach_metadata_server() and all(required_execution_config)

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+        config.get('service_account_email'),
+        config.get('service_account_credentials_file'),
+        config.get('domain_super_admin_email')]
+
+    if _is_our_environment_gce():
+        required_execution_config = required_gcp_execution_config
+    else:
+        required_execution_config = required_local_execution_config
+
+    if False in required_execution_config:
```
carise: Would False be found in the required_execution_config if config.get('missing_key') can't find that key? I thought config.get('missing_key') returns None.

What about

return metadata_server.can_reach_metadata_server() and all(required_execution_config)

mirons-google:
I took a little bit of both from your nit's. PTAL.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
mirons-google: ack. I think it's more explicit this way.

google/cloud/security/common/data_access/load_data_sql_provider.py
ahoying: Right, see https://google.github.io/styleguide/pyguide.html#Imports.

mirons-google: done

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -45,20 +47,28 @@
...
-flags.DEFINE_integer('max_crm_api_calls_per_100_seconds', 400,
-                     'Cloud Resource Manager queries per 100 seconds.')
-
+flags.DEFINE_bool('inventory_groups', False,
+                  'Whether to inventory GSuite Groups.')
+flags.DEFINE_string('domain_super_admin_email', None,
+                    'An email address of a super-admin in the GSuite domain.')
+flags.DEFINE_string('service_account_email', None,
+                    'The email of the service account.')
+flags.DEFINE_string('service_account_credentials_file', None,
```
ahoying: Since this only required when not running on GCE, please include that in the flag description.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+def _can_inventory_google_groups(config):
+    """A simple function that validates required inputs for inventorying groups.
+
+    Args:
+        config: Dictionary of configurations built from our config.
+
+    Returns:
+        Boolean
+    """
+    required_gcp_execution_config = [config.get('service_account_email'),
```
ahoying: Is the service_account_email discoverable when running on GCE from the metadata server?

google/cloud/security/common/gcp_api/admin_directory.py
mirons-google: It seemed unimportant for the method to know about it's class. It's placement here was truly for organization reasons as it seemed to be the best spot.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+from google.cloud.security.common.gcp_api.admin_directory import AdminDirectoryClient
+from google.cloud.security.common.util import metadata_server
+from google.cloud.security.common.util.log_util import LogUtil
+from google.cloud.security.inventory import transform_util
+from google.cloud.security.inventory.errors import LoadDataPipelineError
+
+
+LOGGER = LogUtil.setup_logging(__name__)
+RESOURCE_NAME = 'groups'
+REQUIRED_SCOPES = ['https://www.googleapis.com/auth/admin.directory.user',
```
ahoying: Can we use the .readonly version of these scopes?

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+        config.get('service_account_credentials_file'),
+        config.get('domain_super_admin_email')]
+
+    if _is_our_environment_gce():
+        required_execution_config = required_gcp_execution_config
+    else:
+        required_execution_config = required_local_execution_config
+
+    if False or None in required_execution_config:
+        return False
```
mirons-google: cool, done

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        groups_stub = self.service.groups()
+        request = groups_stub.list(customer=customer_id)
+        results = []
+
+        while request is not None:
+            try:
+                with self.rate_limiter:
+                    response = self._execute(request)
+                    results.extend(response.get('groups', []))
```
ahoying: How big can this get? Could the script OOM? Could this be done as a generator instead of returning all the results  at once?

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+from google.cloud.security.common.data_access.errors import MySQLError
+from google.cloud.security.common.gcp_api._base_client import ApiExecutionError
+# TODO: Investigate improving so the pylint disable isn't needed.
+# pylint: disable=line-too-long
+from google.cloud.security.common.gcp_api.admin_directory import AdminDirectoryClient
+from google.cloud.security.common.util import metadata_server
+from google.cloud.security.common.util.log_util import LogUtil
+from google.cloud.security.inventory import transform_util
+from google.cloud.security.inventory.errors import LoadDataPipelineError
+
```
ahoying: Re-enable line-top-long after imports.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
ahoying: Also, this should be a frozenset()

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
```diff
@@ -0,0 +1,123 @@
...
+
+def _build_proper_credentials(config):
+    """Build proper credentials required for accessing the directory API.
+
+    Args:
+        config: Dictionary of configurations.
+
+    Returns:
+        Credentials as built by oauth2client.
```
ahoying: Will any of these calls Raise exceptions?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -17,11 +17,14 @@
 Usage:
 
   $ forseti_inventory \\
+      --inventory_groups \\
+      --service_account_email <email of the service account> \\
```
ahoying: Are the new flags documented in the readme?

mirons-google: I wanted to keep the PRs very focused. A document update is being prepared [here](https://github.com/GoogleCloudPlatform/forseti-security/compare/groups-docs)

google/cloud/security/common/gcp_api/admin_directory.py
mirons-google: Some domains could have 1000s of groups. I'll leave a TODO to investigate improving this.

This was also the recommended best way to page through results.

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
mirons-google: done

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
mirons-google: done and done

google/cloud/security/inventory/pipelines/load_groups_pipeline.py
mirons-google: done

google/cloud/security/inventory/inventory_loader.py
mirons-google: done

mirons-google:
Thanks for the review. PTAL.

RE: Tests
I had hoped to follow-up with that once I've actually started flattening groups data.

google/cloud/security/common/gcp_api/admin_directory.py
ahoying: I like the idea of making it part of the base class if possible, and each API client overriding and calling the base class with the specific rate limits. This can be a future PR.

blueandgold:

PR #168, opened 2017-03-30
Sync 1.0.1 release-version with src-path

mcunha:
This fixes people updating deployment templates and still not deploying 1.0.1 unless they spot src-path doesn't match release-version.

Would it be better to interpolate release-version into src-path ?


carise:
Nice catch. I don't think it's so straightforward for the string interpolation of the release version but it's something to think about for sure.

blueandgold:

PR #175, opened 2017-04-03
Updating documentation for GSuite Groups inventorying

mirons-google:
This is a follow-up to #165 

* It updates the documentation to support the creation of a new service account
* How to enable domain-wide delegation for that service account
* How to pass these flags into `inventory_loader.py`

I'll follow-up with GCP installation specific instructions once @blueandgold is done re-working the pipelines at the same time I create a PR for enabling GSuite Google Group checking on GCP.


blueandgold:

PR #180, opened 2017-04-04
Improving the metadata module

mirons-google:
This PR supports needed functionality to request the value of a specific attribute which is required for fetching of DWD keys when on GCP.


blueandgold:

Issue #62, opened 2017-03-06
Inventory should support Google Groups membership


mirons-google:
* Currently there is no API to expand membership

blueandgold:
A lot more details are added in #126.  Will keep both linked for now.

blueandgold:
Initial PRs #159, #165

mirons-google:
Duplicate of #126 

Closing.

PR #183, opened 2017-04-05
Expanding metadata

mirons-google:
- Support future changes for DWD key fetching from the metadata server to support #62 
- Increased test coverage for metadata_server


google/cloud/security/common/util/metadata_server.py
```diff
@@ -20,29 +20,87 @@
...
+
+METADATA_SERVER_HOSTNAME = 'metadata.google.internal'
 METADATA_SERVER_CONN_TIMEOUT = 2
+REQUIRED_METADATA_HEADER = {'Metadata-Flavor': 'Google'}
+HTTP_SUCCESS = httplib.OK
+HTTP_GET = 'GET'
+
+
+class MetadataBaseError(Exception):
+    """Base error class for handling meta_data errors."""
```
carise: Could you put the errors in a common util_errors module or similar? Also, we are following the following conventions for custom error classes:

1. Have a base error class that inherits from Exception (e.g. https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/gcp_api/errors.py#L18)
2. Custom error classes inherit from the base error class

google/cloud/security/common/util/metadata_server.py
```diff
@@ -20,29 +20,87 @@
...
+    """For a given key return the value.
+
+    Args:
+        attribute: a String representing the key.
+
+    Returns:
+        The value of the request.
+    """
+    path = '/computeMetadata/v1/instance/attributes/'
+    path = path + attribute
```
carise: should this be a string interpolation?

google/cloud/security/common/util/metadata_server.py
```diff
@@ -20,29 +20,87 @@
...
+    http_response = ''
+
+    try:
+        http_response = _issue_http_request(
+            path, HTTP_GET, REQUIRED_METADATA_HEADER)
+        return http_response.read()
+    except MetadataHttpError:
+        return http_response
```
carise: should we return None or raise an error instead of an empty string?

tests/common/util/metadata_server_test.py
```diff
@@ -7,7 +7,7 @@
 #    http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
+  # distributed under the License is distributed on an "AS IS" BASIS,
```
carise: I think we can remove the extra space before the "#"

mirons-google: done

google/cloud/security/common/util/metadata_server.py
mirons-google: done

google/cloud/security/common/util/metadata_server.py
mirons-google: done

mirons-google:
Thanks for the feedback. I've submitted a commit with the requested changes. PTAL.

blueandgold:

PR #187, opened 2017-04-06
Make the api clients cleaner to create.

blueandgold:
This is in initial attempt to make the api clients cleaner to create.  I'd like to get some feedbacks before also doing this for the admin api module.

- move the quota flag to the api module
- i'm not sure if there really should be a get_rate_limiter static method; seems like we can do without it
- re-use the same api client, so that we can re-use the same rate limiter inside the client


blueandgold:

PR #189, opened 2017-04-07
Add groups raw json data.

blueandgold:
Per design, we need the raw api response json data for troubleshooting.


blueandgold:

PR #190, opened 2017-04-07
Remove some json manipulation in test.

blueandgold:
This makes the test & test data cleaner, by asserting on the actual output from the transform().


blueandgold:

PR #193, opened 2017-04-07
Break down the main method into smaller helper functions.

blueandgold:
I am going to defer unit tests in this PR, and chip away in later PRs.  But this should be okay for now, as this should be tested functionally when changes are made here.


blueandgold:

PR #196, opened 2017-04-11
Create group_members pipeline

mirons-google:
This pipeline builds on the previous groups work and creates a new table that stores an enumeration of each group member and its various properties.

The unique key for the table is the group_id (automatically generated by GSuite, not the email address as those can be changed).


google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -78,9 +78,35 @@ def _build_credentials(self):
...
+            groups: A group key, e.g. it's email address.
+
+        Returns:
+            A list of member objects from the API.
+
+        Raises:
+            api_errors.ApiExecutionError
+        """
+        members_stub = self.service.members()
+        request = members_stub.list(groupKey=group)
```
blueandgold: nit: can we make the name "group" into "group_key", such as:
```
request = members_stub.list(groupKey=group_key)
```

google/cloud/security/common/data_access/sql_queries/create_tables.py
```diff
@@ -79,7 +79,23 @@
...
+CREATE_GROUP_MEMBERS_TABLE = """
+    CREATE TABLE `{0}` (
+        `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
+        `group_id` varchar(255) DEFAULT NULL,
+        `member_kind` varchar(255) DEFAULT NULL,
+        `member_role` varchar(255) DEFAULT NULL,
+        `member_type` varchar(255) DEFAULT NULL,
+        `member_status` varchar(255) DEFAULT NULL,
+        `member_id` varchar(255) DEFAULT NULL,
+        `raw_member` json DEFAULT NULL,
```
blueandgold: In addition to this here, we also should have the entirety of the unparsed (i.e. raw) Member response stored in one big chunk in a separate table, with a 1:1 relation to Groups.  Please take a look at the RAW_PROJECT_IAM_POLICIES_TABLE.  Okay, to stick a TODO here and follow-up.

This is by design, as we might need it in case of troubleshooting our pipeline.

Storing this in a separate table is non-ideal, but it's a trade-off between the speed of loading to separate table, vs inserting into the Groups table.

mirons-google: done

google/cloud/security/common/gcp_api/admin_directory.py
mirons-google: done

mirons-google:
Thanks. I made the requested changes, PTAL.

blueandgold:

PR #203, opened 2017-04-11
Adding protection around building necessary dependences on groups inventorying

mirons-google:


google/cloud/security/inventory/inventory_loader.py
```diff
@@ -160,26 +161,37 @@ def _build_pipelines(cycle_timestamp, configs, dao):
...
+        if util.can_inventory_groups(configs):
+            admin_api_client = ad.AdminDirectoryClient()
+            pipelines.extend([
+                load_groups_pipeline.LoadGroupsPipeline(
+                    cycle_timestamp, configs, admin_api_client, dao),
+                load_group_members_pipeline.LoadGroupMembersPipeline(
+                    cycle_timestamp, configs, admin_api_client, dao)
+            ])
+        else:
+            raise inventory_errors.LoadDataPipelineError(
```
blueandgold: nit: if you are raising this here, then we need to add an error handler upstream for this; otherwise this will show up as a fatal exception to user

blueandgold: Or...do you just want to log it, and not raise an error?  This way, the pipeline can keep running?

mirons-google: To get to this point the user would have had to enabled the flag inventory_groups and then not specify the additionally required flags (can_inventory_groups).

It seems in main we do catch that exception as well.

blueandgold:

PR #194, opened 2017-04-10
Make sure the csv files are cleaned up.

blueandgold:
Fixes #164

PR #204, opened 2017-04-11
Fix csv file for scanner.

blueandgold:
This is to keep up with the changes from #194 where a file object is returned.


blueandgold:

PR #207, opened 2017-04-12
Added command line flag for log level

felixbb:
Command line flag addition + method to change the log level globally for all loggers already instantiated or instantiated in the future via log_util.get_logger.


google/cloud/security/inventory/inventory_loader.py
```diff
@@ -274,6 +286,13 @@ def _send_email(organization_id, cycle_time, cycle_timestamp, status, pipelines,
     except util_errors.EmailSendError:
         LOGGER.error('Unable to send email that inventory snapshot completed.')
 
+
+def _configure_logging(configs):
+    """Configures the loglevel for all loggers."""
+    desc = configs.get("loglevel")
+    level = LOGLEVELS[desc]
```
carise: I'm not too familiar with how the gflags enum works, but if someone gives the flag value something other than what's found in LOGLEVELS, will we get an error here (because desc isn't found in the LOGLEVELS dict)?

felixbb:
Addressed the double quotes, squashed into initial commit.

carise:
Looks good, one last comment -- I'm not too familiar with how the gflags enum works, but if someone gives the flag value something other than what's found in LOGLEVELS, will we get an error here (because desc isn't found in the LOGLEVELS dict)?

carise:
@felixbb quick ping on this thread to see if you had a chance to look at my comment regarding the loglevels? [link](https://github.com/GoogleCloudPlatform/forseti-security/pull/207#issuecomment-293740886)

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -274,6 +286,13 @@ def _send_email(organization_id, cycle_time, cycle_timestamp, status, pipelines,
     except util_errors.EmailSendError:
         LOGGER.error('Unable to send email that inventory snapshot completed.')
 
+
+def _configure_logging(configs):
+    """Configures the loglevel for all loggers."""
+    desc = configs.get('loglevel')
+    level = LOGLEVELS[desc]
```
gbrindisi: Hello, I can't push code to this PR but hopefully you can avoid a `KeyError`  here by using a default fallback value:

```
level = LOGLEVELS.setdefault(desc, 'info')
```

blueandgold:

PR #209, opened 2017-04-12
Use .get() in case no email

carise:
This can happen for an organization-wide group.
https://support.google.com/a/answer/66338?hl=en


blueandgold:
LGTM: I recommend that we also add a test data for a member without no email, and make sure this does not regress.
https://github.com/GoogleCloudPlatform/forseti-security/blob/master/tests/inventory/pipelines/test_data/fake_group_members.py

carise:
@blueandgold Thanks! I added some test data. https://github.com/GoogleCloudPlatform/forseti-security/pull/209/files#diff-fd7c233fb8f124283dcfc96268e5186bR36

blueandgold:

PR #210, opened 2017-04-12
Update load_groups_pipeline.py

mirons-google:
using .get() just in case we come across empty fields


blueandgold:

PR #214, opened 2017-04-12
Sample Rules File for Group Members

blueandgold:
This is the initial work for policy scanner to support group/members.  I am not sure what should be the resource id, so would love feedback on that.


blueandgold:

PR #216, opened 2017-04-12
Add git to apt-get

carise:


deployment-templates/py/forseti-instance.py
```diff
@@ -127,7 +127,7 @@ def GenerateConfig(context):
                 'items': [{
                     'key': 'startup-script',
                     'value': """#!/bin/bash
-sudo apt-get install -y unzip
+sudo apt-get install -y unzip git
```
blueandgold: Should we break this up into a separate line?

mirons-google: I was actually going to advocate putting them all on one line.

carise: I can do that too

blueandgold: All on one line sounds good....as long as we are consistent.  :)

blueandgold:

PR #217, opened 2017-04-12
Added pre-commit hook for local pylint checks

felixbb:
Will run pylint locally to speed up development cycles.


blueandgold:

PR #224, opened 2017-04-14
Cleanup projectdao

carise:
resolves #224


blueandgold:
LGTM: Super nice clean-up!  Glad that the raw json data can be re-used!

blueandgold:

PR #225, opened 2017-04-14
Scan/inventory resources that Forseti can see

### LINKED ISSUE #199: Scan/inventory all resources that the service account can see
Forseti will not be scoped to a single org in terms of visibility into its resources. Instead, allow Forseti to see all projects its service account can see, as well as orgs and their folders.

This needs to be done before #142 

- [x] Import all organizations that are visible to Forseti
- [x] Import all projects that are visible to Forseti (i.e. remove the project filter for parent type/id)
- [x] Make sure that the org IAM policies are imported for each org
- [x] Unit tests

carise:
resolves #199


google/cloud/security/common/data_access/sql_queries/select_data.py
```diff
@@ -31,6 +31,18 @@
     pol.member_domain, pol.member_name
 """
 
+ORGANIZATIONS = """SELECT org_id, name, display_name, lifecycle_state,
```
blueandgold: nit: do you think it's more readable if the select line is on it's own line? i.e. the statements are more lined up.

tests/common/data_access/organization_dao_test.py
```diff
@@ -0,0 +1,352 @@
...
+        """
+        conn_mock = mock.MagicMock()
+        cursor_mock = mock.MagicMock()
+
+        self.org_dao.conn = conn_mock
+        self.org_dao.conn.cursor.return_value = cursor_mock
+        cursor_mock.execute.side_effect = DataError
+        organization_dao.LOGGER = mock.MagicMock()
+
+        resource_name = 'organizations'
```
blueandgold: nit: these are used again, can this be factored out?

tests/common/data_access/organization_dao_test.py
```diff
@@ -0,0 +1,352 @@
...
+              * cursor
+              * fetch
+
+        Expect:
+            Raises a MySQLError.
+        """
+        conn_mock = mock.MagicMock()
+        cursor_mock = mock.MagicMock()
+
+        fake_org = [
```
blueandgold: ditto: on factoring out for re-use

tests/common/gcp_api/crm_test.py
```diff
@@ -169,6 +169,57 @@ def test_get_organization(self):
...
         self.assertEquals(1, crm.LOGGER.error.call_count)
 
+    def test_get_organizations(self):
+        """Test get organizations."""
+
+        mock_orgs_stub = mock.MagicMock()
+        self.crm_api_client.service = mock.MagicMock()
+        self.crm_api_client.service.organizations.return_value = mock_orgs_stub
+        
+        fake_orgs = {
```
blueandgold: nit: can we move this to the test data files?  1: to be consistent, 2: to avoid clutter in the test code here?

google/cloud/security/common/data_access/sql_queries/select_data.py
carise: Yes, I can certainly put it on its own line.

tests/common/data_access/organization_dao_test.py
carise: Yep -- does it make sense to put it on self.resource_name?

blueandgold: yup, that sounds good

blueandgold:

Issue #202, opened 2017-04-11
Scanner should support GSuite Google Groups inventories


mirons-google:
Groups (if enabled) are stored in groups_<timestamp> and group_members_<timestamp> accordingly.

blueandgold:
First PR #268 to support whitelisting.

blueandgold:
Whitelist support is in #268 and #275.  Blacklist and required members are still pending.

blueandgold:
Will wait until there are user needs for the blacklist & required members.

PR #226, opened 2017-04-14
Create the Initial Group Rules Engine

blueandgold:
This is the initial work to implement the Group Rules Engine.  What this PR does so far is the first step of creating the rules engine, create the resources, and build the rule book.

Other steps (getting the GCP group data, and finding violations) will follow.
#202


google/cloud/security/common/gcp_type/resource_util.py
```diff
@@ -42,6 +43,11 @@ class ResourceUtil(object):
             'class': project.Project,
             'plural': 'Projects',
             'level': 1
+        },
+        resource.ResourceType.GROUP: {
+            'class': group.Group,
+            'plural': 'Groups',
+            'level': 4
```
blueandgold: @carise , what is "level" here?  it looks sequential to me, and i just picked the next number

carise: I think that was the hierarchy level, but I don't recall if I am even using it. We can revisit it later.

google/cloud/security/scanner/samples/group_rules.yaml
```diff
@@ -0,0 +1,50 @@
+rules:
+  # Only allow Organization users to be in the Organization groups.
+  - name: sample whitelist for group users
+    mode: whitelist
+    resource:
+      - type: group
+        applies_to: self_and_children
+        resource_ids:
+          - my_customer
```
carise: Just a question -- do you build the group "hierarchy" and it all rolls up to "my_customer"?

blueandgold: Hmmm, I'm not sure how this works yet, downstream when we start to gather the gcp group data.  I assumed this was the initial entry point, to start gathering all the group.  So to that effect, I guess it is the "root" where everything rolls up to?

carise: sgtm

blueandgold:

PR #229, opened 2017-04-17
Remove running the group rules engine.

blueandgold:
This is still work in progress, so I think it's better not to expose this yet in the scanner runner.


blueandgold:

PR #230, opened 2017-04-17
More tests

carise:


google/cloud/security/scanner/scanner.py
```diff
@@ -195,12 +190,8 @@ def _get_project_policies(timestamp):
         The project policies.
     """
     project_policies = {}
-    try:
-        dao = project_dao.ProjectDao()
-        project_policies = dao.get_project_policies('projects', timestamp)
-    except MySQLError as err:
```
blueandgold: Hmmm, is this MySQLError being handled elsewhere now?  IIRC, I was working with this part of the code earlier, and this was the only place where the MySQLError would be caught when thrown by the dao.

carise: I think I tweaked the project_dao and organization_dao only to raise the MySQLError because I wanted the pipeline not to fail if an error was raised. (e.g. in a list of projects, do not raise an error because we still want to get as much data as we can) Thoughts?

blueandgold: I see the tweak now.  I guess i need to pull from master.  Regardless, this makes sense now.  👍

carise: Cool, thanks for the review!

blueandgold:

PR #231, opened 2017-04-17
Get the Group Users for the Group Rules Engine

blueandgold:
Some notes:
(0) Only getting group members of USER type, which I think is relevant to the defined rule.

(1) It seems better to select from the GroupMember table, rather than the raw GroupMember json because the select statement can be (more easily) targeted toward fetching ACTIVE, USER only, vs fetching json.

(2) Not clear how to handle the hierarchy or parenting just yet.  So, leaving a TODO in place.  I'm sure this will become obvious in the next step (when doing traversals to find violations).

(3) Tried subclassing the group dao from the (base) dao.  It works, and the code is cleaner.  But there is still much more to be done here, or replaced by ORM.

(4) No tests here just yet, as this is still much in flux.  Will make sure there is test coverage before wrapping up work here.

### DIFF
--- a/google/cloud/security/common/data_access/group_dao.py
+++ b/google/cloud/security/common/data_access/group_dao.py
@@ -1,0 +1,55 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Provides the data access object (DAO) for Groups."""
+
+from google.cloud.security.common.data_access import dao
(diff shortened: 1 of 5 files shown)


google/cloud/security/common/gcp_type/group_member.py
```diff
@@ -24,22 +24,23 @@
...
-        self.type = rule_def_member.get('type')
-        self.email = rule_def_member.get('email')
+        self.member_role = member_role
+        self.member_type = member_type
+        self.member_email = member_email
 
-        if (not self.role or
-                not self.type or
-                not self.email):
+        if (not self.member_role or
```
carise: nit: trailing whitespace

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -0,0 +1,55 @@
...
+from google.cloud.security.common.data_access import dao
+from google.cloud.security.common.data_access.sql_queries import select_data
+from google.cloud.security.common.gcp_type import group_member
+from google.cloud.security.common.util import log_util
+
+LOGGER = log_util.get_logger(__name__)
+
+
+class GroupDao(dao.Dao):
+    """Data access object (DAO) for Organizations."""
```
blueandgold: Done.  Good catch.

google/cloud/security/common/gcp_type/group_member.py
blueandgold: Done.

blueandgold:

PR #232, opened 2017-04-18
Untangle member_id and member_email

blueandgold:
The member_id and member_email are 2 different fields.  However, member_email was put into the member_id field.


blueandgold:

PR #233, opened 2017-04-18
Folders inventory

carise:
Initial folders groundwork. Lots of small changes, mostly for names (e.g. make getters a `@property` in Resource class)

1. Fix the Resource class to have more standardized properties (to match the GCP API)
2. _dao classes are subclasses of Dao.
3. Change inventory_loader to use the appropriate dao classes in the pipelines (e.g. for project* pipelines, use the ProjectDao)
4. Some more unit tests for folders.

### DIFF
--- a/google/cloud/security/common/data_access/project_dao.py
+++ b/google/cloud/security/common/data_access/project_dao.py
@@ -59,7 +59,7 @@ class ProjectDao(dao.Dao):
             return [row['project_number'] for row in rows]
         except (DataError, IntegrityError, InternalError, NotSupportedError,
                 OperationalError, ProgrammingError) as e:
-            raise MySQLError(resource_name, e)
+            raise errors.MySQLError(resource_name, e)
 
 
     def get_project_policies(self, resource_name, timestamp):
--- a/google/cloud/security/common/data_access/project_dao.py
+++ b/google/cloud/security/common/data_access/project_dao.py
@@ -23,7 +23,7 @@ from MySQLdb import NotSupportedError
 from MySQLdb import OperationalError
 from MySQLdb import ProgrammingError
 
-from google.cloud.security.common.data_access import _db_connector
+from google.cloud.security.common.data_access import dao
 from google.cloud.security.common.data_access import errors
 from google.cloud.security.common.data_access.sql_queries import select_data
 from google.cloud.security.common.gcp_type import project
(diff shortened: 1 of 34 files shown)


google/cloud/security/common/data_access/project_dao.py
```diff
@@ -33,15 +33,41 @@
...
+
+        Raises:
+            MySQLError: An error with MySQL has occurred.
+        """
+        try:
+            project_numbers_sql = select_data.PROJECT_NUMBERS.format(timestamp)
+            rows = self.execute_sql_with_fetch(
+                resource_name, project_numbers_sql, ())
+            return [row['project_number'] for row in rows]
+        except (DataError, IntegrityError, InternalError, NotSupportedError,
```
blueandgold: You can actually do without the try/except as self.execute_sql_with_fetch() will already do that, and raise MySQLError, which you can handle here or let it bubble up and handled higher up.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -284,7 +292,12 @@ def main(_):
     configs = FLAGS.FlagValuesDict()
 
     try:
-        pipelines = _build_pipelines(cycle_timestamp, configs, dao)
+        pipelines = _build_pipelines(
+            cycle_timestamp,
+            configs,
+            dao=dao,
+            project_dao=project_dao,
+            organization_dao=organization_dao)
```
blueandgold: nit: Since you are already doing this, do you also want to do the same clean-up for group_dao here?  It's okay if rather not.  :)

google/cloud/security/inventory/pipelines/load_projects_iam_policies_pipeline.py
```diff
@@ -99,7 +99,7 @@ def _retrieve(self):
         """
         # Get the projects for which we will retrieve the IAM policies.
         try:
-            project_numbers = self.dao.select_project_numbers(
+            project_numbers = self.dao.get_project_numbers(
```
blueandgold: Would this get_project_numbers() better belong to the project_dao?  In other words, would dao become the base_dao?  It's okay to follow-up on this later.

google/cloud/security/inventory/inventory_loader.py
carise: I don't mind :) But am wondering if you've already done the work? I don't want to step on your toes there.

google/cloud/security/common/data_access/project_dao.py
carise: Good tip -- I was thinking about changing it and just left it, but I will go ahead and change it.

google/cloud/security/inventory/pipelines/load_projects_iam_policies_pipeline.py
carise: Yes. get_project_numbers() has been moved to project_dao and I also pass the project_dao to the constructor for LoadProjectsIamPoliciesPipeline. I guess it's not so clear here, sorry :|

google/cloud/security/inventory/inventory_loader.py
blueandgold: Nope I have not.  Please knock it out!  :)

carise: Ok, I'll do that tomorrow.

carise:
@blueandgold PTAL on this commit: https://github.com/GoogleCloudPlatform/forseti-security/pull/233/commits/39026772d72132d3086fa98e36d2db5204cf4670

blueandgold:
Looks great!!!  :+1:

carise:
@blueandgold thanks for the review!

blueandgold:

PR #235, opened 2017-04-19
Scanner unit tests

carise:


google/cloud/security/scanner/scanner.py
```diff
@@ -225,7 +228,12 @@ def _output_results(all_violations, **kwargs):
 
         # If output_path specified, upload to GCS.
         if FLAGS.output_path:
-            _upload_csv_to_gcs(FLAGS.output_path, now_utc, csv_file.name)
+            output_path = FLAGS.output_path
+            if not output_path.startswith('gs://'):
+                if not os.path.exists(FLAGS.output_path):
+                    os.makedirs(output_path)
+                output_path = os.path.abspath(output_path)
+            _upload_csv_to_gcs(output_path, now_utc, csv_file.name)
```
blueandgold: nit: I think this method should be better named as "_upload_csv()", since this method seems to do both.
By renaming, the code block here will also make a little more sense, i.e. not passing a non-gcs output_path to _upload_csv_to_gcs()

carise: Done!

blueandgold:

PR #239, opened 2017-04-22
Import scanner violations into Cloud SQL

### LINKED ISSUE #63: Scanner should write violations to CloudSQL
* e.g., What role, member, project, and the required action

carise:
resolves #63

### DIFF
--- a/google/cloud/security/common/data_access/dao.py
+++ b/google/cloud/security/common/data_access/dao.py
@@ -41,6 +41,8 @@ CREATE_TABLE_MAP = {
     'raw_project_iam_policies':
         create_tables.CREATE_RAW_PROJECT_IAM_POLICIES_TABLE,
     'raw_org_iam_policies': create_tables.CREATE_RAW_ORG_IAM_POLICIES_TABLE,
+
+    'violations': create_tables.CREATE_VIOLATIONS_TABLE,
 }
 
 SNAPSHOT_FILTER_CLAUSE = ' where status in ({})'
@@ -48,9 +50,6 @@ SNAPSHOT_FILTER_CLAUSE = ' where status in ({})'
 
 class Dao(_db_connector.DbConnector):
     """Data access object (DAO)."""
-
-    def __init__(self):
-        super(Dao, self).__init__()
 
     def _create_snapshot_table(self, resource_name, timestamp):
         """Creates a snapshot table.
--- a/google/cloud/security/common/data_access/sql_queries/create_tables.py
+++ b/google/cloud/security/common/data_access/sql_queries/create_tables.py
@@ -20,8 +20,8 @@ CREATE_PROJECT_TABLE = """
         `project_number` bigint(20) NOT NULL,
         `project_id` varchar(255) NOT NULL,
         `project_name` varchar(255) DEFAULT NULL,
-        `lifecycle_state` enum('ACTIVE','DELETE_REQUESTED',
-            'DELETE_IN_PROGRESS','DELETED') DEFAULT NULL,
+        `lifecycle_state` enum('LIFECYCLE_STATE_UNSPECIFIED','ACTIVE',
+            'DELETE_REQUESTED','DELETED') NOT NULL,
         `parent_type` varchar(255) DEFAULT NULL,
         `parent_id` varchar(255) DEFAULT NULL,
         `raw_project` json DEFAULT NULL,
@@ -58,8 +58,8 @@ CREATE_ORGANIZATIONS_TABLE = """
         `org_id` bigint(20) unsigned NOT NULL,
         `name` varchar(255) NOT NULL,
         `display_name` varchar(255) DEFAULT NULL,
-        `lifecycle_state` enum('ACTIVE','DELETE_REQUESTED',
-            'DELETED','LIFECYCLE_STATE_UNSPECIFIED') DEFAULT NULL,
+        `lifecycle_state` enum('LIFECYCLE_STATE_UNSPECIFIED','ACTIVE',
+            'DELETE_REQUESTED', 'DELETED') NOT NULL,
         `raw_org` json DEFAULT NULL,
         `creation_time` datetime DEFAULT NULL,
         PRIMARY KEY (`org_id`)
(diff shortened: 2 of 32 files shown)


google/cloud/security/common/data_access/dao.py
```diff
@@ -41,6 +41,8 @@
     'raw_project_iam_policies':
         create_tables.CREATE_RAW_PROJECT_IAM_POLICIES_TABLE,
     'raw_org_iam_policies': create_tables.CREATE_RAW_ORG_IAM_POLICIES_TABLE,
+
+    'violations': create_tables.CREATE_VIOLATIONS_TABLE,
```
mirons-google: nit: would it be worth adding a prefix to the table to denote that it's scanner doing this?

google/cloud/security/common/data_access/sql_queries/create_tables.py
```diff
@@ -20,8 +20,8 @@
         `project_number` bigint(20) NOT NULL,
         `project_id` varchar(255) NOT NULL,
         `project_name` varchar(255) DEFAULT NULL,
-        `lifecycle_state` enum('ACTIVE','DELETE_REQUESTED',
-            'DELETE_IN_PROGRESS','DELETED') DEFAULT NULL,
+        `lifecycle_state` enum('LIFECYCLE_STATE_UNSPECIFIED','ACTIVE',
```
mirons-google: nit:  It might be worth defining these elsewhere as a python-type?

google/cloud/security/common/data_access/violation_dao.py
```diff
@@ -0,0 +1,115 @@
...
+                self.RESOURCE_NAME, snapshot_timestamp)
+        except MySQLdb.Error, e:
+            raise db_errors.MySQLError(self.RESOURCE_NAME, e)
+
+        try:
+            # Insert the violations to the table in one single transaction.
+            self.conn.autocommit(False)
+            cursor = self.conn.cursor()
+            for violation in violations:
+                for formatted_violation in _format_violation(violation):
```
mirons-google: The loop of violations and formatted_violations can occur outside the try

google/cloud/security/common/data_access/violation_dao.py
```diff
@@ -0,0 +1,115 @@
...
+    Args:
+        violation: The RuleViolation.
+
+    Returns:
+        A tuple of the rule violation properties.
+    """
+
+    resource_type = violation.resource_type
+    if resource_type:
+        resource_type = resource_type[:255]
```
mirons-google: Since 255 seems like a magic number can you add a comment of what's contained in resource_type and what you expect after slicing? Please do this throughout.

google/cloud/security/common/data_access/violation_dao.py
```diff
@@ -0,0 +1,115 @@
...
+def _format_violation(violation):
+    """Format the violation data into a tuple.
+
+    Also flattens the RuleViolation, since it consists of the resource,
+    rule, and members that don't meet the rule criteria.
+
+    Args:
+        violation: The RuleViolation.
+
+    Returns:
```
mirons-google: nit: Yields

google/cloud/security/common/data_access/violation_dao.py
```diff
@@ -0,0 +1,115 @@
...
+
+            # Create the violations snapshot table.
+            snapshot_table = self._create_snapshot_table(
+                self.RESOURCE_NAME, snapshot_timestamp)
+        except MySQLdb.Error, e:
+            raise db_errors.MySQLError(self.RESOURCE_NAME, e)
+
+        try:
+            # Insert the violations to the table in one single transaction.
+            self.conn.autocommit(False)
```
blueandgold: Hmmm, can you please explain the use of autocommit()?  By default it is disabled.
https://dev.mysql.com/doc/connector-python/en/connector-python-api-mysqlconnection-autocommit.html

google/cloud/security/common/data_access/violation_dao.py
```diff
@@ -0,0 +1,115 @@
...
+            cursor = self.conn.cursor()
+            for violation in violations:
+                for formatted_violation in _format_violation(violation):
+                    cursor.execute(
```
blueandgold: qq: Is there a reason why dao.execute_sql_with_commit() is not used here?  It can save the try/except here.

google/cloud/security/common/data_access/violation_dao.py
```diff
@@ -0,0 +1,115 @@
...
+
+LOGGER = log_util.get_logger(__name__)
+
+
+class ViolationDao(dao.Dao):
+    """Data access object (DAO) for rule violations."""
+
+    RESOURCE_NAME = 'violations'
+
+    def import_violations(self, violations, snapshot_timestamp=None):
```
blueandgold: nit: Can we change this method naming to "insert_violations", to be more aligned and descriptive of the sql operation?  It would be so much clearer when it's used elsewhere.

google/cloud/security/inventory/pipelines/load_projects_iam_policies_pipeline.py
```diff
@@ -36,21 +36,6 @@ class LoadProjectsIamPoliciesPipeline(base_pipeline.BasePipeline):
...
-        Args:
-            cycle_timestamp: String of timestamp, formatted as YYYYMMDDTHHMMSSZ.
-            configs: Dictionary of configurations.
-            crm_client: CRM API client.
-            dao: Data access object.
-
-        Returns:
-            None
-        """
-        super(LoadProjectsIamPoliciesPipeline, self).__init__(
```
blueandgold: Why is this being removed?  Is this not used for the base pipeline, and later referenced here as self.dao ?
https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/inventory/pipelines/base_pipeline.py

carise: Doesn't load_projects_iam_policies automatically call its parent's __init__() upon instantiation?

google/cloud/security/common/data_access/sql_queries/create_tables.py
carise: Will this suffice? https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/gcp_type/resource.py#L59

mirons-google: Yup!

google/cloud/security/scanner/scanner.py
```diff
@@ -227,18 +230,27 @@ def _output_results(all_violations, **kwargs):
...
-        vdao.import_violations(all_violations, kwargs.get('snapshot_timestamp'))
+        (inserted_rows, violation_errors) = vdao.insert_violations(
+            all_violations, snapshot_timestamp=snapshot_timestamp)
     except db_errors.MySQLError as err:
         LOGGER.error('Error importing violations to database: %s', err)
 
-    # Write the CSV.
+    # TODO: figure out what to do with the errors. For now, just log it.
+    LOGGER.debug('Inserted %s rows with %s errors',
+                 inserted_rows, len(violation_errors))
```
blueandgold: super small nit: "inserted_rows" is an number, rather than actual "row"; so it would read better if this is "inserted_row_count" or "insertion_count")

carise: Thanks, I updated the name to `inserted_row_count`.

blueandgold:

PR #243, opened 2017-04-26
Batch insert group members

gbrindisi:
Hi there,
in an effort to make the `group_members` pipeline more fast/resilient I've implemented a batch retrieve/commit mechanism. 
(This is slightly related to https://github.com/GoogleCloudPlatform/forseti-security/issues/241.)

It is very useful in our environment given our number of groups.

The modification is two fold:
* in the pipeline, group_ids are chunked and processed in batch: this means both querying and storing results.
* in the DAO, the `load_data` method now supports loading data into an already existing table. This part is not great, I pretty much did a quick and dirty exception based control flow. Can you suggest improvements?

Issue #241, opened 2017-04-25
Best Effort Inventory Building


gbrindisi:
Hello,
was wondering if it would make sense to make the inventory building more resilient.

For example sometimes I get errors when querying the APIs and an exception will be raised on the pipeline. The whole snapshot cycle is then labeled as `PARTIAL_SUCCESS` and the pipeline is considered done.

This is mildly annoying when operating on a big pipeline like the `LoadGroupMembersPipeline`: our rationale is that a partial inventory is better than an empty one.

A recent example:

```
2017-04-24 16:43:19,789 __main__     DEBUG    _run_pipelines Running pipeline <google.cloud.security.inventory.pipelines.load_group_members_pipeline.LoadGroupMembersPipeline object at 0x10a2d1490>
2017-04-24 16:44:48,776 __main__     ERROR    _run_pipelines Encountered error loading data.
GCP API Error: unable to get <googleapiclient.discovery.Resource object at 0x10a9cb590> from GCP:
<HttpError 503 when requesting https://www.googleapis.com/admin/directory/v1/groups/XYZ/members?alt=json&pageToken=XXX returned "Service unavailable. Please try again">
2017-04-24 16:44:48,776 __main__     INFO     _run_pipelines Continuing on.
2017-04-24 16:44:48,852 __main__     INFO     _complete_snapshot_cycle Inventory load cycle completed with PARTIAL_SUCCESS: 20170424T143633Z
```

The exception handling is in `inventory_loader.py`:

```
    for pipeline in pipelines:
        try:
            LOGGER.debug('Running pipeline %s', pipeline)
            pipeline.run()
            pipeline.status = 'SUCCESS'
        except inventory_errors.LoadDataPipelineError as e:
            LOGGER.error('Encountered error loading data.\n%s', e)
            pipeline.status = 'FAILURE'
            LOGGER.info('Continuing on.')
        run_statuses.append(pipeline.status == 'SUCCESS')
    return run_statuses
```

Could it be possible to collect the errors received, without raising an exception and try to keep going with the pipeline?
Eventually add the registered errors in a dedicated column within the `snapshot_cycles` table, for audit purposes.

For example:

```
+----+---------------------+---------------------+-----------------+----------------+------------------+------------------+
| id | start_time          | complete_time       | status          | schema_version | cycle_timestamp  | errors           |
+----+---------------------+---------------------+-----------------+----------------+------------------+------------------+
[...]

PR #244, opened 2017-04-26
Admin API maxResult

gbrindisi:
as requested in https://github.com/GoogleCloudPlatform/forseti-security/pull/243

### DIFF
--- a/google/cloud/security/common/gcp_api/admin_directory.py
+++ b/google/cloud/security/common/gcp_api/admin_directory.py
@@ -36,6 +36,8 @@ flags.DEFINE_string('groups_service_account_key_file', None,
                     'runnning locally.')
 flags.DEFINE_integer('max_admin_api_calls_per_day', 150000,
                      'Admin SDK queries per day.')
+flags.DEFINE_integer('max_results_admin_api', 500,
++                     'maxResult param for the Admin SDK list() method')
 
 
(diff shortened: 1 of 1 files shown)


google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -36,6 +36,8 @@
                     'runnning locally.')
 flags.DEFINE_integer('max_admin_api_calls_per_day', 150000,
                      'Admin SDK queries per day.')
+flags.DEFINE_integer('max_results_admin_api', 500,
++                     'maxResult param for the Admin SDK list() method')
```
blueandgold: nit: align the leftmost quotes

blueandgold: Thanks for making this a configurable flag!

blueandgold:
LGTM.  This PR is to alleviate the 503 issue in #241.  It is good, however, I just want to point out that the root cause of repeated list_next() calls might still crop up, especially for larger domains, and worth investigating.  I'll open a separate issue to track this.

blueandgold:
Filed issue #245 for the 503.

blueandgold:
Adding `cla: yes` label, as the PR author is covered by corporate CLA.

PR #246, opened 2017-04-26
Rename OrgRule* to "IamRule*"

carise:
This is the first step in trying to make the "Rules Engine" modular. The current OrgRulesEngine is actually specifically for IAM rules. Thus, the module name and class name should reflect that.


blueandgold:

PR #248, opened 2017-05-01
Add unit test for group member chunks.

blueandgold:
Also address the comments from PR #243 .

### DIFF
--- a/google/cloud/security/inventory/pipelines/load_group_members_pipeline.py
+++ b/google/cloud/security/inventory/pipelines/load_group_members_pipeline.py
@@ -87,7 +87,7 @@ class LoadGroupMembersPipeline(base_pipeline.BasePipeline):
                        'member_email': member.get('email'),
                        'raw_member': json.dumps(member)}
 
-    def _retrieve(self, group_ids):
+    def _retrieve(self, group_ids):  # pylint: disable=arguments-differ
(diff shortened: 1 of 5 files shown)


google/cloud/security/inventory/pipelines/load_group_members_pipeline.py
```diff
@@ -127,13 +128,17 @@ def chunker(seq, size):
             return (seq[pos:pos + size] for pos in xrange(0, len(seq), size))
 
         # TODO: keep track of group_ids that are not retrieved/committed to db
-        for group_ids_chunk in chunker(group_ids, GROUP_CHUNK_SIZE):
-            LOGGER.debug('Retrieving a batch of group members')
-            groups_members_map = self._retrieve(group_ids_chunk)
+        chunk_counter = 0
+        for group_ids_in_chunk in chunker(group_ids, self.GROUP_CHUNK_SIZE):
```
mirons-google: nit: Fine to leave this as a TODO, but the chunker thing seems generic enough that we could move it to g/c/s/c/gcp_api/common.py, thus letting all the other pipelines use it.

blueandgold: Done.  Good point about moving this to common.

blueandgold:

PR #240, opened 2017-04-24
Support for inventorying GCS buckets for projects

thenenadx:
Support for pulling bucket data into Inventory

PR #250, opened 2017-05-02
Improvement on loading rules engines more dynamically.

thenenadx:
This change requires that rules engines have a build_rule_book() function.


thenenadx:
Seems like tests are failing for me :) I will fix that before proceeding with this

blueandgold:
Adding `cla: yes` label, as the PR author has ack'd signing the CLA in PR #240.

PR #256, opened 2017-05-02
Modernize our Travis CI env

mirons-google:
- remove the need for sudo by installing libmysqlclient-dev through the addons option
- use the new trusty image (better machine/build performance)
- consolidate pip installs


blueandgold:

PR #258, opened 2017-05-03
Scanner engine improvementsv2

thenenadx:
* Refactoring of the scanner to support loading of different rules
* Modified scanner tests
* Some tests have been disabled because they need to be rewritten by a bit


google/cloud/security/scanner/scanner.py
```diff
@@ -84,60 +121,112 @@ def main(_):
...
-    # TODO: make this generic
-    org_policies = _get_org_policies(snapshot_timestamp)
-    project_policies = _get_project_policies(snapshot_timestamp)
+    # Load pipeline from mapping file
+    engine_map = pc.get_engine_info(rules_engine_name)
 
-    if not org_policies and not project_policies:
-        LOGGER.warn('No policies found. Exiting.')
-        sys.exit()
+    data_pipeline_class, _ = _create_rules_engine(
```
blueandgold: As noted above, perhaps we can create a map and not need the  ```_create_rules_engine()```?

For example, we can do something like this:
```python

scanner_map  = {
    'IamPolicy': scanners.IamPolicyScanner,
    'Foo': scanners.FooScanner,
    'Bar':  scanners.BarScanner,
  }

# So, that we can do something like this:
scanner = scanner_map['IamPolicy'](snapshot_timestamp)
```

google/cloud/security/scanner/scanner.py
```diff
@@ -69,13 +69,50 @@
...
+        _rules_engine_list(audit_base_dir)
+
+    if not FLAGS.use_engine:
+        LOGGER.warn('Provide an engine file')
+        sys.exit(1)
+    else:
+        rules_engine_filename = FLAGS.use_engine
+
+    # Load rules engine according to command line flag
+    rules_engine_class, rules_engine_name = _create_rules_engine(
```
blueandgold: Hmmm, perhaps we can create a map and not need the _create_rules_engine()?

For example:
```python
ENGINE_TO_DATA_MAP =  {
    'IamRulesEngine': audit.IamRulesEngine,
    'FooRulesEngine': audit.FooRulesEngine,
    'BarRulesEngine': audit.BarRulesEngine,
}

# So, that we can do something like this:
rules_engine = ENGINE_TO_DATA_MAP['IamRulesEngine'](rules_file_path=FLAGS.rules)
```

google/cloud/security/scanner/pipelines/base_data_pipeline.py
```diff
@@ -0,0 +1,47 @@
...
+"""Base data pipeline skeleton."""
+
+import abc
+
+from google.cloud.security.common.util import log_util
+
+LOGGER = log_util.get_logger(__name__)
+
+
+class BaseDataPipeline(object):
```
blueandgold: Pending carise's comment, but perhaps we can have more appropriate naming here?

Instead of ```pipeline```, how about ```scanner```? 
And should use just drop ```data```?
How does this look?
```
google/cloud/security/scanner/scanners
google/cloud/security/scanner/scanners/base_scanner.py
google/cloud/security/scanner/scanners/iam_rules_scanner.py
```

google/cloud/security/scanner/pipelines/load_iam_rules_engine_pipeline.py
```diff
@@ -0,0 +1,125 @@
...
+from google.cloud.security.common.util import log_util
+from google.cloud.security.common.data_access import organization_dao
+from google.cloud.security.common.data_access import project_dao
+from google.cloud.security.common.gcp_type.resource import ResourceType
+from google.cloud.security.scanner.pipelines import base_data_pipeline
+
+LOGGER = log_util.get_logger(__name__)
+
+
+class LoadIamDataPipeline(base_data_pipeline.BaseDataPipeline):
```
blueandgold: nit: naming :)  See my comment above for proposal.

google/cloud/security/scanner/scanner.py
```diff
@@ -69,13 +69,50 @@
                      'the format of the path should be '
                      '"gs://bucket-name/path/for/output".'))
 
+flags.DEFINE_bool('list_engines', False, 'List all rule engines')
+
+flags.DEFINE_string('use_engine', None, 'Which engine to use')
```
thenenadx: Yep, so either it should be scripted to go through all the engines or we can improve it even more to go through the whole map.

thenenadx:
@blueandgold The latest commit addresses all your feedback, except the one from the comment from above :) It's very easy to add if we feel like going through all the rules

tests/scanner/scanner_test.py
```diff
@@ -32,6 +32,7 @@
 from google.cloud.security.scanner import scanner
 from google.cloud.security.scanner.audit import iam_rules_engine as ire
 from google.cloud.security.scanner.audit import rules as audit_rules
+from google.cloud.security.scanner.scanners import iam_rules_scanner as irs
```
carise: should we separate out the individual scanner tests into its own test suite and keep the scanner runner tests here?

thenenadx: Yep, that was my plan as part of fixing all tests, especially the ones I've disabled.

blueandgold:
Adding `cla: yes` label, as the PR author has ack'd signing the CLA in PR # 240.

Issue #257, opened 2017-05-03
Create a utility method for paging through list_next actions.


mirons-google:
This is repeated [here](https://github.com/GoogleCloudPlatform/forseti-security/blob/94a2003524deb41cdb74377bedbeb20f5b532133/google/cloud/security/common/gcp_api/cloud_resource_manager.py#L98) and [here](https://github.com/GoogleCloudPlatform/forseti-security/blob/6ef792584d71602422d1af02b47bba174116b719/google/cloud/security/common/gcp_api/admin_directory.py#L104)

PR #259, opened 2017-05-03
Abstracting paging through results in GCP API requests

mirons-google:
I'm open to making the change to cloud_resource_manager.py if you have a suggestion, but as it seems to be a different implementation I punted.

### DIFF
--- a/google/cloud/security/common/gcp_api/_base_client.py
+++ b/google/cloud/security/common/gcp_api/_base_client.py
@@ -75,6 +75,8 @@ class BaseClient(object):
     def _build_paged_result(self, request, api_stub, rate_limiter):
         """Execute results and page through the results.
 
+        Use of this method requires the API having a .list_next() method.
+
         Args:
             request: GCP API client request object.
             api_stub: The API stub used to build the request.
@@ -87,9 +89,16 @@ class BaseClient(object):
             When the retry is exceeded, exception will be thrown.  This
             exception is not wrapped by the retry library, and will be handled
             upstream.
+
+            UnsupportedApiMethodError when the api_stub does not have the required
+            list_next() method.
         """
+        if not hasattr(api_stub, 'list_next')
+            raise api_errors.UnsupportedApiMethodError
+
         results = []
         response = []
+
         while request is not None:
             try:
                 with rate_limiter:
--- a/google/cloud/security/common/gcp_api/_base_client.py
+++ b/google/cloud/security/common/gcp_api/_base_client.py
@@ -93,7 +93,7 @@ class BaseClient(object):
             try:
                 with rate_limiter:
                     response = self._execute(request)
-                    results.extend(response)
+                    results.append(response)
                     request = api_stub.list_next(request, response)
             except (HttpError, HttpLib2Error) as e:
                 raise api_errors.ApiExecutionError(api_stub, e)
--- a/google/cloud/security/common/gcp_api/_base_client.py
+++ b/google/cloud/security/common/gcp_api/_base_client.py
@@ -81,7 +81,7 @@ class BaseClient(object):
             rate_limiter: An instance of RateLimiter to use.
 
         Returns:
-            API response object.
+            A list of API response objects (dict).
 
         Raises:
             When the retry is exceeded, exception will be thrown.  This
@@ -89,11 +89,12 @@ class BaseClient(object):
             upstream.
         """
         results = []
+        response = []
         while request is not None:
             try:
                 with rate_limiter:
-                    response = self._execute(request)
-                    results.append(response)
+                    response.append(self._execute(request))
+                    results.extend(response)
                     request = api_stub.list_next(request, response)
             except (HttpError, HttpLib2Error) as e:
                 raise api_errors.ApiExecutionError(api_stub, e)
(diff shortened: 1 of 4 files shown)


mirons-google:
This resolves #257

google/cloud/security/common/gcp_api/_base_client.py
```diff
@@ -69,3 +71,31 @@ def _execute(self, request):
...
+            When the retry is exceeded, exception will be thrown.  This
+            exception is not wrapped by the retry library, and will be handled
+            upstream.
+        """
+        results = []
+        while request is not None:
+            try:
+                with rate_limiter:
+                    response = self._execute(request)
+                    results.append(response)
```
blueandgold: QQ: previously, this was using ```.extend()```, which merged the page results into one list, rather than ```.append()``` which would create separate page results inside a list, IIRC.

google/cloud/security/common/gcp_api/_base_client.py
```diff
@@ -69,3 +71,31 @@ def _execute(self, request):
...
+                    request = api_stub.list_next(request, response)
```
blueandgold: QQ: will every api support ```.list_next()```?  For example, looking at the crm api, a pageToken is passed repeatedly into ```.list()```
https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/list

google/cloud/security/common/gcp_api/_base_client.py
mirons-google: Ah, fixed.

google/cloud/security/common/gcp_api/_base_client.py
mirons-google: I believe so. Because we're using the python api client it provides these [methods](https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1/python/latest/cloudresourcemanager_v1.projects.html#list_next) for us.

mirons-google:
Thanks. PTAL.

google/cloud/security/common/gcp_api/_base_client.py
blueandgold: Okay.  :)  Looks like the api doc is out of sync.  But to ask another QQ on this.  I look at the other instance methods of the doc, does it matter if list_next() are not there, like for folder and organizations?
https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1/python/latest/cloudresourcemanager_v1.folders.html

https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1/python/latest/cloudresourcemanager_v1.organizations.html

mirons-google:
Those not existing doesn't effect  this PR. It will change whether or not our api code can use this list_next() method.

google/cloud/security/common/gcp_api/_base_client.py
```diff
@@ -69,3 +71,32 @@ def _execute(self, request):
...
+            upstream.
+        """
+        results = []
+        response = []
+        while request is not None:
+            try:
+                with rate_limiter:
+                    response.append(self._execute(request))
+                    results.extend(response)
+                    request = api_stub.list_next(request, response)
```
blueandgold: nit: It would be nice if we can add a comment here or in the method doc, to note that user should verify that list_next() does exist for the api being used.

mirons-google:
I improved this. PTAL.

google/cloud/security/common/gcp_api/_base_client.py
```diff
@@ -69,3 +71,41 @@ def _execute(self, request):
...
+        Raises:
+            When the retry is exceeded, exception will be thrown.  This
+            exception is not wrapped by the retry library, and will be handled
+            upstream.
+
+            UnsupportedApiMethodError when the api_stub does not have the required
+            list_next() method.
+        """
+        if not hasattr(api_stub, 'list_next')
+            raise api_errors.UnsupportedApiMethodError
```
blueandgold: A couple of small comments:
1: It might be nice to just raise ```api_errors.ApiExecutionError``` with the proper error message
```
     raise api_errors.ApiExecutionError(api_stub, 'No list_next() in API.)
```
2: by reusing ```api_errors.ApiExecutionError```, you don't have to catch it in admin_directory.py, and can just let this error bubble up to foo_pipeline.py, where it will be caught and handled.

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -132,16 +125,11 @@ def get_groups(self, customer_id='my_customer'):
...
-                    request = groups_stub.list_next(request, response)
-            except (HttpError, HttpLib2Error) as e:
-                raise api_errors.ApiExecutionError(groups_stub, e)
-
-        return results_by_group
+
+        try:
+            results = self._build_paged_result(
+                request, groups_stub, self.rate_limiter)
+        except api_errors.UnsupportedApiError:
```
blueandgold: See my comment above.  Also, by doing this here, I don't think that a specific message about unsupported api would surface.

google/cloud/security/common/gcp_api/_base_client.py
mirons-google: Good call, fixed.

google/cloud/security/common/gcp_api/admin_directory.py
mirons-google: Fixed.

mirons-google:
PTAL.

google/cloud/security/common/gcp_api/_base_client.py
mirons-google: done!

blueandgold:

blueandgold:

PR #260, opened 2017-05-04
Fix page results by merging them.

blueandgold:
The paged results need to merged.  Otherwise, the transform() in the pipelines will be borked.
```
[[page result 1], [page result 2], etc.]
```

### DIFF
--- a/google/cloud/security/common/gcp_api/admin_directory.py
+++ b/google/cloud/security/common/gcp_api/admin_directory.py
@@ -129,6 +129,8 @@ class AdminDirectoryClient(_base_client.BaseClient):
         paged_results = self._build_paged_result(
             request, groups_stub, self.rate_limiter)
 
+        # TODO: Refactor this to a helper function, where a key string can
+        # be passed in, to extract the right values to merge.
         groups = []
         for page in paged_results:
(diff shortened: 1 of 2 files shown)


google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -97,10 +97,13 @@ def get_group_members(self, group_key):
         request = members_stub.list(groupKey=group_key,
                                     maxResults=FLAGS.max_results_admin_api)
 
-        results = self._build_paged_result(
+        paged_results = self._build_paged_result(
             request, members_stub, self.rate_limiter)
 
-        return [item.get('members', []) for item in results]
+        group_members = []
```
mirons-google: If you're up for fixing this once, if not just open an issue, but we should write another helper method to un-bundle the results of a paged request.

It'd accept just one argument, the string of the key and do what you're doing on lines 103-106

blueandgold: Okay, I added a TODO and issue #263 for this, as I think it should go in the _base_client.py, and will be a bigger change than can be done quickly here.  Also, we should write tests for this.  :)

blueandgold:

PR #261, opened 2017-05-04
Get all recursive members of a group.

blueandgold:
This will be used for the group members scanner: when evaluating a rule for a specific group, the rule will need to be applied to all recursive members.

The most important item to look at is the ```get_recursive_members_of_group()```.  Other changes are purging items that I don't think will be needed.

### DIFF
--- a/google/cloud/security/common/data_access/group_dao.py
+++ b/google/cloud/security/common/data_access/group_dao.py
@@ -26,18 +26,18 @@ LOGGER = log_util.get_logger(__name__)
 class GroupDao(dao.Dao):
     """Data access object (DAO) for Groups."""
 
-    def get_group_id(self, resource_name, email, timestamp):
-        """Get the id of a group from an email.
+    def get_group_id(self, resource_name, group_email, timestamp):
+        """Get the group_id for the specified group_email.
 
         Args:
             resource_name: String of the resource name.
-            group_id: String of the group id.
+            group_email: String of the group email.
             timestamp: The timestamp of the snapshot.
 
         Returns:
              String of the group id.
         """
-        sql = select_data.GROUP_ID.format(email, timestamp)
+        sql = select_data.GROUP_ID.format(group_email, timestamp)
         result = self.execute_sql_with_fetch(resource_name, sql, None)
         return result[0].get('group_id')
 
--- a/google/cloud/security/common/data_access/group_dao.py
+++ b/google/cloud/security/common/data_access/group_dao.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 """Provides the data access object (DAO) for Groups."""
 
 from Queue import Queue
@@ -77,8 +78,7 @@ class GroupDao(dao.Dao):
         all_members = []
         queue = Queue()
 
-        group_id = self.get_group_id(
-            'group', group_email, timestamp)
+        group_id = self.get_group_id('group', group_email, timestamp)
         queue.put(group_id)
 
         while not queue.empty():
--- a/google/cloud/security/common/data_access/group_dao.py
+++ b/google/cloud/security/common/data_access/group_dao.py
@@ -37,8 +37,8 @@ class GroupDao(dao.Dao):
         Returns:
              String of the group id.
         """
-        sql = select_data.GROUP_ID.format(group_email, timestamp)
-        result = self.execute_sql_with_fetch(resource_name, sql, None)
+        sql = select_data.GROUP_ID.format(timestamp)
+        result = self.execute_sql_with_fetch(resource_name, sql, (group_email,))
         return result[0].get('group_id')
 
     def get_group_members(self, resource_name, group_id, timestamp):
@@ -57,8 +57,8 @@ class GroupDao(dao.Dao):
                'member_role': 'OWNER',
                'member_type': 'USER'}, ...)
         """
-        sql = select_data.GROUP_MEMBERS.format(group_id, timestamp)
-        return self.execute_sql_with_fetch(resource_name, sql, None)
+        sql = select_data.GROUP_MEMBERS.format(timestamp)
+        return self.execute_sql_with_fetch(resource_name, sql, (group_id,))
 
     def get_recursive_members_of_group(self, group_email, timestamp):
         """Get all the recursive members of a group.
--- a/google/cloud/security/common/data_access/group_dao.py
+++ b/google/cloud/security/common/data_access/group_dao.py
@@ -11,12 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 """Provides the data access object (DAO) for Groups."""
+
+from Queue import Queue
 
 from google.cloud.security.common.data_access import dao
 from google.cloud.security.common.data_access.sql_queries import select_data
-from google.cloud.security.common.gcp_type import group_member
 from google.cloud.security.common.util import log_util
 
 LOGGER = log_util.get_logger(__name__)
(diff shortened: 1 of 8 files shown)


google/cloud/security/common/data_access/group_dao.py
```diff
@@ -11,12 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
```
mirons-google: nit: I think we've been (in)consistent in leaving a blank line here.

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
+        Args:
+            resource_name: String of the resource name.
+            group_id: String of the group id.
+            timestamp: The timestamp of the snapshot.
+
+        Returns:
+             String of the group id.
+        """
+        sql = select_data.GROUP_ID.format(email, timestamp)
+        result = self.execute_sql_with_fetch(resource_name, sql, None)
```
mirons-google: nit: passing in none here is confusing and I've seen this a couple times. Can we just give execute_sql_with_fetch a default for that kwarg.

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
+        return result[0].get('group_id')
```
mirons-google: nit: [0] seems like a magic item in the list, can we make this more readable?

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
+             [{'group_id': '00lnxb',
+               'member_email': 'foo@mygbiz.com',
+               'member_id': '11111',
+               'member_role': 'OWNER',
+               'member_type': 'USER'}, ...]
+        """
+        all_members = []
+        queue = Queue()
+
+        group_id = self.get_group_id(
```
mirons-google: nit: can this fit on one line?

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
+            'group', group_email, timestamp)
+        queue.put(group_id)
+
+        while not queue.empty():
+            group_id = queue.get()
+            members = self.get_group_members('group_members', group_id,
```
mirons-google: nit: 'group_members' can we reference this type somewhere that's statically defined?

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
+                                             timestamp)
+            for member in members:
+                all_members.append(member)
+                if member.get('member_type') == 'GROUP':
```
mirons-google: you're calling member.get('member_type') twice here seems logical to stash this.

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
-                row.get('member_email'))
-            if row.get('group_id') not in group_users:
-                group_users[row.get('group_id')] = [group_user]
-            else:
-                group_users[row.get('group_id')].append(group_user)
-        return group_users
+        sql = select_data.GROUP_MEMBERS.format(group_id, timestamp)
+        return self.execute_sql_with_fetch(resource_name, sql, None)
+
+    def get_recursive_members_of_group(self, group_email, timestamp):
```
mirons-google: nit: Naming would be more clear with 'get_all_members_of_group'

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
+    def get_group_id(self, resource_name, email, timestamp):
+        """Get the id of a group from an email.
+
+        Args:
+            resource_name: String of the resource name.
+            group_id: String of the group id.
+            timestamp: The timestamp of the snapshot.
+
+        Returns:
+             String of the group id.
```
mirons-google: nit: can you provide an example of a group_id?

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
     """Data access object (DAO) for Groups."""
 
-    def get_group_users(self, resource_name, timestamp):
-        """Get the group members who are users.
+    def get_group_id(self, resource_name, email, timestamp):
+        """Get the id of a group from an email.
+
+        Args:
+            resource_name: String of the resource name.
+            group_id: String of the group id.
```
mirons-google: This should be s/group_id/email/

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
 class GroupDao(dao.Dao):
     """Data access object (DAO) for Groups."""
 
-    def get_group_users(self, resource_name, timestamp):
-        """Get the group members who are users.
+    def get_group_id(self, resource_name, email, timestamp):
```
mirons-google: nit: s/email/group_email/

This is because it's stored that way in the database.

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -25,28 +25,68 @@
...
+        """Get the id of a group from an email.
```
mirons-google: nit: Get the group_id for the specified group_email.

tests/common/data_access/test_data/fake_group_dao_data.py
```diff
@@ -0,0 +1,65 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Test data for group dao tests."""
+
+GET_GROUP_MEMBERS_SIDE_EFFECT = [
```
mirons-google: Didn't we generate some similar test data for admin_directory_test.py we can re-use?

google/cloud/security/common/data_access/group_dao.py
blueandgold: Done

google/cloud/security/common/data_access/group_dao.py
blueandgold: Done

google/cloud/security/common/data_access/group_dao.py
blueandgold: Yah, the more I look at the passing of these resource names (for capturing in the error message), the more I want to get rid of them.  Opening issue #264 to do it separately.

google/cloud/security/common/data_access/group_dao.py
blueandgold: Hmmm, I think it's only used once here?

tests/common/data_access/test_data/fake_group_dao_data.py
blueandgold: Hmmm, there is no admin_directory_test.py, or at least not where I think it should be. The other test data for groups that we have is the fake_group.py and fake_group_members.py, and they are not structured to test for recursive membership, which the data here are.  And it kind of makes more sense to keep them here, closer to where it's being used in the group_dao.py.

google/cloud/security/common/data_access/group_dao.py
blueandgold: Done

google/cloud/security/common/data_access/sql_queries/select_data.py
```diff
@@ -65,10 +65,16 @@
...
+    FROM groups_{1}
+    WHERE group_email = "{0}";
+"""
+
+GROUP_MEMBERS = """
     SELECT group_id, member_role, member_type, member_id, member_email
-    FROM group_members_{0}
-    WHERE member_status = "ACTIVE" and member_type = "USER";
+    FROM group_members_{1}
+    WHERE group_id = "{0}";
```
carise: should the group_id = "{0}" be 

group_id = %s

You'd still keep the {} for the group_members_{} for the tablename but make the filter clause parameterized.

blueandgold: Hmmm, I am not sure I understand.  Can you give me an example?
One my part, this is how I'm currently formatting the sql query.  So, positionally, I need to explicitly use {0} for group_id, and {2} for the timestamp.
```
sql = select_data.GROUP_MEMBERS.format(group_id, timestamp)
```

google/cloud/security/common/data_access/group_dao.py
mirons-google: You're correct, disregard.

google/cloud/security/common/data_access/sql_queries/select_data.py
carise: Correct, you would still use the .format() for the timestamp. However, format does not protect against sql injection, so you'd want to do something like this:

```
sql = select_data.GROUP_MEMBERS.format(timestamp)
cursor.execute(sql, group_id)
```

where GROUP_MEMBERS is:

```
GROUP_MEMBERS = """
      SELECT group_id, member_role, member_type, member_id, member_email		      
      FROM group_members_{}
      WHERE group_id = %s;
```

(more information: http://bobby-tables.com/python, http://stackoverflow.com/questions/7929364/python-best-practice-and-securest-to-connect-to-mysql-and-execute-queries)

blueandgold:
Awesome, done, fixed the string formating in the sql statement.

carise:
Cool! 💯

blueandgold:

PR #268, opened 2017-05-08
Initial Groups Scanner.

blueandgold:
This is the initial implementation of the groups scanner.  Please read the following notes:

- Data are organized as nodes in a tree structure.  This is so that we can know how to apply the rules in a hierarchy.  i.e. foo rule applies to foo node, but bar rule applies to bar node.
- Only whitelist mode is supported for this PR.  Blacklist and required modes are left as TODOs.
- There are some deviations from the established rules structure, I'm trying out using conditions.
- The only inheritance that seems sensible for group members is self_and_children, and no inherit_from_parent option.
- Making separate execution path in the scanner.py, since the group scanner works and returns differently.  Also, no output yet.  Will clean this up.
- I have tested this functionally, but have not written actual tests.  I prefer to wait for this PR to be reviewed first as things might change.
- There are also a lot of stuff to delete from earlier work.  I will delete that later, so not to muddle this PR more than is necessary.

### DIFF
--- a/google/cloud/security/common/data_access/group_dao.py
+++ b/google/cloud/security/common/data_access/group_dao.py
@@ -11,10 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from anytree.node import NodeMixin
 
 """Provides the data access object (DAO) for Groups."""
 
 from Queue import Queue
+
+from anytree import Node
+from anytree import RenderTree
+from anytree import AsciiStyle
 
 from google.cloud.security.common.data_access import dao
 from google.cloud.security.common.data_access.sql_queries import select_data
@@ -25,6 +30,13 @@ LOGGER = log_util.get_logger(__name__)
 
 class GroupDao(dao.Dao):
     """Data access object (DAO) for Groups."""
+
+    def get_all_groups(self, resource_name, timestamp):
+        sql = select_data.GROUPS.format(timestamp)
+        result = self.execute_sql_with_fetch(resource_name, sql, None)
+        return result
+
+
 
     def get_group_id(self, resource_name, group_email, timestamp):
         """Get the group_id for the specified group_email.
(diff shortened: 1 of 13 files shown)


google/cloud/security/scanner/scanners/groups_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+            node.rules.append(rule)
+        return root
+
+    def run(self):
+        """Runs the groups scanner."""
+
+        dao = group_dao.GroupDao()
+        root = dao.build_group_tree(self.snapshot_timestamp)
+
+        rules_path = 'google/cloud/security/scanner/samples/group_rules.yaml'
```
carise: Did you intend to hardcode the local file path for the rules?

google/cloud/security/scanner/scanners/groups_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+                elif rule.get('mode') == 'required':
+                    pass  # TODO
+
+                else:
+                    pass  # TODO
+    
+            # Determine if the node is in violations or not.
+            # All rules must be fulfilled, for a node to not be in violation.
+            # If any rule is not fulfilled, then node is in violation.
+            #
```
carise: I think this is really a nit, but the truth table is not necessary, given that someone who is writing python can just refer to the python official documentation for explanation of any()/all().

google/cloud/security/scanner/samples/group_rules.yaml
```diff
@@ -1,50 +1,21 @@
...
-          - my_customer
-    inherit_from_parents: true
-    members:
-      - role: MEMBER
-        type: USER
-        email: '*@YOUR_ORG_EMAIL_DOMAIN'
+- name: Allow my company users and gmail users to be in my company groups.
+  group_email: my_customer
+  mode: whitelist
+  conditions:
```
carise: Nit: I know what conditions means because we discussed this offline already, but for someone new to the scanner, it may not be clear what exactly is the condition (i.e. a condition implies something like an "if" statement, you're checking if something eq/ne/lte/gte etc.)

Suggestion is to use some kind of stronger indicator (or property name) to imply that you're checking equality

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -90,3 +108,73 @@ def get_recursive_members_of_group(self, group_email, timestamp):
...
+        while not queue.empty():
+            node = queue.get()
+            members = self.get_group_members('group_members', node.member_id,
+                                             timestamp)
+
+            for member in members:
+                member_node = MemberNode(member.get('member_id'),
+                                         member.get('member_email'),
+                                         member.get('member_type'),
+                                         member.get('member_satus'),
```
carise: do you mean "member_status"?

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -16,6 +16,11 @@
 
 from Queue import Queue
 
+from anytree import Node
```
carise: nit: 

```
import anytree as at
```

and use at.Node, at.RenderTree, etc.

google/cloud/security/common/data_access/group_dao.py
```diff
@@ -90,3 +108,73 @@ def get_recursive_members_of_group(self, group_email, timestamp):
...
+
+        Args:
+            timestamp: String of snapshot timestamp, formatted as
+                YYYYMMDDTHHMMSSZ.
+
+        Returns:
+            The root node that holds the tree structure of all the groups
+                in the organization.
+        """
+        root = MemberNode('my_customer', 'my_customer', None, None, None)
```
carise: Should you make 'my_customer' a constant somewhere and reference that wherever it's being used?

carise: Also, is there a reason not to change the MemberNode default arguments values to None such that you're instead calling MemberNode('my_customer', 'my_customer') instead of the current one with the Nones?

google/cloud/security/scanner/scanners/groups_scanner.py
```diff
@@ -0,0 +1,168 @@
...
+            # +-----------------------------------------+---------+---------+
+            # | All Falsy values                        |  False  |  False  |
+            # +-----------------------------------------+---------+---------+
+            # | One Truthy value (all others are Falsy) |  True   |  False  |
+            # +-----------------------------------------+---------+---------+
+            # | One Falsy value (all others are Truthy) |  True   |  False  |
+            # +-----------------------------------------+---------+---------+
+            # | Empty Iterable                          |  False  |  True   |
+            # +-----------------------------------------+---------+---------+
+            if any(whitelist_rule_statuses) is False:
```
carise: Is there a reason you chose "if any(...) is False" as opposed to "if not any(...)" ?

google/cloud/security/scanner/scanners/groups_scanner.py
blueandgold: Good catch.  Using the flag now.

google/cloud/security/scanner/scanners/groups_scanner.py
blueandgold: Yah, using ```if any(...) is False``` just seemed easier to read.  :D  But changing this to ```if not any(...)```

google/cloud/security/scanner/samples/group_rules.yaml
blueandgold: Adding a TODO, as this kind of change will also required being baked ito the code.

blueandgold:

PR #269, opened 2017-05-09
Remove files that will not be used for the groups scanner.

blueandgold:
These were created for the previous approach to create a groups scanner, but are not needed anymore for the existing approach.


blueandgold:

PR #275, opened 2017-05-11
Unit tests for groups scanner.

blueandgold:
This is my first time to unit test trees, so I would appreciate any suggestions to do this more easily or better.

The current approach involves:
1: take advantage of python's pickle library to stash a tree object in a file, which can later be re-loaded for testing
2: take advantage of anytree's ability to print out an ascii representation of a tree, which can be used for assertions
3: this saves some amount of work to write test codes to build or traverse the tree
4: it would be helpful to look at ```fake_gorups_scanner_data.py```, how the test data are setup, which will explain how the tests work


blueandgold:

PR #276, opened 2017-05-11
Fix data format

carise:


carise:
@mirons-google done, PTAL

blueandgold:

PR #283, opened 2017-05-15
Fix scanner inheritance

carise:
This fixes the IAM rules scanner to find org resource parents via database lookup.

Folder rules will be handled in a followup PR.


carise:
Thanks for the review!

blueandgold:

PR #286, opened 2017-05-16
Reorganization of our samples.

mirons-google:


samples/deployment-manager/deploy-forseti.yaml.sample
```diff
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 # The toplevel deployment script for Forseti Security.
-# Be sure to edit this file and its templates to suit your environment.
 
+# Change these import paths as necessary to suit your environement.
```
carise: Users will need to edit properties for the compute instance. Should we include verbiage for that?

mirons-google:
Done. PTAL.

blueandgold:

PR #291, opened 2017-05-22
More organization of samples and scripts

mirons-google:


mirons-google:
Note: tests are failing because we changed the location of protoc_url.txt. This will be fixed once the new path is reflected in `master`.

blueandgold:

PR #292, opened 2017-05-22
Deployment template cleanup (including imports)

mirons-google:


samples/deployment-manager/deploy-forseti.yaml.sample
```diff
@@ -17,13 +17,13 @@
 
 # Change these import paths as necessary to suit your environement.
 imports:
-- path: py/inventory/cloudsql-instance.py
+- path: ../../deployment-templates/inventory/cloudsql-instance.py
```
blueandgold: QQ: What is the consideration to use relative paths like this?  How about absolute path?  Would something like this work?
```
<root path>/deployment-templates/inventory/cloudsql-instance.py
```

mirons-google: Deployment manager doesn't like absolute path. I suspect because it doesn't know the root to start looking from.

blueandgold:

PR #297, opened 2017-05-22
pylint improvements

mirons-google:
1. Updated version on Travis
2. Fixed overridden _transform throughout (pylint warning otherwise)
3. Tweaked the travis pylint script


google/cloud/security/inventory/pipelines/load_firewall_rules_pipeline.py
```diff
@@ -30,14 +30,14 @@ class LoadFirewallRulesPipeline(base_pipeline.BasePipeline):
 
     RESOURCE_NAME = 'firewall_rules'
 
-    def _transform(self, firewall_rules_map):
+    def _transform(self, resource_from_api):
```
blueandgold: nit: Is it necessary have "resource_from_api" here?  I realize it's from the base pipeline, but that was a generic placeholder there, because that is an abstract class.  Here in the concrete classes, it's nicer if parameter can reflect the actual resource that is passed in.

mirons-google:
This issue is that pylint complains with a W0221 if the argument isn't named exactly as the base class.

blueandgold:
Okay, cool.... pylint sometimes really goes overboard.  :)

blueandgold:

PR #302, opened 2017-05-24
Make inventory pipelines to be configurable.

blueandgold:
Sending this out for initial review.  Will polish it and add test at later step.

- This make inventory pipelines configurable via an external yaml file.
- A dependency tree will be built.
- Organizations pipeline is root, and will always be run.
- The main rule is:
```
If child pipeline is true, then all parent pipelines will become true.
Even if the parent(s) is(are) configured as false in the yaml.
```
```
organizations
|-- folders
|-- groups
|   +-- group_members
|-- org_iam_policies
+-- projects
    |-- bigquery_datasets
    |-- firewall_rules
    |-- forwarding_rules
    |-- buckets
    |   +-- buckets_acls
    |-- cloudsql
    +-- projects_iam_policies
True
|-- False
|-- True
|   +-- True
|-- True
+-- True
    |-- True
    |-- False
    |-- True
    |-- True
    |   +-- True
    |-- True
    +-- True
```

They dynamic module loading is adopted from spotify, but is tweaked to reduce code.  i.e. the module name has to be maintained in the yaml.  (This is similar to appengine's approach.)  I think there is a way to maintain outside of this config file, if this is not desired.


google/cloud/security/inventory/inventory_loader.py
```diff
@@ -191,98 +191,56 @@ def _build_pipelines(cycle_timestamp, configs, **kwargs):
...
-            pipelines.extend([
-                load_groups_pipeline.LoadGroupsPipeline(
-                    cycle_timestamp,
-                    configs,
-                    admin_api_client,
-                    dao),
-                load_group_members_pipeline.LoadGroupMembersPipeline(
+
+    inventory_pipeline_configs = file_loader.read_and_parse_file(
+        'google/cloud/security/inventory/inventory_pipeline_configs.yaml')
```
carise: Just a note, though you may already be aware -- you might need to indicate the yaml as a data file in setup.py, if forseti_inventory can't find the yaml.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -191,98 +191,56 @@ def _build_pipelines(cycle_timestamp, configs, **kwargs):
...
+    # correctly on all the nodes
+    for i in inventory_pipeline_configs:
+        parent_name = i.get('depends_on')
+        if parent_name is not None:
+            parent_node = map_of_all_pipeline_nodes[parent_name]
+            map_of_all_pipeline_nodes[i.get('resource')].parent = parent_node
+
+    root = map_of_all_pipeline_nodes.get('organizations')
+
+    # If child pipeline is true, then all parents will become true.
```
carise: This is nice, thanks for the explanation 👍

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -191,98 +191,56 @@ def _build_pipelines(cycle_timestamp, configs, **kwargs):
...
                     cycle_timestamp,
                     configs,
-                    admin_api_client,
-                    dao)
-            ])
-        else:
-            raise inventory_errors.LoadDataPipelineError(
-                'Unable to inventory groups with specified arguments:\n%s',
-                configs)
+                    crm_v1_api_client,
```
carise: I think I might've missed something but it's not clear to me how each pipeline knows which GCP API client or dao class to use? (e.g. organization pipeline uses organization_dao, project pipeline uses project_dao, etc.)

google/cloud/security/inventory/inventory_pipeline_configs.yaml
```diff
@@ -0,0 +1,66 @@
+
+- resource: bigquery_datasets
+  module_name: load_bigquery_datasets_pipeline
+  depends_on: projects
```
mirons-google: Can we avoid putting the onus of knowing the dependency on the user creating a config and store this statically somewhere? We'll know better than the user.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -18,7 +18,7 @@
 
 Usage:
   $ forseti_inventory \\
-      --inventory_groups <true|false> (optional) \\
+      --pipeline_config (required) \\
       --groups_service_account_key_file <path to file> (optional)\\
```
carise: Does it make sense to put the groups account keyfile/super admin email in the pipeline config, like in a "metadata" property?

e.g.

```
- resource: groups
   should_run: false
   service_acct_keyfile: path/to/keyfile
   domain_superadmin_email: superadmin@domain.local
```

blueandgold: @mirons-google What do you think about this?  I am okay with it.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +242,67 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
...
-        folder_dao = folder_resource_dao.FolderDao()
+        api_map = {
+            'admin_api': admin_directory.AdminDirectoryClient(),
+            'bigquery_api': bigquery.BigQueryClient(),
+            'cloudsql_api': cloudsql.CloudsqlClient(),
+            'compute_api': compute.ComputeClient(),
+            'compute_beta_api': compute.ComputeClient(version='beta'),
+            'crm_api': crm.CloudResourceManagerClient(),
+            'crm_v2beta1': crm.CloudResourceManagerClient(version='v2beta1'),
+            'gcs_api': gcs.StorageClient(),
```
carise: Does it make sense to eventually (not necessarily for this PR) make inventory loader not instantiate daos and clients before running the pipelines, and move the dao/api client instantiation to the pipelines? As inventory loader becomes more of a pure "loader"/"runner" module, it seems like it should not know about the internals of the pipelines and just pass configuration details to each pipeline.

google/cloud/security/inventory/inventory_loader.py
mirons-google: That's fine. We'll need to adjust the GCP deployment script to account for this though.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -96,6 +78,10 @@
 }
 flags.DEFINE_enum('loglevel', 'info', LOGLEVELS.keys(), 'Loglevel.')
 
+flags.DEFINE_string('pipeline_configs', None,
```
mirons-google: nit: Naming the flag -config is much cleaner.

mirons-google: Can we also default it to <root>/config/inventory.yaml?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +242,68 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
...
-        project_dao = proj_dao.ProjectDao()
-        organization_dao = org_dao.OrganizationDao()
-        bucket_dao = buck_dao.BucketDao()
-        cloudsql_dao = sql_dao.CloudsqlDao()
-        fwd_rules_dao = fr_dao.ForwardingRulesDao()
-        folder_dao = folder_resource_dao.FolderDao()
-    except data_access_errors.MySQLError as e:
-        LOGGER.error('Encountered error with Cloud SQL. Abort.\n%s', e)
+
+    flag_configs = FLAGS.FlagValuesDict()
```
mirons-google: nit: s/flag_configs/flags/

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +242,68 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
...
+    _configure_logging(flag_configs)
```
mirons-google: Can we change this function to just take the loglevel flag?

_configure_logging(flags.get('loglevel'))

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -423,17 +316,18 @@ def main(_):
     else:
         snapshot_cycle_status = 'FAILURE'
 
-    _complete_snapshot_cycle(dao, cycle_timestamp, snapshot_cycle_status)
+    _complete_snapshot_cycle(dao_map.get('dao'), cycle_timestamp,
+                             snapshot_cycle_status)
 
-    if configs.get('email_recipient') is not None:
+    if flag_configs.get('email_recipient') is not None:
```
mirons-google: nit: can you factor this out of main()?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +242,68 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
...
+            'forwarding_rules_dao': forwarding_rules_dao.ForwardingRulesDao(),
+            'organization_dao': organization_dao.OrganizationDao(),
+            'project_dao': project_dao.ProjectDao(),
+        }
+    except data_access_errors.MySQLError as e:
+        LOGGER.error('Encountered error to create dao map.\n%s', e)
+        sys.exit()
 
-    _configure_logging(configs)
+    cycle_time, cycle_timestamp = _start_snapshot_cycle(dao_map.get('dao'))
```
mirons-google: nit: you fetch 'dao' a few times from dao_map below, want to store it in a variable?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +242,68 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
...
+    _configure_logging(flag_configs)
+
+    if FLAGS.pipeline_configs is None:
+        LOGGER.error('Path to pipeline config needs to be specified.')
         sys.exit()
 
-    cycle_time, cycle_timestamp = _start_snapshot_cycle(dao)
+    # Make these re-usable so that the rate-limiter can apply across
+    # different pipelines.
+    try:
```
mirons-google: nit: can you factor this out of main()?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +242,68 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
...
+    # A lot of potential errors can be thrown by different APIs.
+    # So, handle generically.
+    except:
+        LOGGER.error('Error to create the api map.\n%s', sys.exc_info()[0])
+        sys.exit()
 
-    configs = FLAGS.FlagValuesDict()
+    # Make these re-usable so that the db connection can apply across
+    # different pipelines.
+    try:
```
mirons-google: nit: can you factor this out of main()?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +242,68 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
...
+            'cloudsql_api': cloudsql.CloudsqlClient(),
+            'compute_api': compute.ComputeClient(),
+            'compute_beta_api': compute.ComputeClient(version='beta'),
+            'crm_api': crm.CloudResourceManagerClient(),
+            'crm_v2beta1': crm.CloudResourceManagerClient(version='v2beta1'),
+            'gcs_api': gcs.StorageClient(),
+        }
+    # A lot of potential errors can be thrown by different APIs.
+    # So, handle generically.
+    except:
```
mirons-google: We should avoid this naked except. We control the underlying exceptions that can be raised here. In the least we should leave a TODO to come back to this very soon because this is against style guidelines.

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,241 @@
...
+        self.flag_configs = flag_configs
+        self.api_map = api_map
+        self.dao_map = dao_map
+
+    def _find_runnable_pipelines(self, root):
+        """Initialize the pipeline builder.
+
+        Args:
+            root: PipelineNode representing the top-level starting point
+                of the pipeline dependency tree.  The entire pipeline
```
mirons-google: nit: one space after the period, fix throughout

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,241 @@
...
+                dependency tree are children of this root.
+
+        Returns:
+            runnable_pipelines: List of the pipelines that will be run.  The
+                order in list represents the order they need to be run.
```
mirons-google: ..order in the list..

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,241 @@
...
+            pipeline_configs)
+        self.flag_configs = flag_configs
+        self.api_map = api_map
+        self.dao_map = dao_map
+
+    def _find_runnable_pipelines(self, root):
+        """Initialize the pipeline builder.
+
+        Args:
+            root: PipelineNode representing the top-level starting point
```
mirons-google: Do you have an example you can provide in the doc here of what a sample 'root' might be?

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,241 @@
...
+            pipeline_configs: String of the path to the pipeline configs file.
+            flag_configs: Dictionary of flag values.
+            api_map: Dictionary of GCP API instances, mapped to each resource.
+            dao_map: Dictionary of DAO instances, mapped to each resource.
+
+        Returns:
+            None
+        """
+        self.cycle_timestamp = cycle_timestamp
+        self.pipeline_configs = file_loader.read_and_parse_file(
```
mirons-google: nit: there's technically only one 'config' the plural isn't accurate. the use of pipeline is also somewhat redundant.

s/pipeline_configs/config/
s/flag_configs/flags/

google/cloud/security/inventory/pipeline_configs.yaml
```diff
@@ -0,0 +1,40 @@
+
```
mirons-google: nit: We need the LICENSE here too.
nit: I'd like to store default configs for the core-modules in /config/inventory.(conf|yaml) for example.

google/cloud/security/inventory/pipeline_configs.yaml
```diff
@@ -0,0 +1,40 @@
...
+  should_run: true
+
+- resource: groups
+  should_run: false
+
+- resource: org_iam_policies
+  should_run: true
+
+- resource: organizations
+  should_run: true
```
mirons-google: nit: s/should_run/enabled/

google/cloud/security/inventory/pipeline_configs.yaml
```diff
@@ -0,0 +1,40 @@
+
+- resource: bigquery_datasets
```
mirons-google: As a user how do I know where these resource values come from? Can we add a comment?

google/cloud/security/inventory/pipeline_configs.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+- resource: forwarding_rules
+  should_run: true
+
+- resource: group_members
+  should_run: true
+
+- resource: groups
+  should_run: false
+
```
mirons-google: If this is also going to be our default config. Should we just comment out false resources (but also change them to enabled)

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,241 @@
...
+from google.cloud.security.common.util import file_loader
+from google.cloud.security.common.util import log_util
+
+
+LOGGER = log_util.get_logger(__name__)
+
+
+class PipelineBuilder(object):
+    """Inventory Pipeline Builder."""
+
```
mirons-google: To my earlier comment regarding how a user knows about what resources are available. We could create a flag for inventory that (--list_resources) which prints the a list of resources that you store here? That would be super helpful.

google/cloud/security/inventory/pipeline_configs.yaml
blueandgold: Done

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Done

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Done

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Done.  For further reference, they can follow the PipelineNode class at the bottom.  I'll add a link to the anytree documentation at the class.

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Adding a TODO here.

google/cloud/security/inventory/inventory_loader.py
blueandgold: The problem that we are solving by instantiating these outside of the pipelines is to re-use the rate-limiters and the db connectors across different pipelines.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Done.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Done.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Done.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Then, let's do this in another PR.  The proposed change should happen in conjunction with the deployment script, and in isolation of this PR.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Honestly, I think that accessing it from dictionary multiple times is okay.  Saves a variable.  :)

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Done.

blueandgold:
PTAL, unit tests are added.  I will have to discuss with Carise, the best way to strongly couple the class names.

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -18,7 +18,7 @@
 
 Usage:
   $ forseti_inventory \\
-      --inventory_groups <true|false> (optional) \\
```
mirons-google: Can we add a TODO to move all of these flags to the config?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -124,25 +110,25 @@ def _exists_snapshot_cycles_table(dao):
 
     return False
 
-def _create_snapshot_cycles_table(dao):
+def _create_snapshot_cycles_table(dao_):
```
mirons-google: why dao_ ?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +242,70 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
         LOGGER.error('Unable to send email that inventory snapshot completed.')
 
 
-def _configure_logging(configs):
+def _configure_logging(loglevel):
     """Configures the loglevel for all loggers."""
-    desc = configs.get('loglevel')
-    level = LOGLEVELS.setdefault(desc, 'info')
+    level = LOGLEVELS.setdefault(loglevel, 'info')
```
mirons-google: Don't we have a FLAG for this configuration?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -423,17 +318,17 @@ def main(_):
     else:
         snapshot_cycle_status = 'FAILURE'
 
-    _complete_snapshot_cycle(dao, cycle_timestamp, snapshot_cycle_status)
+    _complete_snapshot_cycle(dao_map.get('dao'), cycle_timestamp,
+                             snapshot_cycle_status)
 
-    if configs.get('email_recipient') is not None:
+    if forseti_flags.get('email_recipient') is not None:
```
mirons-google: nit: s/forseti_flags/inventory_flags/

google/cloud/security/inventory/pipeline_builder.py
mirons-google: For documentation purposes the flag will be in inventory_loader.py and it'll read in and print all the keys() in REQUIREMENTS_MAP

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,277 @@
...
+    def _find_runnable_pipelines(self, root):
+        """Initialize the pipeline builder.
+
+        Args:
+            root: PipelineNode representing the top-level starting point
+                of the pipeline dependency tree. The entire pipeline
+                dependency tree are tuple of children PipelineNodes
+                of this root.
+                Example:
```
mirons-google: nit: new line before the example

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,277 @@
...
+                except (ImportError, TypeError, ValueError) as e:
+                    LOGGER.error('Unable to import %s\n%s', module_name, e)
+                    continue
+
+                class_name = (
+                    self.REQUIREMENTS_MAP
+                    .get(node.resource_name)
+                    .get('module_name')
+                    .title()
+                    .replace('_', ''))
```
mirons-google: It's not clear to me what you're doing with class_name here and why we need to replace the underscores? Shouldn't they just match? A comment would be useful here.

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,277 @@
...
+        # First pass: map all the pipelines to their own nodes,
+        # regardless if they should run or not.
+        map_of_all_pipeline_nodes = {}
+        for i in self.config:
+            map_of_all_pipeline_nodes[i.get('resource')] = PipelineNode(
+                i.get('resource'), i.get('enabled'))
+
+        # Another pass: build the dependency tree by setting the parents
+        # correctly on all the nodes.
+        for i in self.config:
```
mirons-google: nit: avoid using 'i' if you can here.

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,277 @@
...
+        return map_of_all_pipeline_nodes.get('organizations')
+
+    def build(self):
+        """Build the pipelines to load data.
+
+        Returns:
+            List of pipelines instances that will be run.
+        """
+        root = self._build_dependency_tree()
+        return self._find_runnable_pipelines(root)
```
mirons-google: nit: newline before the return

google/cloud/security/inventory/inventory_loader.py
blueandgold: Done

google/cloud/security/inventory/inventory_loader.py
blueandgold: This is to fix lint's complaint about using "dao" locally here, would rewrite the global "dao" from the import.  Let me change this to "inventory_dao" to be consistent with the flags.  (same problem there, too; local "flags" would overwrite the imported "flags")

google/cloud/security/inventory/inventory_loader.py
blueandgold: Hmmm, I was just following the existing code pattern.  :)  Honestly, I was wondering about the entire usage of FLAGS.FlagValuesDict() that bundles the flags into a dictionary.  Unless we change that, we should stick with using it.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Done, much better name indeed!

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Done

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Done

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Good point, adding comment & example.

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Done

google/cloud/security/inventory/pipeline_builder.py
blueandgold: Done

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -32,62 +32,46 @@
       --email_recipient <email address of the email recipient> (optional) \\
       --loglevel <debug|info|warning|error>
 
+TODO: Move all these flags to the config file.
```
mirons-google: nit: We'll need to put the todo before the pydoc because gflags uses the pydoc for --helpfull _I think_

google/cloud/security/inventory/inventory_loader.py
mirons-google: Do you plan on still doing this in the PR?

google/cloud/security/inventory/inventory_loader.py
mirons-google: Do you plan on still doing this in the PR?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -372,45 +244,70 @@ def _send_email(cycle_time, cycle_timestamp, status, pipelines,
...
+        }
+    except data_access_errors.MySQLError as e:
+        LOGGER.error('Encountered error to create dao map.\n%s', e)
+        sys.exit()
 
-    _configure_logging(configs)
+    cycle_time, cycle_timestamp = _start_snapshot_cycle(dao_map.get('dao'))
 
     try:
-        pipelines = _build_pipelines(
```
mirons-google: Can we factor this out of main() too?

tests/inventory/pipeline_builder_test.py
```diff
@@ -0,0 +1,170 @@
+# Copyright 2017 Google Inc.
```
mirons-google: Can we add a test around converting module names from this_is_foo to ThisIsFoo?

google/cloud/security/inventory/inventory_loader.py
blueandgold: Right, done.

tests/inventory/pipeline_builder_test.py
blueandgold: Hmmm, not unless it is extracted into a common util function.   The way it is here, it's got functional test coverage to make sure it is working correctly.  :smile:

google/cloud/security/inventory/inventory_loader.py
blueandgold: Sorry, I had asked you about this offline.  But yes, done now.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Sorry, I had asked you about this offline.  But yes, done now.

matthewg:
FYI, I'm introducing a bunch of new resources, such as #305. Is there a test or process in place that will make sure this new stuff gets updated as we add new inventory types?

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -0,0 +1,284 @@
...
+LOGGER = log_util.get_logger(__name__)
+
+
+class PipelineBuilder(object):
+    """Inventory Pipeline Builder."""
+
+    # TODO: Add a flag --list_resources in inventory_loader.py that
+    # will print all the keys() in REQUIREMENTS_MAP.
+
+    REQUIREMENTS_MAP = {
```
thenenadx: Would it make sense to keep the map in a separate mapping file? From my perspective it would make editing quite easier, from the visibility point of view

google/cloud/security/inventory/pipelines/load_projects_cloudsql_pipeline.py
```diff
@@ -34,6 +34,7 @@ class LoadProjectsCloudsqlPipeline(base_pipeline.BasePipeline):
     """Pipeline to load project CloudSql data into Inventory."""
 
     PROJECTS_RESOURCE_NAME = 'project_iam_policies'
+    RESOURCE_NAME = 'cloudsql'
```
thenenadx: As a long term thing, would it make sense to move the resource names as keys under the requirements map?

thenenadx:
Nice PR! I've added a few comments/ideas :) 👍 for the PR

blueandgold:
@matthewg Good idea, I have added a test to make sure the pipelines and the requirement maps are in sync.  I will wait for your PR to be checked in first, before this one.  Thanks!

blueandgold:

PR #307, opened 2017-05-30
Add inventory for remaining IAP services

### LINKED ISSUE #312: Issues loading backend services data into cloudsql
I'm receiving the following errors on [dao.py:186](https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/data_access/dao.py#L186)

google/cloud/security/common/data_access/dao.py:186: Warning: Incorrect integer value: 'False' for column 'enable_cdn' at row 1
  cursor.execute(load_data_sql)
google/cloud/security/common/data_access/dao.py:186: Warning: Incorrect integer value: '' for column 'affinity_cookie_ttl_sec' at row 2
  cursor.execute(load_data_sql)
google/cloud/security/common/data_access/dao.py:186: Warning: Incorrect integer value: '' for column 'enable_cdn' at row 2
  cursor.execute(load_data_sql)
google/cloud/security/common/data_access/dao.py:186: Warning: Incorrect integer value: '' for column 'port' at row 2
  cursor.execute(load_data_sql)
google/cloud/security/common/data_access/dao.py:186: Warning: Incorrect integer value: '' for column 'affinity_cookie_ttl_sec' at row 3
  cursor.execute(load_data_sql)
google/cloud/security/common/data_access/dao.py:186: Warning: Incorrect integer value: '' for column 'enable_cdn' at row 3
  cursor.execute(load_data_sql)
google/cloud/security/common/data_access/dao.py:186: Warning: Incorrect integer value: '' for column 'port' at row 3

matthewg:
Adds:
* Instance
* InstanceGroup
* InstanceGroupManager
* IntanceTemplate

fixes #312


google/cloud/security/inventory/pipelines/load_instance_group_managers_pipeline.py
```diff
@@ -0,0 +1,101 @@
...
+        for project in projects:
+            project_igms = []
+            try:
+                response = self.api_client.get_instance_group_managers(
+                    project.id)
+                for page in response:
+                    items = page.get('items', {})
+                    for group_igms in items.values():
+                        igm_instances = group_igms.get(
+                            'instanceGroupManagers', [])
```
blueandgold: Hmmm.  Additional processing is done here by the caller of ```get_instance_group_managers()``` to get the actual instance group manager data, and aggregate the paged results.  What is your thought about having this done in the ```get_instance_group_managers()```, so that any other potential callers of this api won't have to do the additional processing?  (This will also apply to the other retrieve()'s in this PR.)

For example, you can take a look at the ```get_firewall_rules()``` in the same compute.py that returns only the firewall rules data, rather than just the raw gcp response.
https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/gcp_api/compute.py#L130
```python
        firewall_rules_api = self.service.firewalls()
        request = firewall_rules_api.list(project=project_id)

        paged_results = self._build_paged_result(request, firewall_rules_api,
                                                 self.rate_limiter)

        firewall_rules = []
        for page in paged_results:
            firewall_rules.extend(page.get('items', []))
        return firewall_rules
```

This way, the ```_retrieve()``` can be more simple as well:
```python
        firewall_rules_map = {}
        projects = self.dao.get_projects(self.cycle_timestamp)
        for project in projects:
            try:
                firewall_rules = self.api_client.get_firewall_rules(project.id)
                firewall_rules_map[project.id] = firewall_rules
            except api_errors.ApiExecutionError as e:
                LOGGER.error('Unable to get firewall rules for '
                             'project id: %s\n%s', project.id, e)
        return firewall_rules_map
```

google/cloud/security/common/gcp_api/compute.py
```diff
@@ -47,23 +47,20 @@ def __init__(self, credentials=None, version=None):
     # TODO: Migrate helper functions from gce_firewall_enforcer.py
     # ComputeFirewallAPI class.
 
-    def get_backend_services(self, project_id):
-        """Get the backend services for a project.
+    def _get_paged_list(self, resource_name, list_request, list_next_request):
```
carise: This can be done later, but could you put this in the base client class and call it something like

_build_paged_aggregated_list() or something (there's a method in that class that does this for list_next, but doesn't handle aggregatedList_next).

google/cloud/security/inventory/pipelines/load_instance_group_managers_pipeline.py
matthewg: Done: 7a32561 and e5c6efe

google/cloud/security/common/gcp_api/compute.py
```diff
@@ -47,23 +47,91 @@ def __init__(self, credentials=None, version=None):
     # TODO: Migrate helper functions from gce_firewall_enforcer.py
     # ComputeFirewallAPI class.
 
-    def get_backend_services(self, project_id):
-        """Get the backend services for a project.
+    @classmethod
```
blueandgold: Hmmm, is there a reason to use ```@classmethod``` here?  I see this is being invoked via class instances ```self._flatten_aggregated_list()``` which would make ```@classmethod``` to be not needed.

google/cloud/security/common/gcp_api/compute.py
```diff
@@ -47,23 +47,91 @@ def __init__(self, credentials=None, version=None):
...
+        """
+        items = []
+        for aggregated_results in aggregated_iterator:
+            aggregated_items = aggregated_results.get('items', {})
+            for items_for_grouping in aggregated_items.values():
+                for item in items_for_grouping.get(item_key, []):
+                    items.append(item)
+        return items
+
+    @classmethod
```
blueandgold: Ditto, hope I'm not missing something here.

google/cloud/security/common/gcp_api/compute.py
```diff
@@ -72,7 +140,28 @@ def get_backend_services(self, project_id):
...
-            raise api_errors.ApiExecutionError('backend_services', e)
+            raise api_errors.ApiExecutionError(resource_name, e)
+
+    def get_backend_services(self, project_id):
+        """Get the backend services for a project.
+
+        Args:
+            project_id: The project id.
+
+        Yield:
```
blueandgold: nit: should be a ```Return``` instead of yield.  I see this occur at least one more time down below. Can you please double check everywhere?

google/cloud/security/common/gcp_api/compute.py
matthewg: Done: 0d4a483

google/cloud/security/common/gcp_api/compute.py
matthewg: Done: bfd4d4b

matthewg: @blueandgold and I discussed this offline.  By using the _flatten_list / _flatten_aggregated list, if there's an APIError exception raised by retrieving *any* page, the pipeline won't load *any* of the items.  With the previous approach where we returned an iterator and did the flattening inside the pipeline code, only the single page that raised an error would be omitted but the pipeline loader would log and continue fetching the next page.

That is, if there is an error when retrieving a page, then the pipeline won't load any data for a project, but the pipeline will move to the next project. 

If we care about getting partial results, the APIError exception can be handled in the _flatten_list / _flatten_aggregated list methods instead. However, we decided to leave it as-is for now because:

* Converting from generators to simple lists earlier rather than later simplifies the code.
* Most errors, like access denied, will happen on the first page (given that base_client already retries on transient errors like connection refused.)

If we get user reports of frequent non-first-page errors we can revisit.

blueandgold:

PR #320, opened 2017-06-03
Move scanner summary email to an email pipeline.

blueandgold:
There will be more work to do, to tie this to the notifier.  I will need to talk to gianluca about this, but this is a first step.


blueandgold:

Issue #311, opened 2017-05-30
Ignore "Access not configured"


mirons-google:
<HttpError 403 when requesting https://www.googleapis.com/compute/v1/projects/sample--project/aggregated/backendServices?alt=json returned "Access Not Configured. Compute Engine API has not been used in project 11 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/compute-component.googleapis.com/overview?project=11 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.">

I haven't diagnosed why this is occurring in some projects in my test org, but we should handle these types of errors.

mirons-google:
This is also happening on /forwardingRules

2017-05-30 16:15:45,271 google.cloud.security.inventory.pipelines.load_forwarding_rules_pipeline ERROR    _retrieve GCP API Error: unable to get forwarding_rules from GCP:
<HttpError 403 when requesting https://www.googleapis.com/compute/v1/projects/sample--project/aggregated/forwardingRules?alt=json returned "Project 11 is not found and cannot be used for API calls. If it is recently created, enable Compute Engine API by visiting https://console.developers.google.com/apis/api/compute-component.googleapis.com/overview?project=11 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.">

mirons-google:
In my project where the service account exists these APIs are enabled:
- Google Compute Engine API
- Google Compute Engine Instance Group Manager API
- Google Compute Engine Instance Group Updater API
- Google Compute Engine Instance Groups API

mirons-google:
@matthewg mind taking a look at this for us?

carise:
Do the target projects have compute engine API enabled? I recall getting errors on the projects that don't have compute engine API enabled.

mirons-google:
I don't know exactly and I don't have permissions necessarily to check each project.

Is the answer here different error handling for 403s instead of letting it raise?

carise:
I think these kinds of errors (API not enabled) should just be ignored, but I don't think we can make any assumptions about 403s. We currently don't have a formal policy about pipeline errors, so we might want to wait on this until we decide what to do.

mirons-google:
They're disconcerting for the average user as they're reported as errors on the console.

Arguably, each API should know which responses are harmless and should be ignored. They should be caught and checked. It might also be worth writing a row in that table for the run with a reason as to why it couldn't be retrieved.

[...]

PR #321, opened 2017-06-04
Ignore "API not enabled" errors when listing results.

matthewg:
If the API isn't enabled, then there are no resources of the given type, so
we don't need to surface it as a generic 403 error.

Also updates compute API client to use the same paged-result builder
as _base_client.

Helps with one common case of #311.


blueandgold:

PR #322, opened 2017-06-05
Travis cleanup

mirons-google:
This script is no longer necessary now that travis is built using a docker image.


blueandgold:

PR #325, opened 2017-06-05
Adding --list_resources to Inventory

mirons-google:
This consolidates the configuration file to the samples directory. It also organizes the resources in the configuration file. This separation will be useful once all the inventory flags are moved to the inventory_conf.yaml.sample


google/cloud/security/inventory/inventory_loader.py
```diff
@@ -274,14 +275,19 @@ def _create_dao_map():
...
     inventory_flags = FLAGS.FlagValuesDict()
+
+    if inventory_flags.get('list_resources'):
+        inventory_util.list_resource_pipelines()
+        sys.exit()
+
     _configure_logging(inventory_flags.get('loglevel'))
 
-    if inventory_flags.get('config_path') is None:
+    inventory_conf = inventory_flags.get('config_path')
```
blueandgold: nit: Can we make ```inventory_conf``` to be ```config_path``` to be consistent?  So, this line would be:
```python
config_path = inventory_flags.get('config_path')
```

samples/inventory/inventory_conf.yaml
```diff
@@ -0,0 +1,54 @@
+# Copyright 2017 Google Inc.
```
blueandgold: QQ: If this is in samples/ now, where should it be when actually deployed?

google/cloud/security/inventory/inventory_loader.py
mirons-google: done!

samples/inventory/inventory_conf.yaml
mirons-google: Discussed offline with @blueandgold on possible deployment options.

blueandgold:
LGTM on the changes

blueandgold:

PR #330, opened 2017-06-07
Notifier: Violations Agnostic

gbrindisi:
Hello,
small PR: I've imported the resource map in the notifier to dynamically select violations.
This way we don't need to maintain another map in the notifier.


blueandgold:
Adding `cla: yes` label, as the PR author has ack'd signing the CLA in PR #243.

PR #333, opened 2017-06-07
Add ratelimiter for cloudsql and push the rate limiter down to the _execute()

blueandgold:
There are 2 changes here:
- add rate limiter for sql api calls
- push down the rate limiter context usage to the _execute() of the base client so that the rate limiter is used more consistently everywhere

### DIFF
--- a/google/cloud/security/common/gcp_api/cloud_resource_manager.py
+++ b/google/cloud/security/common/gcp_api/cloud_resource_manager.py
@@ -65,10 +65,9 @@ class CloudResourceManagerClient(_base_client.BaseClient):
         projects_api = self.service.projects()
 
         try:
-            with self.rate_limiter:
-                request = projects_api.get(projectId=project_id)
-                response = self._execute(request)
-                return response
+            request = projects_api.get(projectId=project_id)
+            response = self._execute(request, self.rate_limiter)
+            return response
         except (HttpError, HttpLib2Error) as e:
             raise api_errors.ApiExecutionError(project_id, e)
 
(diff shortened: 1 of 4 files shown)


google/cloud/security/common/gcp_api/cloud_resource_manager.py
```diff
@@ -96,14 +95,13 @@ def get_projects(self, resource_name, **filterargs):
 
         # TODO: Convert this over to _base_client._build_paged_result().
```
mirons-google: Is it time to complete this TODO now that we're touching this code?

google/cloud/security/common/gcp_api/cloud_resource_manager.py
```diff
@@ -161,17 +157,16 @@ def get_organizations(self, resource_name):
...
-                    req_body = {}
-                    if next_page_token:
-                        req_body['pageToken'] = next_page_token
-                    request = organizations_api.search(body=req_body)
-                    response = self._execute(request)
-                    yield response
-                    next_page_token = response.get('nextPageToken')
-                    if not next_page_token:
-                        break
+            while True:
```
mirons-google: Same here.

google/cloud/security/common/gcp_api/cloud_resource_manager.py
```diff
@@ -241,18 +234,17 @@ def get_folders(self, resource_name, **kwargs):
...
-                    request = folders_api.search(body=req_body)
-                    response = self._execute(request)
-                    yield response
-                    next_page_token = response.get('nextPageToken')
-                    if not next_page_token:
-                        break
+            while True:
+                req_body = {}
+                if next_page_token:
+                    req_body['pageToken'] = next_page_token
```
mirons-google: Same here.

mirons-google:
Thanks for doing this!

google/cloud/security/common/gcp_api/cloud_resource_manager.py
blueandgold: The paged result is a separate change (and a separate line item in the workstream), so will tackle that next.

blueandgold:
Yes, this improves the rate limiter usage for all the APIs.

blueandgold:
Just one note here, I did think about how to enforce that rate limiters are used.  But for now, rate_limiter is None for both "unlimited" rate and when someone doesn't define a rate limiter.  So, can not be enforced.  (and ratelimiter does not have a native way to be defined as "unlimited")  We can visit this again if it is an issue.

blueandgold:

PR #335, opened 2017-06-08
Move Flatten Paged Results to Base Client

blueandgold:
There are some paged results which need 2 item_keys to extract the objects of interests, i.e. they are 2 layers deep.  These are not handled in this PR.


google/cloud/security/common/gcp_api/compute.py
```diff
@@ -247,9 +187,11 @@ def get_instance_templates(self, project_id):
         instance_templates_api = self.service.instanceTemplates()
         list_request = instance_templates_api.list(
             project=project_id)
-        return self._flatten_list(
-            self._build_paged_result(list_request, instance_templates_api,
-                                     self.rate_limiter))
+
+        paged_resuls = self._build_paged_result(
```
carise: should we name this "paged_results"?

google/cloud/security/common/gcp_api/compute.py
```diff
@@ -247,9 +187,11 @@ def get_instance_templates(self, project_id):
...
+            list_request, instance_templates_api, self.rate_limiter)
+
+        return self._flatten_list_results(paged_resuls, 'items')
```
carise: "paged_results"?

google/cloud/security/common/gcp_api/compute.py
blueandgold: Done

google/cloud/security/common/gcp_api/compute.py
blueandgold: Done

blueandgold:

PR #337, opened 2017-06-08
Handle paged results that need additional parsings.

blueandgold:
This is for paged results that needs additional processing beyond combining them into a list.


google/cloud/security/common/gcp_api/bigquery.py
```diff
@@ -47,7 +47,7 @@ def get_rate_limiter(self):
         return RateLimiter(FLAGS.max_bigquery_api_calls_per_100_seconds,
                            self.DEFAULT_QUOTA_TIMESPAN_PER_SECONDS)
 
-    def get_bigquery_projectids(self):
+    def get_bigquery_projectids(self, key='projects'):
         """Request and page through bigquery projectids.
```
carise: nit: arg description for "key"?

google/cloud/security/common/gcp_api/bigquery.py
```diff
@@ -64,10 +64,12 @@ def get_bigquery_projectids(self):
...
         paged_results = self._build_paged_result(
             request, bigquery_projects_api, self.rate_limiter)
+
+        flattened_result = self._flatten_list_results(paged_results, key)
+                
         project_ids = []
-        for page in paged_results:
-            for project in page.get('projects', []):
-                project_ids.append(project.get('id'))
+        for result in flattened_result:
```
carise: this is super nitpicky, and I think we are probably not consistent about it, but does it make sense to make this (and the other for loop) a list comprehension? you don't have to, if you don't want to.

google/cloud/security/common/gcp_api/bigquery.py
blueandgold: Actually, this should not even been arg since it's hardcoded.  Wanted to move it earlier inside the method. but this will make me do it.  :)

google/cloud/security/common/gcp_api/bigquery.py
blueandgold: I actually prefer going the other way and not use any list comprehension at all.  :smile:   Reason is it just  readability seems lower with list comprehension.

google/cloud/security/common/gcp_api/bigquery.py
carise: hm, yes, agree it should be moved. I thought 'projects' was just the default for the key.

google/cloud/security/common/gcp_api/bigquery.py
carise: haha. I've been told to make for loops list comprehensions *for* readability. ;)

google/cloud/security/common/gcp_api/bigquery.py
```diff
@@ -59,19 +59,22 @@ def get_bigquery_projectids(self):
             If there are no project_ids enabled for bigquery an empty list will
             be returned.
         """
+        key='projects'
```
mirons-google: I would make this a karg and default it to 'projects'. Otherwise this becomes a magic string that a caller can't influence if things change in how the data is key'd.

google/cloud/security/common/gcp_api/bigquery.py
```diff
@@ -59,19 +59,22 @@ def get_bigquery_projectids(self):
...
-        for page in paged_results:
-            for project in page.get('projects', []):
-                project_ids.append(project.get('id'))
+        for result in flattened_result:
+            project_ids.append(result.get('id'))
 
         return project_ids
 
-    def get_datasets_for_projectid(self, project_id, key='datasets'):
+    def get_datasets_for_projectid(self, project_id):
```
mirons-google: See above.

google/cloud/security/common/gcp_api/bigquery.py
blueandgold: Hmmmm.  I feel that this should not be passed in as an arg for a few reasons:
- the method naming defines a very specific data is retrieved
- the response data structure is very rigid
- even if the the response data structure changes, this should be fixed inside the api method here, rather than letting the caller changing the key

blueandgold: Checked with michael, and this is okay as is.  Merging.

blueandgold:

PR #343, opened 2017-06-12
Update rules GCS paths to match #5

rnhsp:
Documentation update to match the GCS `gs://<bucket-name>` syntax introduced in #5.


blueandgold:
Adding `cla: yes` label, as the PR author is covered by corporate CLA.

PR #351, opened 2017-06-13
Enabling pylint plugin pylint.extensions.emptystring

mirons-google:


google/cloud/security/inventory/pipelines/base_pipeline.py
```diff
@@ -85,7 +85,7 @@ def _to_int(value):
         # pylint: disable=no-else-return
         if value is None:
             return None
-        elif value == '':
+        elif not value:
```
carise: Would it make sense to change lines 86-91 to:

```
if not value:
  return None
else:
  return int(value)
```

Also I wonder if we should wrap in a try/except, in case the int conversion fails. Thoughts?

mirons-google: So as to not break things and keep this PR in-scope I'll add a TODO here.

blueandgold:

PR #356, opened 2017-06-14
pylint changes

mirons-google:
This PR does a few things

1. Adds an example githook for pre-push that looks for TODOs
2. Globally enable doc checking and provide temporary disables in all offending files
  2a. It MUST be the case that the next editor removes the disables and fixes the lint issues
3. Globally disable too-few-public-methods and remove any local disables.


blueandgold:

PR #358, opened 2017-06-14
Fix deploy templates

carise:
1. Bump the ubuntu version, which magically fixes the python version conflict.
2. Don't bomb out the entire inventory loader if an API fails; just move on to the next pipeline. Probably need to do something smart to prevent dependent pipelines from running...
3. Fix errors in the startup script.

### DIFF
--- a/google/cloud/security/inventory/inventory_loader.py
+++ b/google/cloud/security/inventory/inventory_loader.py
@@ -61,6 +61,7 @@ from google.cloud.security.common.data_access import instance_template_dao
 from google.cloud.security.common.data_access import organization_dao
 from google.cloud.security.common.data_access import project_dao
 from google.cloud.security.common.data_access.sql_queries import snapshot_cycles_sql
+from google.cloud.security.common.gcp_api import errors as api_errors
 from google.cloud.security.common.util import log_util
 from google.cloud.security.inventory import api_map
 from google.cloud.security.inventory import errors as inventory_errors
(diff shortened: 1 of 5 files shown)


google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -90,10 +90,15 @@ def _get_api(self, api_name):
                 raise api_errors.ApiInitializationError(e)
 
             api_version = self.api_map.get(api_name).get('version')
-            if api_version is None:
-                api = api_class()
-            else:
-                api = api_class(version=api_version)
+            try:
```
mirons-google: Sorry to do this, since you're touching this file can you resolve the pylint doc messages above?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -61,6 +61,7 @@
 from google.cloud.security.common.data_access import organization_dao
 from google.cloud.security.common.data_access import project_dao
 from google.cloud.security.common.data_access.sql_queries import snapshot_cycles_sql
+from google.cloud.security.common.gcp_api import errors as api_errors
```
mirons-google: Here too please.

google/cloud/security/inventory/pipeline_builder.py
carise: Done

google/cloud/security/inventory/inventory_loader.py
carise: Done

blueandgold:

PR #359, opened 2017-06-15
Verifying that the snapshot table exists in the current database

RajUmadas:


google/cloud/security/common/data_access/sql_queries/snapshot_cycles_sql.py
```diff
@@ -33,7 +33,10 @@
 
 SELECT_SNAPSHOT_CYCLES_TABLE = """
     SELECT TABLE_NAME from information_schema.tables
-    WHERE TABLE_NAME = 'snapshot_cycles';"""
+    WHERE TABLE_NAME = 'snapshot_cycles'
```
carise: Would it be easier to do:

```
WHERE TABLE_SCHEMA = %s
AND TABLE_NAME = 'snapshot_cycles'
```

and pass in the schema name in _exists_snapshot_cycles() (inventory_loader.py)?

carise: Actually I think this is fine, less code change. :)

RajUmadas:
I signed it!

PR #369, opened 2017-06-20
Fix dockerfile to be sorta more usable

carise:


carise:
I'm thinking about adding one more line just for the sake of Travis --

```
RUN yes | which forseti_inventory
```

blueandgold:

PR #373, opened 2017-06-20
Improve site structure & content

denzel-morris:
Major changes in this PR:

- Search input added to navbar
- "Edit on GitHub" button added to documentation page navtabs
- Pull docs navtabs out into a data file
- Pull docs page out into a page collection
- Auto-gen docs two-level hierarchical category navigation from data collection directory structure
- Auto-gen docs article TOC from headers
- Pull FAQs out into data collection
- Factor all inline styles out into appropriate SASS files
- Canonicalize page URLs for search engines

All links should be functioning properly now (and img alt properties added), therefore the CI build should pass. I updated the build process to pass "--assume-extension" to _htmlproofer_ in order to support checking Jekyll's pretty URLs. 

Finally, I will create a follow-on PR to document the _docs_ page collection since ordering the files is slightly unintuitive due to the way Jekyll handles page collections as a flat array-like structure... and we can't build a nested map-like structure that supports hierarchies without plugins which GitHub Pages does not support.


js/toc.js
```diff
@@ -0,0 +1,98 @@
+// https://github.com/ghiculescu/jekyll-table-of-contents
```
mirons-google: Can you point the serving version to the one in third_party/jekyll-table-of-contents/, rather than this copy?

faq/index.md
```diff
@@ -0,0 +1,29 @@
+---
+title: FAQ
+---
```
mirons-google: Can we go ahead and start adding permalinks to the front-matter? If so please fix throughout.

`permalink: /faq/`

_sass/components/_docs_category_nav.scss
```diff
@@ -13,5 +18,10 @@
       color: #1976d2;
       text-decoration: underline;
     }
+
+    ul {
+      list-style-type: none;
+      padding-left: 20px;
+    }
   }
 }
```
mirons-google: nit: Please have an emtpy newline in all files.

_layouts/base.html
```diff
@@ -54,7 +56,14 @@
 
   <!-- jQuery (necessary for Bootstrap's JavaScript plugins) -->
   <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js"></script>
+  <!-- Table of Contents rendering for docs page -->
+  <script src="{{ site.baseurl }}/js/toc.js"></script>
```
mirons-google: See my comment about referencing the version checked into third_party/

_layouts/base.html
```diff
@@ -4,6 +4,8 @@
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
+  <link rel="canoncial" href="{{ page.url | remove:'index.html' | remove:'.html' | absolute_url }}">
```
mirons-google: I'm not sure what you're doing here, but maybe the permalink front-matter I referenced below might help.

https://jekyllrb.com/docs/permalinks/

_data/doc_categories.yml
```diff
@@ -0,0 +1,12 @@
+# The order of the categories in this file (first-to-last) determines their
+# order in the UI (left-to-right)
+- title: Overview
+  href: docs
```
mirons-google: I'm not sure what href: is here, but could permalink front-matter help?

https://jekyllrb.com/docs/permalinks/

_docs/development/index.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Overview
+order: 001
```
mirons-google: Help me understand what `order` is here for?

_includes/docs_category_nav.html
```diff
@@ -0,0 +1,53 @@
+<div class="left-sidebar">
```
mirons-google: Question: what site is built from this logic?

denzel-morris:
Re: Travis, that's to be expected until this pull request is merged. It's looking for an external link to this GitHub repo on the `gh-pages` branch that is not there yet because it's added in this pull request. Once merged, those external link errors will disappear.

js/toc.js
denzel-morris: Sure thing, I'll update this.

_sass/components/_docs_category_nav.scss
denzel-morris: Yes, I'll update this.

_data/doc_categories.yml
denzel-morris: This data file represents a mapping for the docs page category headers. These cannot be inferred from permalink front-matter nor filename, therefore they must be explicitly specified as a site relative link.

_docs/development/index.md
denzel-morris: Within a category, the `order` controls where the page is displayed on the `#left-sidebar` navigation menu. Furthermore, within a category, there can be subcategories, e.g., "Install Locally" and "Install on GCP" exist under a general "Install" sub-category which exists under "How To" category so... "How To" > "Install" > "Install Locally"... each having a respective page.

With that said, `order` is used to control sorting within a sub-category as well. But the mapping is flat and can't be constructed hierarchically without a plugin (which GitHub Pages does not support), so you have to flatten the two orders into one integer. Thus you end up with say:
```
"Overview" order: 101
"Install" order: 201
    "Install Locally" order: 202
    "Install GCP" order: 203
"Another Article" order: 301
```
The first N digits (N=1 in this case) representing the order in the first-level and the last M digits (M=2) representing the order in the second-level of the hierarchy.

Another solution would be to break these out into two separate integers: `category_order` and `subcategory_order` say, and then flatten them in the template by sorting with something like `1000 * category_order + subcategory_order`. That's also an option.

What're your thoughts?

_includes/docs_category_nav.html
denzel-morris: I'm sorry I don't follow what you mean by "site"? If you mean position within the layout, then this is the logic that builds the navigation in the left-sidebar of the docs page layout. So all docs pages use this include by way of the docs layout (_layouts/docs.html).

faq/index.md
denzel-morris: Generally, from my experience, the best practice is to forgo `permalink` unless strictly necessary because it represents a duplication of data. For example, this page currently exists at `/faq/` already.

If you'd like for me to add permalinks though, I can.

_layouts/base.html
denzel-morris: Yes, I'm aware of the permalink functionality. Pretty links are supported by default in Jekyll, but you can also add a filename and extension to reach the same file. This merely represents a canonicalization so that `/docs/quickstarts/inventory` and `/docs/quickstarts/inventory.html` are represented as the same document.

Jekyll's `link` tag doesn't offer pretty link generation out of the box. (You'd need a custom tag which GitHub Pages doesn't support.) Instead, Jekyll's `link` tag produces a URL with an extension (or the filename and extension in the case of index files) when `permalink` isn't present.

So to answer your question, yes `permalink` front-matter would eliminate the need to do this, but I don't think the trade-off is worth it. You're trading 1 place with this logic for N places specifying permalinks that are largely just duplications of data that is already inferred.

If you feel it'd be cleaner to use permalinks though, I'll do that.

_docs/development/index.md
mirons-google: That makes sense. Is it possible to do this based on the directory structure underneath docs/install/gcp/index.md and docs/install/local/index.md? If not, this is fine for now.

denzel-morris: No, it's not with GitHub Pages. That's what I wanted to do, but I'm afraid there's no way to build hierarchical data in Liquid templates (because [no type exists](https://shopify.github.io/liquid/basics/types/)) without plugins.

blueandgold:

PR #376, opened 2017-06-20
Adjusting the instance type.

mirons-google:
We've had a few issues raised about running out of CPU on forseti installations. I've increased the machine type. I've also standardized the way instance-type is loaded between forseti and forseti+explain deployments.


blueandgold:

PR #378, opened 2017-06-21
Address a couple of inventory loading errors.

blueandgold:
- fix truncation error in description fields by increasing space to 512 characters
- document potential fixes for the Incorrect Integer Value message after investigating what this is all about


google/cloud/security/inventory/pipelines/load_instances_pipeline.py
```diff
@@ -99,7 +99,22 @@ def _retrieve(self):
         return instances
 
     def run(self):
-        """Run the pipeline."""
```
mirons-google: Thanks for this. Since you're the first to touch this file after we enabled doc checking can you remove the disables at the top and fix any issues as well?

blueandgold: Done.

blueandgold:

PR #383, opened 2017-06-21
Using the beta 'stages' feature of Travis

mirons-google:
This should launch the three stages in parallel. It also isolates just the `python setup.py install` to docker. The other tests don't need to be run in docker.

It should be noted that a failure in the first stage will fail the other stages.


carise:
I'm wondering if it is helpful to still run the tests in docker. This will ensure that the appropriate dependencies are in place, including protos.

mirons-google:
I don't see how running pylint and unittests need to be run inside the container. I'm happy to switch back, but I think we gain some efficiencies here.

The docker container test still ensures everything is (consistently) built with the same dependencies.

blueandgold:

PR #384, opened 2017-06-22
Implemented model naming

felixbb:


google/cloud/security/iam/client.py
```diff
@@ -79,12 +79,13 @@ def is_available(self):
         data = binascii.hexlify(os.urandom(16))
         return self.stub.Ping(explain_pb2.PingRequest(data=data)).data == data
 
-    def new_model(self, source):
+    def new_model(self, source, name):
```
blueandgold: Do you think that we can start to add the method docs for the code that are touched here.  Or do you have a grand plan to do it later?  :)

blueandgold:

PR #386, opened 2017-06-22
Local docker test execution and linting

felixbb:
Uses caching of ubuntu and pip packages for getting results faster and saving bandwidth.


scripts/run_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+# Delete all running containers
+docker rm -f $(docker ps -a -q)
+
+docker build -t forseti/base -f scripts/docker/base . || exit 1
+docker build -t forseti/build -f scripts/docker/forseti --no-cache . || exit 1
+docker run -it -d --name build forseti/build /bin/bash || exit 1
+
+docker exec -it build /bin/sh -c "coverage run --source='google.cloud.security' --omit='__init__.py' -m unittest discover -s . -p '*_test.py'" || exit 1
```
mirons-google: We don't technically need a coverage run here unless if you're planning on having travis execute this?

scripts/run_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+docker exec -it build /bin/sh -c "pylint --rcfile=pylintrc --ignore-patterns='.*_test.*,.*test_.*,.*test.py,.*pb2.*' --output-format=text google" || exit 1
```
mirons-google: We have many of these pylint commands in various places and need to standardize on it's invocaiton.

felixbb: We should have no more than two - One in travis file and one here.

felixbb: If we want to have one place, we can make travis call this script but it'd be a little ugly and I don't want to touch the build process for now.

scripts/run_tests.sh
felixbb: True, but it also runs the tests and I want to make sure the tests are ran exactly as in travis.

blueandgold:

PR #390, opened 2017-06-22
Added curated role definitions

felixbb:


google/cloud/security/iam/explain/importer/importer.py
```diff
@@ -335,8 +338,13 @@ def _convert_group(self, forseti_group):
                                 type='group',
                                 member_name=forseti_group))
 
+    def _get_permissions_for_role(self, rolename):
+        """Returns permissions defined for that role name."""
```
blueandgold: Hi Felix, can we start to document the args here?  Even better, would you mind to start to enable the pylint checks for this file by removing the above disables?  If not now, do you have any plan to take care of this?  Once again, in the future, this might become someone's else code.  :)

felixbb:
I added the documentation for the new methods. Haven't gone backwards though to fill in all the other documentations since I'd like to do this gradually so the effort is constant per change and doesn't hold back code edits. Once the whole file is documented we can remove the linter disable:...

google/cloud/security/iam/explain/importer/importer.py
```diff
@@ -149,20 +191,55 @@ def run(self):
...
 
+def load_roles():
+    """Load curated roles.
+
+    Returns:
+        defaultdict(set): Map from role name to set of containing permissions.
+    """
+    curated_roles = defaultdict(set)
+
+    csvcontent = roles.CURATED_ROLES_CSV
```
blueandgold: nit: should use underscores to break up the words in the variables ```csv_content```

google/cloud/security/iam/explain/importer/importer.py
```diff
@@ -149,20 +191,55 @@ def run(self):
...
+    rolelist = csv.reader(csvcontent.splitlines(),
```
blueandgold: nit: instead of rolelist, can just say ```roles```

tests/iam/import_tests/importer_test.py
```diff
@@ -48,6 +48,12 @@ class ImporterTest(ForsetiTestCase):
 
     def setUp(self):
         """Create an import from a Forseti dump."""
+        
+        pass
```
blueandgold: nit: if this setUp() is not doing anything, do you want to just drop it?

tests/iam/import_tests/importer_test.py
```diff
@@ -70,7 +76,6 @@ def setUp(self):
                 self.service_config)
             import_runner.run()
 
-    def test_status_done(self):
-        """Test if the status of the import is 'done'."""
         model = self.model_manager.model(self.model_name)
+        print model.message
```
blueandgold: What is this print for?  Usually, we don't checking print statements.

google/cloud/security/iam/explain/importer/importer.py
```diff
@@ -273,27 +383,14 @@ def _convert_policy(self, forseti_policy):
...
-                    'foo.bar.baz',
-                    ]
-                return (list(set([random.choice(perms)
-                                  for _ in range(int(random.random()*10))])))
-
             try:
                 role = self.session.query(self.dao.TBL_ROLE).filter(
                     self.dao.TBL_ROLE.name == binding.get_role()).one()
-            except:
+            except Exception:
```
blueandgold: Can you move the disable:broad-except here?  Please put a TODO to improve this.
```
except Exception:  # pylint: disable=broad-except
```

google/cloud/security/iam/explain/importer/importer.py
```diff
@@ -16,21 +16,13 @@
 
 import json
 import traceback
+import csv
```
blueandgold: nit: pylint didn't complain about this not being sorted?

```
from collections import defaultdict
import csv
import json
from StringIO import StringIO
from time import time
import traceback
```

felixbb: No, otherwise the build would be broken. Please configure pylint to complain.

blueandgold:

PR #393, opened 2017-06-23
Prereqs wizard

carise:
I'd like to get feedback on actual usage of this... Would either of you mind trying it out and giving feedback?

### DIFF
--- a/scripts/gcp_setup/environment/gcloud_env.py
+++ b/scripts/gcp_setup/environment/gcloud_env.py
@@ -749,23 +749,25 @@ class ForsetiGcpSetup(object):
                 'deploy-forseti-{}.yaml'.format(os.getpid())))
 
         # Ask for SendGrid API Key
-        sendgrid_api_key = raw_input('What is your SendGrid API key? ').strip()
+        sendgrid_api_key = raw_input(
+            'What is your SendGrid API key? (press [enter] to '
+            'leave blank) ').strip()
         if sendgrid_api_key:
             self.sendgrid_api_key = sendgrid_api_key
 
         # Ask for notification sender email
-        notif_sender_email = raw_input(
+        notification_sender_email = raw_input(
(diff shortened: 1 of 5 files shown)


scripts/gcp_setup/environment/gcloud_env.py
```diff
@@ -721,13 +737,89 @@ def create_cloudsql_instance(self):
...
+            os.path.join(
+                self.DEPLOY_TPL_DIR_PATH,
+                'deploy-forseti.yaml.in'))
+        out_tpl_path = os.path.abspath(
+            os.path.join(
+                self.DEPLOY_TPL_DIR_PATH,
+                'deploy-forseti-{}.yaml'.format(os.getpid())))
+
+        # Ask for SendGrid API Key
+        sendgrid_api_key = raw_input('What is your SendGrid API key? ').strip()
```
blueandgold: Is there something that tell the user what to do, if they don't have one?

scripts/gcp_setup/environment/gcloud_env.py
```diff
@@ -721,13 +737,89 @@ def create_cloudsql_instance(self):
...
+
+        # Ask for notification sender email
+        notif_sender_email = raw_input(
+            'What is the sender email address for your notifications? '
+            '(press [enter] to leave blank) ').strip()
+        if notif_sender_email:
+            self.notification_sender_email = notif_sender_email
+
+        # Ask for notification recipient email
+        notif_recip_email = raw_input(
```
blueandgold: These abbreviations look a bit odd to read.  If you think the full spellings are too long, then how about just using ```recipient_email```?  Ditto for ```notif_sender_email``` above.

blueandgold:
Sorry for the lateness, but I would love to try this out.  Have been caught up in my yaml config PR, but should have time to do this.

blueandgold:

PR #413, opened 2017-06-28
Page-configurable hiding of left and right sidebars on documentation pages

denzel-morris:
Allow authors to control whether the left- or right-sidebar is displayed on docs page with front-matter:

```yaml
hide:
  left_sidebar: true
  right_sidebar: true
```

Along with description of convention in README.


blueandgold:

PR #418, opened 2017-06-29
Htmlproofer fix

mirons-google:
I found that Github had introduced a change to their `github-pages` gem that updated the `{{ site.github.page }}` URL to be the URL of the respective pages site. I found this by running `bundle update` and `bundle exec jekyll build && htmlproofer _/site/`.

This PR fixes the gh-pages build by using the `{{ site.github.repository_url }}` for edit link generation instead.


blueandgold:

PR #419, opened 2017-06-29
Add codecov.yml

mirons-google:
This sets the threshold for a 1% drop in coverage before failing the PR.


blueandgold:

PR #421, opened 2017-06-30
Fixed error when violation table is already present

gbrindisi:
When running different scanners if the table already exist the violation_dao will throw an error halting the insertion flow.

Now the exception is parsed and handled to allow for insertions on the same table.

Also the notifier couldn't load violations anymore.


gbrindisi:
How should I fix the line too long in this:

```
from google.cloud.security.common.data_access.sql_queries import select_data as sd
```

mirons-google:
Hey @gbrindisi you can put a pylint disable in:

`# pylint: disable=line-too-long`
....
`# pylint: enable=line-too-long`

gbrindisi:
@mirons-google ah right! thanks

google/cloud/security/common/data_access/sql_queries/select_data.py
```diff
@@ -149,11 +149,24 @@
     WHERE project_number = {1};
 """
 
-
-SELECT_VIOLATIONS = """
+SELECT_POLICY_VIOLATIONS = """
     SELECT * FROM violations_{0};
```
blueandgold: Does this need a WHERE clause?

google/cloud/security/common/data_access/sql_queries/select_data.py
```diff
@@ -149,11 +149,24 @@
...
+    SELECT * FROM violations_{0}
+    WHERE violation_type = 'BUCKET_VIOLATION';
+"""
+SELECT_CLOUDSQL_VIOLATIONS = """
+    SELECT * FROM violations_{0}
+    WHERE violation_type = 'CLOUD_SQL_VIOLATION';
+"""
+
 SELECT_GROUPS_VIOLATIONS = """
     SELECT * FROM groups_violations_{0};
```
blueandgold: Does this need to be updated too?  I think that nenad moved the groups violations here too?

google/cloud/security/common/data_access/sql_queries/select_data.py
```diff
@@ -149,11 +149,24 @@
...
 
-
-SELECT_VIOLATIONS = """
+SELECT_POLICY_VIOLATIONS = """
     SELECT * FROM violations_{0};
 """
 
+SELECT_BIGQUERY_ACL_VIOLATIONS = """
+    SELECT * FROM violations_{0}
+    WHERE violation_type = 'BIGQUERY_VIOLATION';
```
blueandgold: I think there is a way to reduce these sql to a generic statement, where the violation type can be specified as {1}.  It might take some work, so please put a TODO.

carise: Please use %s instead of {1}, because of sql injections. We only use the format placeholders for tablenames.

https://stackoverflow.com/questions/7929364/python-best-practice-and-securest-to-connect-to-mysql-and-execute-queries

google/cloud/security/common/data_access/sql_queries/select_data.py
gbrindisi: Right, nice catch.

google/cloud/security/common/data_access/sql_queries/select_data.py
gbrindisi: Groups violations should still be in a separate table, see: https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/data_access/sql_queries/create_tables.py#L487

@thenenadx to confirm

thenenadx: Yes, group violations are in a separate table, because they have a different format. @blueandgold @gbrindisi

google/cloud/security/common/data_access/sql_queries/select_data.py
```diff
@@ -149,9 +149,25 @@
     WHERE project_number = {1};
 """
 
+# TODO: reduce these sql to a generic statement
+SELECT_POLICY_VIOLATIONS = """
+    SELECT * FROM violations_{0}
+    WHERE violation_type = 'ADDED' 
+    OR violation_type = 'REMOVED';
```
blueandgold: Can we change this ```OR``` to use a list?
```WHERE violation_type in ['ADDED', 'REMOVED']```
https://stackoverflow.com/a/8945575/2830207

blueandgold:
Adding `cla: yes` label, as the PR author has ack'd signing the CLA in PR #243  .

PR #423, opened 2017-06-30
Add custom 404 page

denzel-morris:
Adds a custom 404 page. Minor heads up: the image provided had a background color of #d8d8d8 vs. #bdbdbd, so I went with the former.


blueandgold:

PR #426, opened 2017-06-30
prettify markdown tables

mirons-google:
Note this only does it for the one table that we have. Going forward this will need to be done for each markdown generated table. Thanks to @denzel-morris for figuring this out.


blueandgold:
Adding `cla: yes` label, as the googlebot approved.

PR #427, opened 2017-06-30
Fix email scanner pipeline to handle flattened violations

carise:
This is to fix the email scanner summary to handle the flattened violations.


blueandgold:

PR #429, opened 2017-07-05
Copy text button for code blocks

denzel-morris:
Add a 'copy text to clipboard' button that displays on hover of a code-block.

I've chosen **not** to display the button on mobile phones because I don't think copying code is a useful on mobile phones. If you feel that's an incorrect assumption, then please comment, and I can enable the on mobile phones as well.


blueandgold:

PR #430, opened 2017-07-05
fix the bucket sample rule

blueandgold:
These fields are needed or else the bucket scanner would trip.


blueandgold:

PR #433, opened 2017-07-06
Use the IN operator

blueandgold:
A bit cleaner instead of the OR operator.


blueandgold:

PR #435, opened 2017-07-06
Make the projects inventory pipeline depends on folders.

blueandgold:
Otherwise the scanner will complain if we only inventory org and projects.


blueandgold:

PR #436, opened 2017-07-06
Update Guides structure

Tabrarian:
Remove redundant directory, consolidate Guides landing page, create landing page for Contributing.


_docs/guides/contributing.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Contributing to Forseti
+order: 001
```
mirons-google: This will need to be 003

Tabrarian:
I updated order to address conflicts, but set it so contributing is above testing. Since we only have 2 pages, I don't think we need to create a sub-directory at this time, but will keep in mind if we add more content later. It looks like it would be good to add sub-directories to the Howtos though; I'll start a separate PR for that.

carise:
Sorry for not responding sooner on this PR. Could you resolve the merge conflict on the files noted above?

Tabrarian:
The merge conflicts are too complex to resolve in the web UI (per the UI tooltip), but I don't have a local repository since I've only been working in the web UI. I'm going to close this one out and try smaller command-line changes to make these fixes. Apologies for the trouble!

blueandgold:

PR #437, opened 2017-07-06
Homepage

mirons-google:
This adds iconography to the home page.

@denzel-morris please also take a look


blueandgold:

PR #440, opened 2017-07-07
Update to favicon with transparent background

denzel-morris:
Simple swap of the favicon as requested in the requirements doc.


blueandgold:

PR #441, opened 2017-07-07
Deep link to any heading; FAQ

denzel-morris:
All headers (`h1-6`) with an `id` attribute on any page are deep linkable. When a user hovers over the heading, a clickable link-icon appears that can be used for deep linking within a page.

Furthermore, FAQs have the same link-icon on hover; however, in the case of FAQs, a deep link will open the appropriate accordion panel as well.

### Implementation Note

Kramdown automatically generates IDs for Markdown headers based upon the content, e.g., `# Header` becomes `<h1 id="header">Header</h1>`. This is all well and good until you need to include Markdown content in another page where those headings may not necessarily be unique... as is the case with the FAQs.

Therefore, for now, we must disable auto ID generation on every FAQ item under `_faq/` until an issue within Jekyll is fixed. I've [reported the issue](https://github.com/jekyll/jekyll/issues/6209) to Jekyll.

Once that is fixed, we should be able to access the unrendered Markdown and disable auto ID generation programatically before rendering.


blueandgold:

PR #442, opened 2017-07-07
Add Releases page

denzel-morris:
The */releases* page lists all versions along with:

- Release date
- ZIP/TAR archive download links
- Release notes snippet
- Link to "Read more..." of the release notes

The length (measured in number of words) of the release notes snippet can be controlled by the `maxwords` parameter passed into `_includes/releases/version_item.html` (which is forwarded to `_includes/markdown_snipify.liquid`). It's currently set to 50. It can be increased or decreased as necessary.


blueandgold:

PR #444, opened 2017-07-08
Wildcard resource rules

### LINKED ISSUE #432: Add wildcard support for resource ids in IAM policy scanner
Currently there is no way to specify a resource type + resource id where the resource id is just "*". We will need this so users can create rules that apply to all resource types (e.g. "for all folders, do not allow any users to have 'roles/someRole'.")

carise:
Allow IAM rules where the resource id is a wildcard.

resolves #432


carise:
Thanks for the quick review!

blueandgold:

PR #447, opened 2017-07-10
Folders IAM policies

carise:
Add folder IAM policies.

Also add some tests and fix some bugs.


carise:
Thanks for the review!

blueandgold:

PR #461, opened 2017-07-13
Deployment Manager templates

carise:
This has been tested on Deployment Manager with a newly created deployment.

One thing to note is that this does NOT solve the config problem; the changes here is mainly to move all the relevant Deployment Manager templates and scripts in the same place. Also, I've created schema files for each template to help the user check for type correctness and that required properties are present in their deployment configuration.


blueandgold:

PR #479, opened 2017-07-18
Explain pre release

felixbb:

- [X] All of the [unit-tests](http://forsetisecurity.org/docs/development/#executing-tests) still pass
- [X] Running `pylint --rcfile=pylintrc` passes.


google/cloud/security/common/data_access/forseti.py
```diff
@@ -254,6 +257,33 @@ def __repr__(self):
                 self.name,
                 self.display_name)
 
+    if schema_number >= 2.0:
```
blueandgold: Is this ```if``` condition necessary here?  You seem to have implemented this on the caller below, which is a good thing in general.

felixbb: Defining the class can potentially access the database with the assumption that such table exists or will be created otherwise. I'd rather get a 'undefined type error' somewhere then accidentally writing this to a database with an earlier schema. Hence the if is required.

blueandgold:

PR #487, opened 2017-07-19
Fix IAP scanner email subject

matthewg:


google/cloud/security/notifier/pipelines/email_scanner_summary_pipeline.py
```diff
@@ -142,8 +144,8 @@ def _send(  # pylint: disable=arguments-differ
             disposition='attachment',
             content_id='Scanner Violations'
         )
-        scanner_subject = 'Policy Scan Complete - {} violation(s) found'.format(
-            total_violations)
+        scanner_subject = '{} Complete - {} violation(s) found'.format(
+            total_violations, email_description)
```
blueandgold: I think, the order of the formatted item needs to be: email description first, and then total_violations

blueandgold:

PR #488, opened 2017-07-19
(correctly) append the user-agent

mirons-google:

- [x] All of the [unit-tests](http://forsetisecurity.org/docs/development/#executing-tests) still pass


google/cloud/security/common/gcp_api/_base_client.py
```diff
@@ -41,8 +41,11 @@ def _attach_user_agent(request):
     Returns:
         HttpRequest: A modified googleapiclient request object.
     """
-    user_agent = request.headers['user-agent']
-    request.headers['user-agent'] = user_agent + ', %s/%s ' % (
+    user_agent = request.headers.get('user-agent', '')
+    if not user_agent or user_agent.contains(forseti_security.__package_name__):
```
blueandgold: qq: why the ```if not user_agent``` here?  Since you are explicitly forcing a default of '', you would want to add the user agent?

blueandgold: IOW: my question here is really why are we not adding a user-agent, when the request doesn't contain a user-agent in the first place.  Might be good to also add a comment to explain this in the future.  :)

mirons-google: My thought was not muck with things if they didn't exist (too many possible reasons as to why that could not be set).

So only when it's set will we append to it.

blueandgold: Got it, sounds good.

blueandgold:

PR #491, opened 2017-07-19
Fix getting instance tags in IAP scanner

matthewg:


blueandgold:

PR #500, opened 2017-07-21
Tweak Bucket ACL pipeline

carise:
Since we are trying to avoid assigning storageAdmin as a role for Forseti, instead we will pull bucket ACLs with the `buckets().list(project=project_id, projection='full')` query and then read the ACLs from the bucket raw json. There are no database schema changes in this PR, only code changes.

@thenenadx just pinging you on this PR as an FYI, so you have a heads up about the changes.

### DIFF
--- a/google/cloud/security/inventory/pipelines/base_pipeline.py
+++ b/google/cloud/security/inventory/pipelines/base_pipeline.py
@@ -54,6 +54,7 @@ class BasePipeline(object):
     @abc.abstractmethod
     def run(self):
         """Runs the pipeline."""
+        # TODO: should this raise an error?
         pass
 
     @abc.abstractmethod
--- a/google/cloud/security/inventory/pipelines/load_projects_buckets_acls_pipeline.py
+++ b/google/cloud/security/inventory/pipelines/load_projects_buckets_acls_pipeline.py
@@ -80,8 +80,7 @@ class LoadProjectsBucketsAclsPipeline(base_pipeline.BasePipeline):
 
         # Get the projects for which we will retrieve the buckets.
         try:
-            raw_buckets = self.dao.get_raw_buckets(
-                'buckets', self.cycle_timestamp)
+            raw_buckets = self.dao.get_raw_buckets(self.cycle_timestamp)
         except data_access_errors.MySQLError as e:
             raise inventory_errors.LoadDataPipelineError(e)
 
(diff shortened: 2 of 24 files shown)


google/cloud/security/inventory/pipelines/load_projects_buckets_acls_pipeline.py
```diff
@@ -78,40 +76,29 @@ def _retrieve(self):
...
             LoadDataPipelineException: An error with loading data has occurred.
         """
+        buckets_acls = []
+
         # Get the projects for which we will retrieve the buckets.
         try:
-            project_numbers = self.dao.get_project_numbers(
-                self.PROJECTS_RESOURCE_NAME, self.cycle_timestamp)
+            raw_buckets = self.dao.get_raw_buckets(
+                'buckets', self.cycle_timestamp)
```
thenenadx: can we replace the hardcoded 'buckets' string with a variable for code style consistency?

thenenadx:
@carise I shared some thoughts. Other than that, looks good to me! Great PR

google/cloud/security/inventory/pipelines/load_projects_buckets_acls_pipeline.py
carise: Yes, I will move the resource name to the dao.

google/cloud/security/inventory/pipelines/base_pipeline.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, cycle_timestamp, global_configs, api_client, dao):
     @abc.abstractmethod
     def run(self):
         """Runs the pipeline."""
+        # TODO: should this raise an error?
```
ahoying: Since this is an abstract method of the class, subclasses have to define their own implementation of foo or they will fail with TypeError at runtime:

>>> import abc
>>> class test(object):
...   __metaclass__ = abc.ABCMeta
...   @abc.abstractmethod
...   def run(self):
...     pass
... 
>>> class subtest(test):
...   def other(self):
...     pass
... 
>>> s = subtest()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class subtest with abstract methods run

carise:
Thanks all for the review!

blueandgold:

PR #499, opened 2017-07-21
Explain our access requirements

mirons-google:
This PR clarifies and 'abstracts' our IAM role requirements along with the required APIs. It also created a new guide on best practices and explains in details why we need certain service accounts and access.

PR #501, opened 2017-07-21
Fix setupwizard roles

carise:
Assign roles to org/project as needed.


mirons-google:
Feel free to push back on this, but in #499 it was suggested by @blueandgold to use the naming convention for service accounts in that best-practices document to align with what this wizard creates. It might be best to sneak this in with this PR.

blueandgold:

PR #514, opened 2017-07-28
Howto structure

Tabrarian:
Creating a sub-structure for the How-to guides so the ToC is more readable. Also removing the scanner rule violations PH since we have policy violations under the scanner quickstart (see http://forsetisecurity.org/docs/quickstarts/scanner/policy-violations). If there's unique content that should be separate, we can create a new page when the content is ready.

### DIFF
--- a/_docs/howto/configure/index.md
+++ b/_docs/howto/configure/index.md
@@ -1,0 +1,13 @@
+---
+title: Enable or Change Forseti Security features
+order: 101
+---
+# {{ page.title }}
+
+Learn how to enable or change features about Forseti Security.
+
+- [Enabling Email Notifications]({% link _docs/howto/configure/email-notification.md %})
+- [Enabling GSuite Google Groups Collection]({% link _docs/howto/configure/gsuite-group-collection.md %})
+- [Collecting and Storing New Data in Forseti Inventory]({% link _docs/howto/configure/inventory-new-data.md %})
+- [Defining Rules for Forseti Scanner]({% link _docs/howto/configure/scanner-rules.md %})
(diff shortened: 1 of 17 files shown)


Tabrarian:
As soon as I clicked submit, I remembered I wanted to check for link breakage - doing that now and will add commits as needed.

Tabrarian:
I think I have all the links updated across the site now. Links to the new pages are failing in Travis because the pages don't exist yet, but the syntax is correct.

_docs/howto/configure/index.md
```diff
@@ -0,0 +1,13 @@
+---
+title: Configure Forseti Security Features
```
mirons-google: nit: In previewing this these titles run long in the left navbar. I suggest either creating our own title and not using {{page.title}} or just renaming the doc to `title: Configure Features`

_docs/howto/deploy/index.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Deploy Forseti Security
```
mirons-google: Same suggestion as above `title: Deploy`

_docs/howto/configure/index.md
```diff
@@ -0,0 +1,13 @@
...
+order: 101
```
mirons-google: In previewing this all the articles are listing under deploy in the left sidebar. To get this to work we need a new leading digit for articles in `/configure/...`.

This would become `order: 201`.

mirons-google: (seems github is having issues and my comment was lost)

All articles under `/configure/` will need the sam leading digit.

PR #515, opened 2017-07-28
Change the iam rule to blacklist.

blueandgold:
This fits better with the description of "disallow the service account" and also highlight the blacklist mode.


blueandgold:

PR #520, opened 2017-08-01
Update article titles to more readable ones

mirons-google:


Tabrarian:
I'm concerned about clipping too much to where the title doesn't make sense on its own (e.g., for a screen reader, "Locally" doesn't tell them much). What's the max character count you'd like to aim for? Do you think we could trim to "Deploy Forseti Locally" and "Deploy Forseti on GCP"?

_docs/howto/deploy/local-deployment.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Deploying Forseti in a Local Environment
+title: Locally
```
carise: Local/Local deployment?

_docs/howto/deploy/gcp-deployment.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Deploying Forseti on Google Cloud Platform
+title: Google Cloud Platform
```
carise: Google Cloud Platform Deployment?

mirons-google:
Fair point.

@Tabrarian I'm open to a suggestion. Repeating things like "Deploy" and "Configure" seemed redundant now that they're in that sub-level. Maybe the answer here is to not use {{page.title}} as the page title and rather use a longer form there?

Tabrarian:
I'd rather err on the side of some redundancy for readability (rather than assume a user will read it as "Deploy Locally"). There are also other deployment-related topics there like "Changing a Deployment" that break the pattern.

Consider also that the page title is what's displayed on the browser tab label; from a user perspective, it's useful to have a complete title (the user otherwise would only see "Forseti Security / Locally"). I see the case for dropping "Forseti" from the title at least and go with Deploy Locally (or Local Deployment) and Deploy on GCP (or GCP Deployment), which should be short enough to avoid line-wrapping.

mirons-google:
Great, agreed. I'll take the suggestion of "Local Deployment" and "GCP Deployment"

mirons-google:
@Tabrarian @carise PTAL

PR #521, opened 2017-08-01
Fix codeblock formatting

mirons-google:


_docs/quickstarts/scanner/index.md
```diff
@@ -28,9 +28,9 @@ To run Forseti Scanner, follow the process below:
 
   1. Run the scanners:
 
-          ```bash
-          $ forseti_scanner --forseti_config <path to forseti_conf.yaml>
-          ```
+  ```bash
+  $ forseti_scanner --forseti_config <path to forseti_conf.yaml>
```
carise: Should we remove the "$ " preceding the command so users can just click the "copy" icon and paste into their terminal?

mirons-google: Good point, we'd want to make a decision about doing this globally and just not here.

PR #534, opened 2017-08-02
Change page title

carise:
I think it makes sense to have a page title that describes what the page is about.


mirons-google:
nit: This is more than setup. I'd prefer the word 'Installation'.

PR #536, opened 2017-08-04
Initialize All Tables

blueandgold:
Ensure all the tables are created, to support client usages & joins, which expect the tables to exist even if empty.


mirons-google:
Approved, but it would be nice to include test coverage for this and anything else here you're willing to do.

blueandgold:
Let me punt on the testing here as this whole piece might be changing soon.  Thanks!

PR #540, opened 2017-08-04
Track Inventory Data Status (WIP)

blueandgold:
This PR is an effort to track whether or not, we are successful to get all the data from GCP, apart from whether the inventory pipeline is successful.  

IOW: It is possible for an inventory pipeline to complete successfully, even if some data is not retrieved from the GCP api.  e.g. we prefer pipeline to keep going to the next project, if there is a problem to retrieve the IAM policies from one project.

This is a WIP, that shows the change for 2 pipelines.  Would like to get the initial review before doing this for all pipelines.

### DIFF
--- a/google/cloud/security/common/data_access/sql_queries/snapshot_cycles_sql.py
+++ b/google/cloud/security/common/data_access/sql_queries/snapshot_cycles_sql.py
@@ -24,6 +24,7 @@ CREATE_TABLE = """
         `complete_time` datetime DEFAULT NULL,
         `status` enum('SUCCESS','RUNNING','FAILURE',
                       'PARTIAL_SUCCESS','TIMEOUT') DEFAULT NULL,
+        `has_all_data` bool DEFAULT NULL,
         `schema_version` varchar(255) DEFAULT NULL,
         `cycle_timestamp` varchar(255) DEFAULT NULL,
          PRIMARY KEY (`id`),
@@ -46,6 +47,6 @@ INSERT_CYCLE = """
 
 UPDATE_CYCLE = """
     UPDATE snapshot_cycles
-    SET status=%s, complete_time=%s
+    SET status=%s, has_all_data=%s, complete_time=%s
     WHERE cycle_timestamp=%s;
 """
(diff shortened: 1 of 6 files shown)


google/cloud/security/common/data_access/sql_queries/snapshot_cycles_sql.py
```diff
@@ -24,6 +24,7 @@
         `complete_time` datetime DEFAULT NULL,
         `status` enum('SUCCESS','RUNNING','FAILURE',
                       'PARTIAL_SUCCESS','TIMEOUT') DEFAULT NULL,
+        `has_all_data` bool DEFAULT NULL,
```
mirons-google: Why can't this be determined by 'PARTIAL_SUCCESS' in the `status` field?

blueandgold: Sorry if I didn't explain this clearly in the PR description.  So, the status fields refer to the pipelines themselves, whether they finished running or not (i.e. crashed).  So, it is possible for the pipelines to continue running and considered to be ```successful```, even if there are problems to retrieve subsets of data (e.g. the iam policies of a project) along the way.  That is why we need some way to know if data has been all retrieved or not.  We could change our definition of the current ```status``` to refer to data as well.  Please let me know.

mirons-google: Gotcha, then I think we have a need to rename both fields:

`pipeline_execution_status`
`pipeline_data_status`

The later should be a boolean. We should also define these and describe the enum in our docs for `pipeline_execution_status`

blueandgold: I think that effect of renaming the ```status``` column will be minimal.  It seems only IAM explain will be impacted, so I won't handle the schema change just for it.  @felixbb, is this okay?

If we want to change the naming to ```pipeline_data_status```, it might be better to have ```SUCCESS``` and ```FAILURE``` in order to be consistent with ```pipeline_execution_status```.  Otherwise, it seems odd if one is an enum, and the other is a bool, even though they are named the same?  @mirons-google

mirons-google: @blueandgold I leave it up to you to know if the other enum statuses are being used and if moving to a boolean is satisfactory given the ways a pipeline can fail.

PR #541, opened 2017-08-07
Service account inventory pipeline

gbrindisi:
Hello!
Here's the inventory pipeline for Service Accounts.

Currently it retrieves SA and associated keys for every project.

-------------

- [X] I signed Google's [Contributor License Agreement](https://opensource.google.com/docs/cla/)
- [X] All of the [unit-tests](http://forsetisecurity.org/docs/development/#executing-tests) still pass
- [X] Running `pylint --rcfile=pylintrc` passes.


google/cloud/security/common/data_access/csv_writer.py
```diff
@@ -385,6 +385,15 @@
     'iam_policy'
 ]
 
+SERVICE_ACCOUNTS_FIELDNAMES = [
+    'project_id',
+    'name',
+    'email',
+    'oauth2_client_id',
```
blueandgold: +100 for this data; I think it will be really useful for tracking the permissioning of gsuite's api scopes, which right now is assigned by client id, but is really hard to determine what service account it's for.

tests/inventory/pipelines/load_service_accounts_pipeline_test.py
```diff
@@ -0,0 +1,145 @@
...
+    def test_retrieve_data_is_correct(
+            self, mock_get_projects, mock_conn):
+        """Test _retrieve() data is correct."""
+        mock_get_projects.return_value = self.projects
+        service_accounts = []
+        for p in self.project_ids:
+            service_accounts.append(
+                fake_service_accounts.FAKE_PROJECT_SERVICE_ACCOUNTS_MAP[p])
+
+        print "aaaaaaaaaaaaa"
```
blueandgold: nit: can we remove these couple of print statements?

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
```diff
@@ -0,0 +1,81 @@
...
+            iterator: Service Accounts in a dict.
+        """
+        for project_id, service_accounts in resource_from_api.iteritems():
+            for service_account in service_accounts:
+                yield {
+                    'project_id': project_id,
+                    'name': service_account.get('name'),
+                    'email': service_account.get('email'),
+                    'oauth2_client_id': service_account.get('oauth2ClientId'),
+                    'account_keys': json.dumps(service_account.get('keys')),
```
blueandgold: nit: Totally your call, but do you think that it might be better to change the naming of ```account_keys``` to just ```keys```?  This would more consistent with the naming with the API data, and less to type?

google/cloud/security/common/gcp_api/iam.py
```diff
@@ -0,0 +1,108 @@
...
+        next_token = ''
+        while True:
+            api_call = endpoint(name=project_name,
+                                pageToken=next_token)
+
+            try:
+                result = api_call.execute()
+            except (errors.HttpError, HttpLib2Error) as e:
+                LOGGER.error(api_errors.ApiExecutionError(project_name, e))
+                # TODO: pass in "buckets" as resource_name variable
```
blueandgold: nit: is this comment about ```buckets``` valid?

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
```diff
@@ -0,0 +1,81 @@
...
+                    'raw_service_account': json.dumps(service_account),
+                }
+
+    def run(self):
+        """Run the pipeline."""
+        service_accounts = self._retrieve()
+        loadable_service_accounts = self._transform(service_accounts)
+        self._load(self.RESOURCE_NAME, loadable_service_accounts)
```
blueandgold: I see from the documentation, that service accounts also have methods to get IAM policies, which we have determined is important to inventory along with the resource.  Do you have plans to get these IAM policies as well?  It's okay if you don't do it in this CL, but can you please add a TODO?
https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
blueandgold: qq: It looks like the ```privateKeyData``` will be stored in Forseti inventory.  Is this okay?  I will defer to the best security practice here.  :)

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
```diff
@@ -0,0 +1,81 @@
...
+
+from google.cloud.security.common.data_access import project_dao as proj_dao
+from google.cloud.security.common.util import log_util
+from google.cloud.security.inventory.pipelines import base_pipeline
+
+
+LOGGER = log_util.get_logger(__name__)
+
+
+class LoadServiceAccountsPipeline(base_pipeline.BasePipeline):
```
blueandgold: Is this pipeline added to the ```forseti-security/configs/forseti_conf.yaml.sample``` file?

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
gbrindisi: Turns out `keys` is a reserved word in MySQL, so after banging our heads against a random sql error we decided to rename the column name and for consistency sake we change in the code as well

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
gbrindisi: We don't have plans for that right now, but definitely I'm going to add a todo

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
gbrindisi: Mmm, according to the docs `privateKeyData` is only returned on `keys.create` calls:
https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts.keys#ServiceAccountPrivateKeyType

Should be safe, but let me run this again and check what I get.

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
gbrindisi: Added

tests/inventory/pipelines/load_service_accounts_pipeline_test.py
gbrindisi: Done

google/cloud/security/inventory/pipelines/load_service_accounts_pipeline.py
blueandgold: Ah, didn't know that this was a reserved word in mysql.  :(

blueandgold: Yah, I read the doc more closely, and you are correct that this is not returned by the list.
https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts.keys#ServiceAccountKey

gbrindisi:
Yay! ✨ 🍰 ✨

PR #542, opened 2017-08-07
Add scanner violations image

Tabrarian:
Adding image to scanner violations to show sample output. Staged and working locally.


carise:
Hi Tabitha, is it ok if I send you a scrubbed version of the scanner violation output?

Tabrarian:
Of course, thanks!

PR #559, opened 2017-08-15
hide the main page landing page from sidebar nav

carise:
Hide the landing page link from the left navbar links, because it's just redundant.

Example:

Current: 
screen shot 2017-08-14 at 10 36 55 pm

After fix:
screen shot 2017-08-14 at 10 37 01 pm

### DIFF
--- a/_includes/site/layout/docs/category_nav.html
+++ b/_includes/site/layout/docs/category_nav.html
@@ -45,9 +45,11 @@
         </li>
       {% endif %}
 
+      {% if forloop.index != 1 %}
       <li {% if page.url == doc.url %}class="active"{% endif %}>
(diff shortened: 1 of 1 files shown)


mirons-google:
Thanks for this as I wanted to do this too! One concern though is how someone get's back to `index.md`? Should we link the heading there?

carise:
Good idea, I was considering that -- can make that change later today.

carise:
Now the sidebar header is directly linked to the category's root url. Note how mousing over the header shows the link to the root url, even though you're on a different page.

screen shot 2017-08-15 at 1 48 48 pm

mirons-google:
Is the header always underlined?

carise:
No, I'm mousing over it

PR #560, opened 2017-08-16
Update admin sdk quota limits to use the 100sec bucket

mirons-google:
Switching the admin directory API to use the lower of the two buckets.

### DIFF
--- a/google/cloud/security/common/gcp_api/admin_directory.py
+++ b/google/cloud/security/common/gcp_api/admin_directory.py
@@ -42,7 +42,7 @@ class AdminDirectoryClient(_base_client.BaseClient):
             api_name=self.API_NAME)
 
         self.rate_limiter = RateLimiter(
-            self.global_configs.get('max_admin_api_calls_per_day'),
+            self.global_configs.get('max_admin_api_calls_per_100_secs'),
             self.DEFAULT_QUOTA_TIMESPAN_PER_SECONDS)
 
     def _build_credentials(self, global_configs):
--- a/google/cloud/security/common/gcp_api/admin_directory.py
+++ b/google/cloud/security/common/gcp_api/admin_directory.py
@@ -42,7 +42,7 @@ class AdminDirectoryClient(_base_client.BaseClient):
             api_name=self.API_NAME)
 
         self.rate_limiter = RateLimiter(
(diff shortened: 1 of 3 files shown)


google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, global_configs):
             api_name=self.API_NAME)
 
         self.rate_limiter = RateLimiter(
-            self.global_configs.get('max_admin_api_calls_per_day'),
+            self.global_configs.get('max_admin_api_calls_per_100_seconds'),
```
ahoying: What does this do to existing deployments if this new config entry isn't in the yaml? Is there a path to update existing configurations while still accepting the previous configuration format?

mirons-google: We don't have a good process. The current plan is to let things break and put a notice in the release notes.

carise: I have some ideas that I've sent out in a doc. Probably doesn't need to be implemented in this PR, but it would be nice to nail down a process for the future.

blueandgold: Awesome, this is part of Carise's work to define a change management process.

blueandgold: I can't find carise's change management doc yet, but would a check for forseti version number help to manage this change here?
```
   if version <= 1.1.0:
       then exit
```

carise: I've sent the doc out to the team, but yes, essentially it's fine to put the checks in the code for now. The code should handle both config directives and log a warning if a directive will be deprecated in the next version possibly. Let's not try to complicate things for now (ie. don't make a framework for handling deprecation and stuff...) and try to work through a few versions before we actually decide if we need to do more engineering around config management.

ahoying: I'm working this change into my mixins PR, if you want to hold off Michael and see how it looks there?

mirons-google: SGTM

PR #562, opened 2017-08-17
New API base method based on repository model with mix-ins.

### LINKED ISSUE #508: Bump Folders version to v2
https://cloud.google.com/resource-manager/reference/rest/v2/folders

ahoying:

- [x] All of the [unit-tests](http://forsetisecurity.org/docs/development/#executing-tests) still pass

resolves #508


ahoying:
I have a few more tests to add, and need to remerge changes from dev, but other than that this should all be ready for review.

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -14,69 +14,134 @@
 
 """Wrapper for Admin Directory  API client."""
 
-from oauth2client.service_account import ServiceAccountCredentials
-from ratelimiter import RateLimiter
-
-from google.cloud.security.common.gcp_api import _base_client
+from googleapiclient import errors
+from httplib2 import HttpLib2Error
+from google.cloud.security.common.gcp_api import _base_repository
```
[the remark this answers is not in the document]
ahoying: Done.

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,698 @@
...
+              get methods.
+          max_results_field (str): The field name that represents the maximum
+              number of results to return in one page.
+          search_query_field (str): The field name used to filter search
+              results.
+          rate_limiter (object): A RateLimiter object to manage API quota.
+        """
+        self.gcp_service = gcp_service
+        self._credentials = credentials
+        components = component.split('.')
```
atcherniakhovski: There might be a more general way of attaching components.
Concretely, discovery.build sets__is_resource__ attribute on every nested component (ex. firewalls, networks, etc).
setattr(methodResource, '__is_resource__', True)
Therefore, we can inspect the passed service object and dynamically attach all nested repositories.

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,698 @@
...
+            req_body = verb_arguments.get('body', dict())
+            if next_page_token:
+                req_body['pageToken'] = next_page_token
+            request = self._build_request(verb, verb_arguments)
+            response = self._execute(request)
+            number_of_pages_processed += 1
+            LOGGER.debug('Executing paged request #%s',
+                         number_of_pages_processed)
+            next_page_token = response.get('nextPageToken')
+            yield response
```
atcherniakhovski: We should probably yield response['items'] - items being the item key. This implies, that the item key needs to be passed in.

google/cloud/security/common/gcp_api/_base_repository.py
ahoying: That would be interesting to explore to further reduce the boilerplate code for the repository wrapper classes. I'm going to leave as is for now, but I agree that would be useful.

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -14,69 +14,134 @@
...
+                _AdminDirectoryMembersRepository)
 
-    def _build_credentials(self, global_configs):
-        """Build credentials required for accessing the directory API.
+        return self._members
+    # pylint: enable=missing-return-doc, missing-return-type-doc
+
+
+class _AdminDirectoryGroupsRepository(
+        _base_repository.GCPRepository,
```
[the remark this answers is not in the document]
ahoying: Good catch, thanks!

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,698 @@
...
+        if kwargs:
+            arguments.update(kwargs)
+
+        try:
+            for resp in self.execute_paged_query(verb=verb,
+                                                 verb_arguments=arguments):
+                yield resp
+        except api_errors.PaginationNotSupportedError:
+            # Some API list() methods are not actually paginated.
+            del arguments[self._max_results_field]
```
atcherniakhovski: I think this logic should be moved into a decorator  - let's know raise an exception for the case we know how to deal with. The decorator would check if the method supports pagination and del max_results_field.

google/cloud/security/common/gcp_api/_base_repository.py
ahoying: I like keeping these as close to the raw API response as possible, and letting the higher level implementations handle extracting the data from the response.

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,695 @@
...
+            return repository_class(gcp_service=self.gcp_services[version],
+                                    credentials=self._credentials,
+                                    rate_limiter=self._rate_limiter)
+
+
+# pylint: disable=too-many-instance-attributes, too-many-arguments
+class GCPRepository(object):
+    """Base class for GCP APIs."""
+
+    def __init__(self, gcp_service, credentials, component, entity='',
```
atcherniakhovski: Is an empty string a reasonable default for entity?

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,695 @@
...
+        Returns:
+          dict: GCE response.
+
+        Raises:
+          errors.HttpError: When attempting to get a non-existent entity.
+           ex: HttpError 404 when requesting ... returned
+            "The resource '...' was not found"
+        """
+        assert isinstance(self, GCPRepository)
+        assert bool(self._get_key_field)
```
atcherniakhovski: :) The reason for the assert on GCPRepository was to get PyCharm figure out that self is GCPRepository and  provide proper code completion support.
It appears that this could as easily be achieved via a docstring for self.
self (GCPRepository): GCPRepository.

In other words, we should probably remove these asserts; Asserts on types are typically frowned upon :(

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,695 @@
...
+        return self.execute_query(
+            verb=verb,
+            verb_arguments=arguments,
+        )
+
+class SearchQueryMixin(object):
+    """Mixin that implements Search query."""
+
+    def search(self, query=None, fields=None, max_results=500, verb='search'):
+        """Get all organizations the caller has access to.
```
atcherniakhovski: I think the comment does not align the the method.

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,695 @@
...
+        Yields:
+          dict: Service Response.
+        """
+        # Implementation of search does not follow the standard API pattern.
+        # Fields need to be in the body rather than sent seperately.
+        next_page_token = None
+        number_of_pages_processed = 0
+        while True:
+            req_body = verb_arguments.get('body', dict())
+            if next_page_token:
```
atcherniakhovski: This looks very similar to paged_query, why do we need both?

ahoying: The search APIs require nextPageToken to be inside of a body argument, not a top level parameter. And they do not implement the {verb}_next() method. So they need a custom implementation.

google/cloud/security/common/gcp_api/_base_repository.py
ahoying: Some APIs do not use the entity target for Get, just another inconsistency.

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,695 @@
...
+          errors.HttpError: When attempting to get a non-existent entity.
+           ex: HttpError 404 when requesting ... returned
+            "The resource '...' was not found"
+        """
+        assert isinstance(self, GCPRepository)
+        assert bool(self._get_key_field)
+
+        arguments = {self._get_key_field: resource,
+                     'fields': fields}
+        if self._entity and target:
```
atcherniakhovski: I understand that the special case of Projects forces this complexity.
I think a comment would be helpful here to explain the difference between entity, target, get_key_field.
Specifically, the case where entity could be None.

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -13,70 +13,135 @@
...
 
-class AdminDirectoryClient(_base_client.BaseClient):
-    """GSuite Admin Directory API Client."""
 
-    API_NAME = 'admin'
-    DEFAULT_QUOTA_TIMESPAN_PER_SECONDS = 86400 # pylint: disable=invalid-name
-    REQUIRED_SCOPES = frozenset([
-        'https://www.googleapis.com/auth/admin.directory.group.readonly'
-    ])
+class AdminDirectoryRepository(_base_repository.BaseRepositoryClient):
```
atcherniakhovski: I feel that names of the derived and base classes are misaligned. I would expect the name of this class to be AdminDirectoryRepositoryClient.

ahoying: Good call, fixing.

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,699 @@
...
+def _create_service_api(credentials, service_name, version, developer_key=None,
+                        cache_discovery=False):
+    """Builds and returns a cloud API service object.
+
+    Args:
+        credentials (OAuth2Credentials): Credentials that will be used to
+            authenticate the API calls.
+        service_name (str): The name of the API.
+        version (str): The version of the API to use.
+        developer_key (str): The api key to use, .
```
mirons-google: nit: The sentence is incomplete.

google/cloud/security/common/gcp_api/_base_repository.py
atcherniakhovski: Yes, but I feel that the default should reflect the common case, which I believe to be not an empty string. In other words, it might be better to let the exceptions supply an empty string.
At any rate, in the future, we could detect such edge cases via inspection. Is Projects the only edge case at the moment?

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,699 @@
...
+        item_key (str): The name of the key within the inner "items" lists
+            containing the objects of interest.
+
+    Returns:
+        list: A list of items.
+    """
+    results = []
+    for page in paged_results:
+        results.extend(page.get(item_key, []))
+    return results
```
mirons-google: super-nit: Sometimes you leave a new line before the return, other time's you don't. Consistency is probably good here.

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,699 @@
...
+            dict: The response from the API.
+        """
+        if self._rate_limiter:
+            # Since the ratelimiter library only exposes a context manager
+            # interface the code has to be duplicated to handle the case where
+            # no rate limiter is defined.
+            with self._rate_limiter:
+                return request.execute(http=self.http,
+                                       num_retries=self._num_retries)
+        else:
```
mirons-google: The else isn't necessary here. Dropping it could improve readability.

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,699 @@
+# Copyright 2017 Google Inc.
```
mirons-google: Is it worth extracting out the mix-in classes from this file. It's long in the tooth and could provide improved readability.

google/cloud/security/common/gcp_api/admin_directory.py
```diff
@@ -13,70 +13,135 @@
...
+        Args:
+            global_configs (dict): Global configurations.
+            **kwargs (dict): The kwargs.
         """
-        return RateLimiter(
-            self.global_configs.get('max_admin_api_calls_per_day'),
-            self.DEFAULT_QUOTA_TIMESPAN_PER_SECONDS)
+        max_calls = global_configs.get('max_admin_api_calls_per_100_seconds')
+        quota_period = self.DEFAULT_QUOTA_TIMESPAN_PER_SECONDS
+        if not max_calls:
```
mirons-google: Maybe leave a TODO here to re-visit and remove at some point?

google/cloud/security/common/gcp_api/_base_repository.py
```diff
@@ -0,0 +1,713 @@
...
+    items = []
+    for page in paged_results:
+        aggregated_items = page.get('items', {})
+        for items_for_grouping in aggregated_items.values():
+            for item in items_for_grouping.get(item_key, []):
+                items.append(item)
+    return items
+
+
+class BaseRepositoryClient(object):
```
atcherniakhovski: General comment:
This class feels like a factory class for GCP Repositories. Furthermore, it is not intuitive to have abstractions for both: Repositories and RepositoryClients. I would suggest (in future releases) to consider transitioning BaseRepositoryClient into a factory class for GCP Repositories.

ahoying:
Waiting for blueandgold to run a live test of the branch before merging.

blueandgold:
Andrew and I have done live testing of this PR against the dev branch.  Inventory data in both sets of tables are the same.  So, this PR is safe to merge.  Great job Andrew!

PR #582, opened 2017-08-22
Fix parameterized query

### LINKED ISSUE #510: Fix the sql statements where we use .format() for !timestamps
This is to prevent sql injections, e.g. https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/data_access/bucket_dao.py#L61

The proper way to create the prepared statements is: 
https://github.com/GoogleCloudPlatform/forseti-security/blob/master/google/cloud/security/common/data_access/folder_dao.py#L85

@felixbb

carise:
resolves #510


carise:
Thanks for the review!

PR #585, opened 2017-08-23
Change Copyright Notice to Be Shared Among Many Contributors.

blueandgold:
Since we have external contributions, "Copyright Google" becomes obsolete. So, this means the copyright notice needs to be updated so that the copyright of the collective work is shared among the contributors.

This change is based on these examples:
https://github.com/chromium/crashpad/blob/master/AUTHORS
https://github.com/chromium/crashpad/blob/master/CONTRIBUTORS
https://github.com/chromium/crashpad/blob/master/crashpad.gyp

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,10 @@
+# This is the official list of Crashpad authors for copyright purposes.
+# This file is distinct from the CONTRIBUTORS files.
+# See the latter for an explanation.
+
+# Names should be added to this file as:
+# Name or Organization <email address>
+# The email address is not required for organizations.
+
+Google Inc.
+Spotify AB
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,4 @@
-# This is the official list of Crashpad authors for copyright purposes.
+# This is the official list of Forseti Security authors for copyright purposes.
 # This file is distinct from the CONTRIBUTORS files.
 # See the latter for an explanation.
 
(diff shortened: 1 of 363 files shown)


AUTHORS
```diff
@@ -0,0 +1,10 @@
+# This is the official list of Forseti Security authors for copyright purposes.
+# This file is distinct from the CONTRIBUTORS files.
+# See the latter for an explanation.
+
+# Names should be added to this file as:
+# Name or Organization <email address>
+# The email address is not required for organizations.
+
+Google Inc.
+Spotify AB
```
carise: Do we add Denzel here as well?

blueandgold: We can.  But according to the link below, it seems like he would have to initiate the request to be a copyright holder.
https://g3doc.corp.google.com/company/teams/opensource/releasing/authors.md?cl=head

carise: Makes sense. Removed the internal link, btw

CONTRIBUTORS
```diff
@@ -0,0 +1,26 @@
...
+#
+# Names should be added to this file as:
+#     Name <email address or github username>
+
+Adam Cotenoff <acotenoff>
+Andrew Hoying <ahoying>
+Anton Sapozhnikov <snowytoxa>
+Carise Fernandez <carise>
+Carly Schneider <cschnei3)
+Denzel Morris
```
mirons-google: denzel-morris

PR #589, opened 2017-08-23
Fix wizard verbiage

### LINKED ISSUE #587: Setup wizard verbiage
* Create Forseti deployment has a message that "This may take a while." It would be good to specify "This may take several minutes."
* When we prompt for the SendGrid API key, include text that user can set it up later if they skip it now.
* Copy G Suite key step displays "Attempt #1" - change it to "Attempt 1 of #" with the max number of attempts.
* After setup completes, include a link to set up email notification later.

from @Tabrarian

carise:
resolves #587


carise:
Thanks for the review!

PR #593, opened 2017-08-24
Documentation Update for Forseti Quickstart

danieldeleo:
I removed the "git checkout master" line in order to cut down on the steps for quickstart. The master branch is the default branch when users run the git clone so it felt redundant for a quickstart.

Also added some clarity around needing to run inventory setup first before the scanner and enforcer

### DIFF
--- a/_docs/howto/deploy/change-gcp-deployment.md
+++ b/_docs/howto/deploy/change-gcp-deployment.md
@@ -6,7 +6,8 @@ order: 104
 
 This page describes how to change a Forseti Security deployment. Most changes
 to a deployment are usually to the deployment properties, such as the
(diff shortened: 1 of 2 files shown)


carise:
Hey Danny, thanks for the PR! With the current instructions in the quickstart, the user will get the dev branch (we recently switched the default branch to dev). Maybe a better command would be:

```
git clone -b master --single-branch https://github.com/GoogleCloudPlatform/forseti-security
```

What do you think?

carise:
Also wrt the inventory/scanner/enforcer, the quickstart setup sets them up automatically. The next step for the user would be to configure the tools, if they need to.

danieldeleo:
Wow you guys totally improved the setup! Sorry I hadn't re-run the setup in a while. Ok I'll make changes accordingly.

mirons-google:
General comment once this is ready for review, please wrap on 80 characters, thanks!

Issue #598, opened 2017-08-28
Include logwriter role in list of roles to enable on project


carise:
- [x] Add to http://forsetisecurity.org/docs/guides/best-practices#service-account-for-forseti-security
 - [x] update wizard
 - [x] Look into changing log format to add a [forseti] tag/label.

PR #603, opened 2017-08-29
Add logwriter role

carise:
related: #598


carise:
yeah, I can't exactly put both changes into the same PR.

carise:
Actually I just noticed that the docs are already updated.

PR #605, opened 2017-08-30
Update change-gcp-deployment.md

carise:


_docs/howto/deploy/change-gcp-deployment.md
```diff
@@ -1,28 +1,66 @@
...
-to a deployment are usually to the deployment properties, such as the
-instance type or `src-path` when you want your deployment to run a certain 
-version of Forseti Security. To update your deployment, follow the process below:
+### Update local installation
 
-  1. Edit `deploy-forseti.yaml` and update the values you want to change.
+If you want to update your local version of Forseti, pull in the latest changes from
+git:
+
+```
```
mirons-google: can we use the ```bash type (here and throughout)? We've also been prepending the $ to distinguish distinct lines.

_docs/howto/deploy/change-gcp-deployment.md
```diff
@@ -1,28 +1,66 @@
...
+
+And then run the python installation:
+
+```
+python setup.py install
+```
+
+### Change deployment properties
+
```
mirons-google: nit: Style has been no new-line after titles.

_docs/howto/deploy/change-gcp-deployment.md
```diff
@@ -1,28 +1,64 @@
...
+     also require an updated startup script on the GCE instance.
+     
+     To get the latest version of Forseti, sync your master branch:
+     
+     ```bash
+     $ git checkout master
+     $ git pull
+     ```
+     
+  1. Check `deploy-forseti.yaml.sample` to see if there are any new properties
```
Tabrarian: How do I identify if there are new properties?

_docs/howto/deploy/change-gcp-deployment.md
```diff
@@ -1,28 +1,64 @@
...
+
+```bash
+$ python setup.py install
+```
+
+### Change deployment properties
+If you need to change any of the properties for your Forseti deployment, such as the
+`release-version`, follow the process below:
+
+  1. If you **do not** need to update your Forseti version, continue on to step 3.
```
Tabrarian: This could get confusing in future edits. It might be best to use an include for the common steps and have two separate sections - one in which I update Forseti version, and another where I don't.

That said - how do I know if I need to update my Forseti version?

_docs/howto/deploy/change-gcp-deployment.md
```diff
@@ -1,28 +1,64 @@
...
-If you change the properties in the `deploy-forseti.yaml` "Compute Engine" 
+If you changed the properties in the `deploy-forseti.yaml` "Compute Engine" 
 section or the startup script in `forseti-instance.py`, you need to reset 
 the instance for changes to take effect:
 
   ```bash
   $ gcloud compute instances reset COMPUTE_ENGINE_INSTANCE_NAME
   ```
 
+The GCE instance will restart and perform a fresh installation of Forseti, so you do
```
Tabrarian: Compute Engine

_docs/howto/deploy/change-gcp-deployment.md
```diff
@@ -1,28 +1,64 @@
...
+     ```
+     
+  1. Check `deploy-forseti.yaml.sample` to see if there are any new properties 
+     that you need to copy over to your previous `deploy-forseti.yaml`.
+  1. Edit `deploy-forseti.yaml` and update the values you want to change. If you 
+     previously deployed the `master` branch, you don't need to change it.
   1. Run the following update command:
 
       ```bash
       $ gcloud deployment-manager deployments update DEPLOYMENT_NAME \
```
Tabrarian: It's generally best not to include a $ to indicate you're at a command line. The reason is that we want users to be able to copy-paste the relevant commands. If a user pastes "$", the command fails.

_docs/howto/deploy/change-gcp-deployment.md
```diff
@@ -1,28 +1,64 @@
...
 the instance for changes to take effect:
 
   ```bash
   $ gcloud compute instances reset COMPUTE_ENGINE_INSTANCE_NAME
   ```
 
+The GCE instance will restart and perform a fresh installation of Forseti, so you do 
+not need to ssh to the instance to run all the git clone/python install commands.
+
+There are limitations as to what resources you can update in your deployment. If you
```
Tabrarian: Is it likely that I'll need to create a new deployment, or is it a certainty that if I get an error that I can't change a resource, I will have to create a new deployment?

If I have to create a new deployment, do I start from scratch with http://forsetisecurity.org/docs/howto/deploy/gcp-deployment or http://forsetisecurity.org/docs/howto/deploy/local-deployment (could be good to link), or is there a different method?

_docs/howto/deploy/change-gcp-deployment.md
carise: @mirons-google

_docs/howto/deploy/change-gcp-deployment.md
carise: Yes, if you are trying to change an immutable property, deployment update will fail. (see the official docs link below)

Since this is the case of Deployment Manager, the user will either use GCP deployment or the setup wizard.

_docs/howto/deploy/change-gcp-deployment.md
carise: The user will find out about Forseti updates by coming to the Github website. We don't push users to get the newest release version.

_docs/howto/deploy/change-gcp-deployment.md
carise: Use diff

_docs/howto/deploy/change-gcp-deployment.md
mirons-google: This is my fault. I realize it breaks the copy paste. I was hoping to change the copy-code button logic to stop at the $. Until that's changed you're correct that this is the better thing to do.

Tabrarian: I think even if we change the copy button logic, it may still be best to exclude the $ - particularly in cases where it's two separate commands, users will probably manually copy the first, run it, then manually copy the second. Example from above:
```
$ git checkout master
$ git pull
```
For cloud.google.com docs, we have a class that excludes the system prompt from mouse select. Would something like that be possible here?

_docs/howto/deploy/change-gcp-deployment.md
Tabrarian: Thanks; given that, I'd want to use more definitive language here. "Some resources can't be updated in a deployment. If you see an error that you can't change a certain resource, you'll need to create a new [GCP deployment]() or [local deployment]()."

_docs/howto/deploy/change-gcp-deployment.md
Tabrarian: Gotcha, okay, so what I'm getting here is that updating the local installation is optional. If you do that, then you need to do steps 1-2 to update Forseti version. Then you edit deploy-forseti.yaml to make the other changes you want. What I'd suggest is to make separate headings for those, along these lines:

```
## Update local installation (optional)
  (list of steps)

### Update Forseti version
After you update your local installation, follow the steps below to update your Forseti version:
  (list of steps)
Next, continue to change your deployment properties.

## Change deployment properties
To update your deployment, follow the steps below:

1. Edit `deploy-forseti.yaml` ...
```
Let me know if you'd prefer me to submit an edit to the file instead of trying to make sense of my comments :)

_docs/howto/deploy/change-gcp-deployment.md
carise: This would be:

```
git diff origin..HEAD~1 -- deploy-forseti.yaml.sample
```

_docs/howto/deploy/change-gcp-deployment.md
carise: Sorry wasn't clearer -- updating local installation isn't a step; it's for people who want to update their local installation.

To update the version for a Deployment Manager deployment, you should:

1. Pull down the latest code and checkout the appropriate version (git command)
2. Update necessary deployment template (deploy-forseti.yaml), including the Forseti version
3. Run the deployment manager update (gcloud command)


This is basically the same thing as changing a deployment. The only explicit instruction is telling the user to update the Forseti version property.

PR #609, opened 2017-08-30
Add the log tag [forseti] to syslog entries

### LINKED ISSUE #598: Include logwriter role in list of roles to enable on project

carise:
logtag

resolves #598


mirons-google:
nit: Could we make this forseti-security to future proof ourselves from want to use `forseti` for some other reason?

PR #630, opened 2017-09-08
Updating the website footer

mirons-google:


_includes/site/layout/base/footer/links.html
```diff
@@ -11,11 +11,14 @@
...
     <li><a href="{% link about/index.md %}">About</a></li>
     <li><a href="{% link faq/index.md %}">FAQ</a></li>
+    <li><a href="https://www.google.com/policies/terms/">Website Terms</a></li>
+    <li><a href="https://www.google.com/policies/privacy/">Website Privacy Policy</a></li>
   </ul>
 </div>
 <div class="col-sm-4">
   <h2>Community</h2>
   <ul class="links">
+    <li><a href="{% link community/index.md %}">Forseti</a></li>
```
carise: I think I understand why you added this link, but I feel like it's a bit redundant with the Community link that we have in the Header. Is there a way we could make the description more descriptive (i.e. not "Forseti")?

_includes/site/layout/base/footer/links.html
```diff
@@ -11,11 +11,14 @@
   <ul class="links">
     <li><a href="{% link about/index.md %}">About</a></li>
     <li><a href="{% link faq/index.md %}">FAQ</a></li>
+    <li><a href="https://www.google.com/policies/terms/">Website Terms</a></li>
```
carise: Does Website refer to forsetisecurity.org?

mirons-google: Yup, suggestions on different wording if that isn't preferred?

_includes/site/layout/base/footer/links.html
mirons-google: Sure, do you have a suggestion? While editing I noticed we had links to every other top-level heading in the footer except this one. Given that the heading is named Community I chose Forseti. We could rename the heading...

_includes/site/layout/base/footer/links.html
carise: I think it sounds okay as is.

_includes/site/layout/base/footer/links.html
carise: Ah, makes sense. I kind of like how Documentation breaks out the sub categories of documentation, though I see how there's a docs Overview link (which is the header Docs link). What if instead of linking to the Community page in the footer, we link directly to the "Discuss" Google Group, and call the link "Google Groups"?

_includes/site/layout/base/footer/links.html
mirons-google: Done

carise:
👍

PR #631, opened 2017-09-08
Fix error handler.

blueandgold:
This properly handles the error when the admin directory api client can not be created due to missing gsuite key, so that we can get the proper logging message to understand what happened.

```
2017-09-08 16:29:11,241 google.cloud.security.inventory.pipeline_builder ERROR    _get_api Failed to initialize API AdminDirectoryClient, v=None
('Error building admin api credential: %s', IOError(2, 'No such file or directory'))
```

@ahoying, AFAICT, this will not affect the rest of the api clients.  Can you please confirm?

### DIFF
--- a/google/cloud/security/inventory/pipeline_builder.py
+++ b/google/cloud/security/inventory/pipeline_builder.py
@@ -93,7 +93,7 @@ class PipelineBuilder(object):
             except api_errors.ApiInitializationError as e:
                 LOGGER.error('Failed to initialize API %s, v=%s\n%s',
                              api_class_name, api_version, e)
-                raise api_errors.ApiInitializationError(api_class_name, e)
+                raise api_errors.ApiInitializationError(e)
 
             self.initialized_api_map[api_name] = api
 
--- a/google/cloud/security/inventory/pipeline_builder.py
+++ b/google/cloud/security/inventory/pipeline_builder.py
@@ -61,6 +61,9 @@ class PipelineBuilder(object):
 
         Returns:
             object: Instance of the API.
+
+        Raises:
+            ApiInitializationError: The api client can not be created.
         """
         api = self.initialized_api_map.get(api_name)
         if api is None:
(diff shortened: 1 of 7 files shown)


google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -90,8 +90,8 @@ def _get_api(self, api_name):
                 else:
                     api = api_class(self.global_configs,
                                     version=api_version)
-            except api_errors.ApiExecutionError as e:
-                LOGGER.error('Failed to execute API %s, v=%s\n%s',
+            except api_errors.ApiInitializationError as e:
+                LOGGER.error('Failed to initialize API %s, v=%s\n%s',
                              api_class_name, api_version, e)
                 raise api_errors.ApiInitializationError(e)
```
ahoying: This can just be the following, since you are re-raising the same exception now:

raise

google/cloud/security/inventory/pipeline_builder.py
```diff
@@ -90,8 +90,8 @@ def _get_api(self, api_name):
                 else:
                     api = api_class(self.global_configs,
                                     version=api_version)
-            except api_errors.ApiExecutionError as e:
-                LOGGER.error('Failed to execute API %s, v=%s\n%s',
+            except api_errors.ApiInitializationError as e:
```
ahoying: I think you should have an except for both ApiExecutionError and ApiInitializationError. That way you can recover or propagate either type of error.

blueandgold: I would do it, but at this point, we really are just initializing the api clients and not executing them.  Can you please confirm if this is still what would be nice to do?

Changed the raise statement to just a single ```raise```.

ahoying: So long as that's the only exception that we care about at this stage and we know we've consistently made sure that's the exception that is raised, then that makes sense. I think longer term there is a lot more work to do around exception handling and ensuring all APIs are consistent. I tried to make things more consistent in my mixins PR, but I know more work needs to be done.

PR #632, opened 2017-09-11
Update storage client to handle retries and explicit credentials.

ahoying:
* Set file_loader util to use an explicit credential object so that
per thread http object caching is not used for loading and saving files
to a GCS bucket.

### DIFF
--- a/google/cloud/security/common/util/file_loader.py
+++ b/google/cloud/security/common/util/file_loader.py
@@ -86,8 +86,8 @@ def _read_file_from_gcs(file_path):
         dict: The parsed dict from the loaded file.
     """
     # Pass credential in explicitly so that cached credentials are not used.
-    credential = client.GoogleCredentials.get_application_default()
-    storage_client = storage.StorageClient(credential=credential)
+    credentials = client.GoogleCredentials.get_application_default()
+    storage_client = storage.StorageClient(credentials=credentials)
 
     file_content = storage_client.get_text_file(full_bucket_path=file_path)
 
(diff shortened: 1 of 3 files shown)


google/cloud/security/common/util/file_loader.py
```diff
@@ -157,3 +166,16 @@ def _parse_yaml(data):
...
+def _get_storage_client():
+    """Returns a new storage API client with explicit credentials.
+
+    Returns:
+        storage.StorageClient: A new Storage API client with explicit
+            credentials.
+    """
+    # Pass credential in explicitly so that cached credentials are not used.
+    credentials = client.GoogleCredentials.get_application_default()
+    storage_client = storage.StorageClient(credentials=credentials)
```
blueandgold: I am sure I am missing something, so please help me to understand how this is different from the previous existing code.

So before, we create the storage_client like this:
```storage_client = storage.StorageClient()```

Which in turn initialize the super ```_base_repository.BaseRepositoryClient()```

The super also end up getting the application default credential in the __init__().
https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/security/common/gcp_api/_base_repository.py#L130

ahoying: The difference is the setting of the internal _use_cached_http flag, which is only set to True when credentials are not passed in. By passing in credentials it forces the client to generate a new http object for each request instead of using a cached one. That's the behavior we want for the file_loader.

Now this is not the only way that could be accomplished, but it does require the least code changes. If you think it's too confusing I can look at a deeper refactor of the logic.

blueandgold: LGTM

PR #637, opened 2017-09-14
Handle unicode when writing to csv.

blueandgold:
Python 2’s csv module doesn’t handle unicode strings.  So, opting to change to a serializer that handles unicode natively.  I believe this is much better since we don't have to do this ourselves, or handle complex nested dictionaries.

Here is a screenshot that this is written correctly to the database.
image

https://pypi.python.org/pypi/unicodecsv/0.14.1


blueandgold:
Antoine Castex verified this fixed his issue.

PR #646, opened 2017-09-15
Add appengine services and versions to Inventory.

blueandgold:
After looking at the data in the responses, it seems the best way to incorporate the services/versions data is to associate the versions to each service, and then put the whole services payload into a json column named ```services```.  Of course the ```services``` payload is part of the raw json column.

I will add testing if this approach is good.  Otherwise, please let me know what else might work.


google/cloud/security/common/data_access/sql_queries/create_tables.py
```diff
@@ -30,6 +30,7 @@
         `default_bucket` varchar(255) DEFAULT NULL,
         `iap` json DEFAULT NULL,
         `gcr_domain` varchar(255) DEFAULT NULL,
+        `services` json DEFAULT NULL,
```
ahoying: I'd prefer to see an APPENGINE_SERVICES, APPENGINE_VERSIONS and APPENGINE_INSTANCES tables added, with just json columns for the main data, and the required keys:

 CREATE_APPENGINE_SERVICES_TABLE = """
     CREATE TABLE `{0}` (
         `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
         `project_id` varchar(255) DEFAULT NULL,
         `app_id` varchar(255) DEFAULT NULL,
         `service_id` varchar(255) DEFAULT NULL,
         `service` json DEFAULT NULL,
         PRIMARY KEY (`id`),
         UNIQUE KEY `service_key` (`app_id`, `service_id`)
     ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
"""

CREATE_APPENGINE_VERSIONS_TABLE = """
     CREATE TABLE `{0}` (
         `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
         `project_id` varchar(255) DEFAULT NULL,
         `app_id` varchar(255) DEFAULT NULL,
         `service_id` varchar(255) DEFAULT NULL,
         `version_id` varchar(255) DEFAULT NULL,
         `version` json DEFAULT NULL,
         PRIMARY KEY (`id`),
         UNIQUE KEY `version_key` (`app_id`, `service_id`, `version_id`)
     ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
"""

CREATE_APPENGINE_INSTANCES_TABLE = """
     CREATE TABLE `{0}` (
         `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
         `project_id` varchar(255) DEFAULT NULL,
         `app_id` varchar(255) DEFAULT NULL,
         `service_id` varchar(255) DEFAULT NULL,
         `version_id` varchar(255) DEFAULT NULL,
         `instance_id` varchar(255) DEFAULT NULL,
         `instance` json DEFAULT NULL,
         PRIMARY KEY (`id`),
         UNIQUE KEY `instance_key` (`app_id`, `service_id`, `version_id`, `instance_id`)
     ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
"""
Then the pipeline can output the correct data to each table. That will be closer to the one row per resource that the new inventory will output.

blueandgold: Got it, will work on this!

blueandgold: Done.  The only oddity is that it's easier to return the services, versions, instances data from retrieve() in loadable form.  Versus forcing them into the app data, and then parse them out again in transform().

google/cloud/security/inventory/pipelines/load_appengine_pipeline.py
```diff
@@ -78,7 +128,16 @@ def _transform(self, resource_from_api):
...
-        self._load(self.RESOURCE_NAME, loadable_apps)
+        apps, loadable_services, loadable_versions, loadable_instances = (
+            self._retrieve())
+
+        if apps:
+            loadable_apps = self._transform(apps)
+            self._load(self.RESOURCE_NAME, loadable_apps)
         self._get_loaded_count()
+
+        # TODO: Make _get_loaded_count() support multiple resources in a single
```
ahoying: Nit: You could put this under the if apps: since if there are no apps there won't be any other resources either.

blueandgold: Done!

PR #668, opened 2017-09-26
Put back json.dumps() for raw_data.

### LINKED ISSUE #665: repr instead of json.dumps in instance network interface scanner
Hello,
looks like in `dev` the network_interface_scanner is using `repr` insted of `json.dumps`to save the `raw_data` (within `violation_data`): https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/security/scanner/audit/instance_network_interface_rules_engine.py#L246

But according to a previous PR this should not be the case: https://github.com/GoogleCloudPlatform/forseti-security/pull/639/files#diff-52fba022d2d7ec64b70d687d13803eb6R247

We missed a commit perhaps?
Can you look into this?

blueandgold:
This fixes #665 , where json.dumps() was lost.

### DIFF
--- a/google/cloud/security/scanner/audit/instance_network_interface_rules_engine.py
+++ b/google/cloud/security/scanner/audit/instance_network_interface_rules_engine.py
@@ -244,7 +244,8 @@ class Rule(object):
                     project=project,
                     network=network,
                     ip=ips,
-                    raw_data=json.dumps(instance_network_interface, indent=2))
+                    raw_data=json.dumps(instance_network_interface.__dict__,
+                                        indent=2))
 
     # Rule violation.
     # resource_type: string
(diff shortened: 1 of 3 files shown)


google/cloud/security/scanner/audit/instance_network_interface_rules_engine.py
```diff
@@ -243,7 +244,8 @@ def find_violations(self, instance_network_interface_list):
                     project=project,
                     network=network,
                     ip=ips,
-                    raw_data=repr(instance_network_interface))
+                    raw_data=json.dumps(instance_network_interface.__dict__,
```
ahoying: Why not store the original JSON inside of the instance_network_interface object as a private attribute with a getter property.

In common/gcp_type/instance.py, class InstanceNetworkInterface.__init__() set self._json = json.dumps(kwargs, sort_keys=True)

Then add a new method:

def as_json(self):
  return _json

Then you can set raw_data to instance_network_interface.as_json()

Note that an as_json method would be a nice base method for all gcp_type objects.

blueandgold: Really nice suggestion, done!  Added a TODO for the base method.

gbrindisi:
thanks for this!

PR #686, opened 2017-10-04
Modified firewall rule to add policy comparisons and added tests.

mwwolters:

### DIFF
--- a/google/cloud/security/common/gcp_type/firewall_rule.py
+++ b/google/cloud/security/common/gcp_type/firewall_rule.py
@@ -42,19 +42,55 @@ class FirewallRule(object):
         self.network = kwargs.get('firewall_rule_network')
         self._priority = kwargs.get('firewall_rule_priority')
         self.direction = kwargs.get('firewall_rule_direction')
-        self.source_ranges = parser.json_unstringify(
-            kwargs.get('firewall_rule_source_ranges'))
-        self.destination_ranges = parser.json_unstringify(
-            kwargs.get('firewall_rule_destination_ranges'))
-        self.source_tags = parser.json_unstringify(
-            kwargs.get('firewall_rule_source_tags', '[]'))
-        self.target_tags = parser.json_unstringify(
-            kwargs.get('firewall_rule_target_tags', '[]'))
+        self._source_ranges = frozenset(parser.json_unstringify(
+            kwargs.get('firewall_rule_source_ranges', '[]')))
+        self._destination_ranges = frozenset(parser.json_unstringify(
+            kwargs.get('firewall_rule_destination_ranges', '[]')))
+        self._source_tags = frozenset(parser.json_unstringify(
+            kwargs.get('firewall_rule_source_tags', '[]')))
+        self._target_tags = frozenset(parser.json_unstringify(
+            kwargs.get('firewall_rule_target_tags', '[]')))
         self.allowed = parser.json_unstringify(
             kwargs.get('firewall_rule_allowed', '[]'))
         self.denied = parser.json_unstringify(
             kwargs.get('firewall_rule_denied', '[]'))
         self._firewall_action = None
+
+    @property
+    def source_ranges(self):
+        """The sorted source ranges for this policy.
+
+        Returns:
+          list: Sorted source ips ranges.
+        """
+        return sorted(self._source_ranges)
+
+    @property
+    def destination_ranges(self):
+        """The sorted destination ranges for this policy.
+
+        Returns:
+          list: Sorted destination ips ranges.
+        """
+        return sorted(self._destination_ranges)
+
+    @property
+    def source_tags(self):
+        """The sorted source tags for this policy.
+
+        Returns:
+          list: Sorted source tags.
+        """
+        return sorted(self._source_tags)
+
+    @property
+    def target_tags(self):
+        """The sorted target tags for this policy.
+
+        Returns:
+          list: Sorted target tags.
+        """
+        return sorted(self._target_tags)
 
     @property
     def priority(self):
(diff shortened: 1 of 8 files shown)


google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -66,3 +69,325 @@ def priority(self):
...
+          other (FirewallAction): Another FirewallAction.
+
+        Returns:
+          bool: Whether these two FirewallActions are functionally equivalent.
+        """
+        return (self.action == other.action and
+                self.rules.keys() == other.rules.keys() and
+                all([
+                    self.ports_are_equal(
+                        self.rules.get(protocol, []),
```
ahoying: This should use expanded_rules in place of rules for all comparisons.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -66,3 +69,325 @@ def priority(self):
...
+        """
+        if self._applies_to_all is None:
+            self._applies_to_all = False
+            for rule in self.rules:
+                protocol = rule.get('IPProtocol')
+                if protocol == 'all':
+                    self._applies_to_all = True
+                    return self._applies_to_all
+        return self._applies_to_all
+
```
ahoying: nit: remove extra line break.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -66,3 +69,325 @@ def priority(self):
...
+    def __eq__(self, other):
+        """Equals
+
+        Args:
+          other (FirewallAction): The FirewallAction to compare to.
+
+        Returns:
+          bool: If this action is the exact same as the other FirewallAction.
+        """
+        return self.action == other.action and self.rules == other.rules
```
ahoying: The dicts in self.rules and other.rules could be in different orders, and the ports could be in different orders.

What about in __init__() doing something like this:

self.rules = sort_rules(firewall_rule_allowed)
...

def sort_rules(rules):
sorted_rules = []
for rule in sorted(rules, key=lambda k: k.get('IPProtocol', '')):
if 'ports' in rule:
\# Sort ports numerically, handle ranges through sorting by start port
rule['ports'] = sorted(rule['ports'], key=lambda k: int(k) if '-' not in k else int(k.split('-')[0]))
sorted_rules.append(rule)
return sorted_rules

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -66,3 +69,325 @@ def priority(self):
...
+    can be of the form "<number>-<number>".
+
+    Args:
+      port_range (string): A string of format "<number_1>-<number_2>".
+
+    Returns:
+      list: A list of string integers from number_1 to number_2.
+    """
+    start, end = port_range.split('-')
+    return [str(i) for i in range(int(start), int(end) + 1)]
```
ahoying: Nit: You could use xrange here so the whole list isn't built twice.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -66,3 +69,325 @@ def priority(self):
...
+                self.rules.keys() == other.rules.keys() and
+                all([
+                    self.ports_are_equal(
+                        self.rules.get(protocol, []),
+                        other.rules.get(protocol, []))
+                    for protocol in self.rules
+                ]))
+
+    def __lt__(self, other):
+        """Less than
```
ahoying: Nit: First line of docstring should end with . (Same below)

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -66,3 +69,325 @@ def priority(self):
...
+        Returns:
+          bool: Whether this applies to all ports and protocols or not.
+        """
+        if self._applies_to_all is None:
+            self._applies_to_all = False
+            for rule in self.rules:
+                protocol = rule.get('IPProtocol')
+                if protocol == 'all':
+                    self._applies_to_all = True
+                    return self._applies_to_all
```
ahoying: This return could be a break; then you only need the one return at the end.

tests/common/gcp_type/firewall_rule_test.py
```diff
@@ -0,0 +1,433 @@
...
+                    [{'IPProtocol': 'tcp', 'ports': ['22']}]),
+            },
+        ),
+    ])
+    def test_firewall_action_init_error(self, action_1_dict):
+        """Tests that initializing a bad FirewallAction raises an error."""
+        with self.assertRaises(ValueError):
+            _ = firewall_rule.FirewallAction(**action_1_dict)
+
+    @parameterized.parameterized.expand([
```
ahoying: I recommend having one or more parameterized tests where there are multiple protocols and ports listed in a rule. Include them in different orders in self and other to ensure sorting works correctly. Include ports in different orders as well.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -17,10 +17,14 @@
 See: https://cloud.google.com/compute/docs/reference/latest/firewalls
 """
 
+import netaddr
```
ahoying: Your tests are failing because netaddr isn't included in the list of python modules to install in setup.py and scripts/docker/base

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -66,3 +105,359 @@ def priority(self):
...
+        """
+        return (self.direction == other.direction and
+                self.network == other.network and
+                self.source_tags == other.source_tags and
+                self.target_tags == other.target_tags and
+                self.source_ranges == other.source_ranges and
+                self.destination_ranges == other.destination_ranges and
+                self.firewall_action == other.firewall_action)
+
+    def is_equivalent(self, other):
```
ahoying: Add tests, please. You can see the code coverage and missed tests here: https://codecov.io/gh/GoogleCloudPlatform/forseti-security/pull/686/diff. There are a few other conditions below that are untested, add tests for them as well.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -66,3 +105,359 @@ def priority(self):
...
+
+        Args:
+          other(FirewallRule): object to compare to
+
+        Returns:
+          bool: comparison result
+        """
+        return (self.direction == other.direction and
+                self.network == other.network and
+                self.source_tags == other.source_tags and
```
ahoying: Nit: This could use self._source_tags == other._source_tags and then it's an O(1) operation to compare two frozensets, which is just a hash value compare. Otherwise you are creating lists and comparing them.

PR #690, opened 2017-10-09
Api additions get quota

therealsmartalec:

### DIFF
--- a/google/cloud/security/common/gcp_api/compute.py
+++ b/google/cloud/security/common/gcp_api/compute.py
@@ -694,6 +694,45 @@ class ComputeClient(object):
                 raise api_errors.ApiNotEnabledError(details, e)
             raise api_errors.ApiExecutionError(project_id, e)
 
+    def get_quota(self, project_id):
+        """Returns the quota for any metric
+
+        Args:
+            project_id (str): The project id.
+            metric
+
+        Returns:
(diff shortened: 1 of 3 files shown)


google/cloud/security/common/gcp_api/compute.py
```diff
@@ -800,6 +800,57 @@ def get_project(self, project_id):
...
+                  "metric": "FIREWALLS",
+                  "limit": 100.0,
+                  "usage": 9.0
+                }
+
+        Raises:
+            KeyError: Metric was not a firewall resource.
+        """
+        metric = 'FIREWALLS'
+        resource = self.get_quota(project_id, metric)
```
ahoying: You can return resource here without checking again since get_quota will raise the KeyError if the firewall quota does not exist. So you don't need to check if 'metric' is in resource below.

tests/common/gcp_api/compute_test.py
```diff
@@ -117,6 +117,21 @@ def test_get_firewall_rules_errors(self, name, response, status,
         with self.assertRaises(expected_exception):
             list(self.gce_api_client.get_firewall_rules(self.project_id))
 
+    def test_get_quota(self):
```
ahoying: Please add one more test where you call get_quota for a metric that does not exist and verify that a KeyError is raised.

PR #691, opened 2017-10-09
Added handling in FirewallRule and FirewallAction for empty fields.

mwwolters:
This makes it so the rules won't have to define each field but still can do
comparisons.

### DIFF
--- a/google/cloud/security/common/gcp_type/firewall_rule.py
+++ b/google/cloud/security/common/gcp_type/firewall_rule.py
@@ -242,10 +242,10 @@ class FirewallAction(object):
             raise ValueError(
                 'Firewall rule action must be either allow or deny, got: %s' % (
                     firewall_rule_action))
-        assert isinstance(firewall_rules, list)
         self.action = firewall_rule_action
         self._any_value = None
         if firewall_rules:
+            assert isinstance(firewall_rules, list)
             self.rules = sort_rules(firewall_rules)
         else:
             self.rules = []
--- a/google/cloud/security/common/gcp_type/firewall_rule.py
+++ b/google/cloud/security/common/gcp_type/firewall_rule.py
@@ -244,18 +244,14 @@ class FirewallAction(object):
                     firewall_rule_action))
         self.action = firewall_rule_action
         self._any_value = None
-        if firewall_rules == ['*']:
-            self._any_value = True
-            self.rules = []
-        elif firewall_rules:
+        if firewall_rules:
             self.rules = sort_rules(firewall_rules)
         else:
             self.rules = []
 
         self._applies_to_all = None
-        self._any_value = None
 
-        self._expanded_rules = {}
+        self._expanded_rules = None
 
     @property
     def applies_to_all(self):
@@ -291,11 +287,9 @@ class FirewallAction(object):
         Returns:
           dict: A dict of protocol to all port numbers.
         """
-        if self._expanded_rules:
+        if self._expanded_rules is None:
             self._expanded_rules = {}
-        if not self._expanded_rules:
             if not self.any_value:
-                self._expanded_rules = {}
                 for rule in self.rules:
                     protocol = rule.get('IPProtocol')
                     ports = rule.get('ports', ['all'])
--- a/google/cloud/security/common/gcp_type/firewall_rule.py
+++ b/google/cloud/security/common/gcp_type/firewall_rule.py
@@ -242,6 +242,7 @@ class FirewallAction(object):
             raise ValueError(
                 'Firewall rule action must be either allow or deny, got: %s' % (
                     firewall_rule_action))
+        assert isinstance(firewall_rules, list)
         self.action = firewall_rule_action
         self._any_value = None
         if firewall_rules:
--- a/google/cloud/security/common/gcp_type/firewall_rule.py
+++ b/google/cloud/security/common/gcp_type/firewall_rule.py
@@ -54,6 +54,7 @@ class FirewallRule(object):
             kwargs.get('firewall_rule_allowed', '[]'))
         self.denied = parser.json_unstringify(
             kwargs.get('firewall_rule_denied', '[]'))
+        self._action = kwargs.get('firewall_rule_action', 'allow')
         self._firewall_action = None
 
     @property
@@ -123,6 +124,7 @@ class FirewallRule(object):
             action_dict = {
                 'firewall_rule_allowed': self.allowed,
                 'firewall_rule_denied': self.denied,
+                'firewall_rule_action': self._action,
             }
             self._firewall_action = FirewallAction(**action_dict)
         return self._firewall_action
@@ -139,8 +141,10 @@ class FirewallRule(object):
         Returns:
           bool: comparison result
         """
-        return (self.direction == other.direction and
-                self.network == other.network and
+        return ((self.direction == other.direction or
+                 self.direction is None or
+                 other.direction is None) and
+                (self.network == other.network or other.network is None) and
                 set(self.source_tags).issubset(other.source_tags) and
                 set(self.target_tags).issubset(other.target_tags) and
                 self.firewall_action < other.firewall_action and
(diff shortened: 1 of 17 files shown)


google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -178,12 +185,13 @@ def __eq__(self, other):
...
                 self.network == other.network and
-                self.source_tags == other.source_tags and
-                self.target_tags == other.target_tags and
+                self._source_tags == other._source_tags and
+                self._target_tags == other._target_tags and
                 self.source_ranges == other.source_ranges and
                 self.destination_ranges == other.destination_ranges and
                 self.firewall_action == other.firewall_action)
 
+    # pylint: disable=protected-access
```
ahoying: At the end of the function add another comment to re-enable protected-access checks.

\# pylint: enable=protected-access

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -205,14 +213,16 @@ def is_equivalent(self, other):
...
+    def __init__(self, firewall_rule_allowed=None, firewall_rule_denied=None,
+                 firewall_rule_action='allow'):
         """Initialize.
 
         Args:
           firewall_rule_allowed (list): A list of dictionaries of allowed ports
             and protocols.
           firewall_rule_denied (list): A list of dictionaries of denied ports
             and protocols.
+          firewall_rule_action (str): The action, either allow or deny.
```
ahoying: Now that you have the action, does it make sense to just include firewall_rules and firewall_rule_action, and not separate out allow and deny rules into distinct arguments? It seems odd to currently be able to call FirewallAction(firewall_rule_denied=denied_rules) and have it be valid, even though firewall_rule_action is still defaulting to 'allow'.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -195,40 +213,49 @@ def is_equivalent(self, other):
...
+    ANY_VALUE = '*'
+
+    def __init__(self, firewall_rules=None, firewall_rule_action='allow'):
         """Initialize.
 
         Args:
-          firewall_rule_allowed (list): A list of dictionaries of allowed ports
-            and protocols.
-          firewall_rule_denied (list): A list of dictionaries of denied ports
+          firewall_rules (list): A list of dictionaries of allowed ports
```
ahoying: I don't like that the same variable can be either a list or a string ('*'). Can this always be a list, but if the list contains '\*' then it is treated as all rules match? See my comment below on using all() to test for that. Make sure  sort_rules can handle this case sanely as well, if it doesn't already.

Commenting here because it won't let me below.

You are missing a comma between 'IPProtocol' and '' in sort_rules, so it ends up returning None instead of an empty string if the key is missing.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -195,40 +213,49 @@ def is_equivalent(self, other):
...
+    # pylint: enable=protected-access
 
 
 class FirewallAction(object):
     """An association of allowed or denied ports and protocols."""
 
-    def __init__(self, firewall_rule_allowed=None, firewall_rule_denied=None):
+    ALLOW = 'allow'
+    DENY = 'deny'
+    ANY_VALUE = '*'
```
ahoying: I think these globals should not share names with properties in the object. Even though self.ANY_VALUE and self.any_value are different, it is confusing.

Maybe call this MATCH_ANY or call the property match_any?

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -195,40 +213,49 @@ def is_equivalent(self, other):
...
                 self.destination_ranges == other.destination_ranges and
                 self.firewall_action.is_equivalent(other.firewall_action))
+    # pylint: enable=protected-access
 
 
 class FirewallAction(object):
     """An association of allowed or denied ports and protocols."""
 
-    def __init__(self, firewall_rule_allowed=None, firewall_rule_denied=None):
+    ALLOW = 'allow'
```
ahoying: Since these are only used to verify that firewall_rule_action's value is valid, why not:

VALID_ACTIONS = frozenset(['allow', 'deny'])

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -249,12 +276,27 @@ def applies_to_all(self):
...
+
+    @property
     def expanded_rules(self):
         """Returns an expanded set of ports.
 
         Returns:
           list: A list of every string port number.
         """
+        if self.any_value:
+            return []
```
ahoying: Expanded rules is a dict in one path and a list in the other. I don't like the special casing to check any_value outside of the check for the protected attribute and with a different return path.

I recommend something like:

if not self._expanded_rules:
  self._expanded_rules = {}
  if not self.any_value:
    for rule in self.rules:
...
return self._expanded_rules

Also make sure the docstring matches the type (dict)

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -326,6 +369,9 @@ def __lt__(self, other):
         """
         return (self.action == other.action and
                 (other.applies_to_all or
+                 self.any_value or
```
ahoying: I recommend keeping the order the same, so you always check self.any_value then other.any_value then (self|other).applies_to_all, then the expanded rules checks.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -195,40 +213,49 @@ def is_equivalent(self, other):
...
-            if firewall_rule_denied:
-                raise ValueError(
-                    'Rule cannot have deny (%s) and allow (%s) actions' %
-                    (firewall_rule_denied, firewall_rule_allowed))
-            self.action = 'allow'
-            self.rules = sort_rules(firewall_rule_allowed)
+        if firewall_rule_action not in (FirewallAction.ALLOW,
+                                        FirewallAction.DENY):
+            raise ValueError(
+                'Firewall rule action must be either ALLOW or DENY, got: %s' % (
```
ahoying: Shouldn't it be 'allow' or 'deny' ? Capitals aren't allowed.

google/cloud/security/common/gcp_type/firewall_rule.py
mwwolters: Done

google/cloud/security/common/gcp_type/firewall_rule.py
mwwolters: Done

google/cloud/security/common/gcp_type/firewall_rule.py
mwwolters: Done

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -249,12 +276,27 @@ def applies_to_all(self):
...
 
     @property
+    def any_value(self):
+        """Returns whether this rule matches any value.
+
+        Returns:
+          bool: Whether this rule matches any value.
+        """
+        if self._any_value is None:
+            self._any_value = False
```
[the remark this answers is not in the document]
mwwolters: Done

google/cloud/security/common/gcp_type/firewall_rule.py
mwwolters: Done

google/cloud/security/common/gcp_type/firewall_rule.py
mwwolters: Done

tests/common/gcp_type/firewall_rule_test.py
```diff
@@ -173,11 +222,13 @@ def test_firewall_rule_lt(self, rule_1_dict, rule_2_dict, expected):
                 'firewall_rule_source_ranges': json.dumps(['0.0.0.0/0']),
                 'firewall_rule_direction': 'ingress',
                 'firewall_rule_network': 'n1',
+                'firewall_rule_allowed': json.dumps('*'),
```
ahoying: These should be json.dumps(['*']). The reason the code still works is it treats the string '*' as a list and checks each character against the MATCH_ANY value. It works by accident since MATCH_ANY is a single character, but the code should fail.

You probably should add an assert that firewall_rules is a list in FirewallAction init.

google/cloud/security/common/gcp_type/firewall_rule.py
```diff
@@ -242,6 +242,7 @@ def __init__(self, firewall_rules=None, firewall_rule_action='allow'):
             raise ValueError(
                 'Firewall rule action must be either allow or deny, got: %s' % (
                     firewall_rule_action))
+        assert isinstance(firewall_rules, list)
```
ahoying: Final nit, should this move under if firewall_rules: below?

PR #699, opened 2017-10-11
Added firewall rule for finding policy violations in firewalls.

mwwolters:

### DIFF
--- a/google/cloud/security/scanner/audit/fw_rules_engine.py
+++ b/google/cloud/security/scanner/audit/fw_rules_engine.py
@@ -100,38 +100,109 @@ class Rule(object):
         Args:
           firewall_policies (list): A list of FirewallRule.
 
-        Yields:
+        Returns:
           iterable: A generator of RuleViolations.
         """
         if self.mode == scanner_rules.RuleMode.MATCHES:
-            if (len(firewall_policies) != len(self.match_rules) or
-                    any([is_rule_exists_violation(rule, firewall_policies)
-                         for rule in self.match_rules])):
+            violations = self._yield_match_violations(firewall_policies)
+        elif self.mode == scanner_rules.RuleMode.REQUIRED:
+            violations = self._yield_required_violations(firewall_policies)
+        elif self.mode == scanner_rules.RuleMode.WHITELIST:
+            violations = self._yield_whitelist_violations(firewall_policies)
+        elif self.mode == scanner_rules.RuleMode.BLACKLIST:
+            violations = self._yield_blacklist_violations(firewall_policies)
+        return violations
+
+    def _yield_match_violations(self, firewall_policies):
+        """Finds policies that don't match the required policy.
+
+        Args:
+          firewall_policies (list): A list of FirewallRules to check.
+
+        Yields:
+          iterable: A generator of RuleViolations.
+        """
+        recommended_actions = []
+        for rule in self.match_rules:
+            if is_rule_exists_violation(rule, firewall_policies):
+                recommended_action = self._create_recommendation(
+                    'INSERT_FIREWALL_RULE', rule)
+                recommended_actions.append(recommended_action)
+
+        for policy in firewall_policies:
+            if not any(policy == rule for rule in self.match_rules):
+                recommended_action = self._create_recommendation(
+                    'DELETE_FIREWALL_RULE', policy)
+                recommended_actions.append(recommended_action)
+
+        if recommended_actions:
+            yield self._create_violation(
+                firewall_policies, 'FIREWALL_MATCHES_VIOLATION',
+                recommended_actions=[recommended_action])
+
+    def _yield_required_violations(self, firewall_policies):
+        """Finds missing policies that are required.
+
+        Args:
+          firewall_policies (list): A list of FirewallRules to check.
+
+        Yields:
+          iterable: A generator of RuleViolations.
+        """
+        for rule in self.match_rules:
+            if is_rule_exists_violation(rule, firewall_policies):
+                recommended_action = self._create_recommendation(
+                    'INSERT_FIREWALL_RULE', self)
                 yield self._create_violation(
-                    firewall_policies, 'FIREWALL_MATCHES_VIOLATION')
-        if self.mode == scanner_rules.RuleMode.REQUIRED:
-            if any([is_rule_exists_violation(rule, firewall_policies)
-                    for rule in self.match_rules]):
-                yield self._create_violation(
-                    firewall_policies, 'FIREWALL_REQUIRED_VIOLATION')
+                    firewall_policies, 'FIREWALL_REQUIRED_VIOLATION',
+                    recommended_actions=[recommended_action])
+
+    def _yield_whitelist_violations(self, firewall_policies):
+        """Finds policies that aren't whitelisted.
+
+        Args:
+          firewall_policies (list): A list of FirewallRules to check.
+
+        Yields:
+          iterable: A generator of RuleViolations.
+        """
         for policy in firewall_policies:
             if not any([policy > rule for rule in self.match_rules]):
                 continue
-            if self.mode == scanner_rules.RuleMode.WHITELIST:
-                if is_whitelist_violation(self.verify_rules, policy):
-                    yield self._create_violation(
-                        [policy], 'FIREWALL_WHITELIST_VIOLATION')
-            if self.mode == scanner_rules.RuleMode.BLACKLIST:
-                if is_blacklist_violation(self.verify_rules, policy):
-                    yield self._create_violation(
-                        [policy], 'FIREWALL_BLACKLIST_VIOLATION')
+            if is_whitelist_violation(self.verify_rules, policy):
+                recommended_action = self._create_recommendation(
+                    'DELETE_FIREWALL_RULE', policy)
+                yield self._create_violation(
+                    [policy], 'FIREWALL_WHITELIST_VIOLATION',
+                    recommended_actions=[recommended_action])
 
-    def _create_violation(self, policies, violation_type):
+    def _yield_blacklist_violations(self, firewall_policies):
+        """Finds blacklisted policies.
+
+        Args:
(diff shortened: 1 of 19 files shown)


google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,199 @@
...
+    def find_policy_violations(self, firewall_policies):
+        """Finds policy violations in a list of firewall policies.
+
+        Args:
+          firewall_policies (list): A list of FirewallRule.
+
+        Yields:
+          iterable: A generator of RuleViolations.
+        """
+        if self.mode == scanner_rules.RuleMode.MATCHES:
```
ahoying: I would move each mode into a separate private method.

google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,199 @@
...
+            if (len(firewall_policies) != len(self.match_rules) or
+                    any([is_rule_exists_violation(rule, firewall_policies)
+                         for rule in self.match_rules])):
+                yield self._create_violation(
+                    firewall_policies, 'FIREWALL_MATCHES_VIOLATION')
```
ahoying: It would be nice if the violation pointed out which rules were missing or were extra, etc.

In fact, for all of the violations it would be nice to have something like a 'recommended actions' section with a list of actions like DELETE FIREWALL RULE rule_name: <rule_json>, INSERT FIREWALL RULE rule_name: <rule_json>, and/or UPDATE FIREWALL RULE rule_name: <rule_json>.

google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,199 @@
...
+        if self.mode == scanner_rules.RuleMode.REQUIRED:
+            if any([is_rule_exists_violation(rule, firewall_policies)
+                    for rule in self.match_rules]):
+                yield self._create_violation(
+                    firewall_policies, 'FIREWALL_REQUIRED_VIOLATION')
+        for policy in firewall_policies:
```
ahoying: This block will get run even for REQUIRED and MATCHES modes, it just won't do anything. I don't think that was the intention?

google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,199 @@
...
+    """Checks if the rule is the same as one of the policies.
+
+    Args:
+      rule (FirweallRule): A FirewallRule.
+      policies (list): A list of FirewallRule that must have the rule.
+
+    Returns:
+      bool: If the required rule is in the policies.
+    """
+    return not any([policy == rule for policy in policies])
```
ahoying: Is == correct here, or would is_equivalent be better? Or maybe matches should use ==, but required should use is_equivalent?

google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+    # Rule violation.
+    # resource_type: string
+    # resource_id: string
+    # rule_name: string
+    # rule_index: int
+    # violation_type: FIREWALL_VIOLATION
+    # policy_name: string
+    RuleViolation = namedtuple('RuleViolation',
```
ahoying: Is there a reason to nest this in the class? I think it could be a module level class.

tests/scanner/audit/fw_rules_engine_test.py
```diff
@@ -0,0 +1,932 @@
+"""Tests for google3.experimental.users.mwwolters.forseti.forseti-security.tests.scanner.audit.fw_rules_engine."""
```
ahoying: Fix comment (information leakage)

ahoying: Also add copyright header.

google/cloud/security/scanner/audit/fw_rules_engine.py
ahoying: Something like:
if self.mode == MATCHES:
  violations = self._yield_match_violations(firewall_policies)
elif self.mode == REQUIRED:
  violations = self._yield_required_violations(firewall_policies)
...
\# Return a generator object that yields policy violations
return violations

mwwolters: Done.

google/cloud/security/scanner/audit/fw_rules_engine.py
mwwolters: Done

google/cloud/security/scanner/audit/fw_rules_engine.py
mwwolters: Done

google/cloud/security/scanner/audit/fw_rules_engine.py
mwwolters: Done

google/cloud/security/scanner/audit/fw_rules_engine.py
mwwolters: I'm not committed either way, but I was thinking it was similar to a "matches" policy in that we want a specific policy to be applied. I think if we wanted to audit for certain things being allowed but not a specific policy then we would want a separate violation type for that.

tests/scanner/audit/fw_rules_engine_test.py
mwwolters: Whoops

google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,292 @@
...
+        recommended_actions = []
+        for rule in self.match_rules:
+            if is_rule_exists_violation(rule, firewall_policies,
+                                        self._exact_match):
+                recommended_action = self._create_recommendation(
+                    'INSERT_FIREWALL_RULE', rule)
+                recommended_actions.append(recommended_action)
+
+        for policy in firewall_policies:
+            if not any(policy == rule for rule in self.match_rules):
```
ahoying: Can this use: is_rule_exists_violation(policy, self.match_rules, self._exact_match) ? Then you keep the same logic for both the not in current rules and not in defined policy rules.

google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,292 @@
...
+                recommended_action = self._create_recommendation(
+                    'DELETE_FIREWALL_RULE', policy)
+                recommended_actions.append(recommended_action)
+
+        if recommended_actions:
+            yield self._create_violation(
+                firewall_policies, 'FIREWALL_MATCHES_VIOLATION',
+                recommended_actions=[recommended_action])
```
ahoying: Shouldn't this be recommended_actions=recommended_actions? Is there a test that has multiple recommended actions? The test should have failed?

google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,292 @@
...
+                                        self._exact_match):
+                recommended_action = self._create_recommendation(
+                    'INSERT_FIREWALL_RULE', rule)
+                recommended_actions.append(recommended_action)
+
+        for policy in firewall_policies:
+            if not any(policy == rule for rule in self.match_rules):
+                recommended_action = self._create_recommendation(
+                    'DELETE_FIREWALL_RULE', policy)
+                recommended_actions.append(recommended_action)
```
ahoying: One thing you could do is make recommended_actions a dictionary.

recommended_actions = {
  'INSERT_FIREWALL_RULE': {rule.name: rule, rule.name: rule},
  'DELETE_FIREWALL_RULE': {rule.name: rule, rule.name: rule},
  'UPDATE_FIREWALL_RULE': {rule.name: rule, rule.name: rule},
}

Then you can check if rule.get('name') is in both insert and delete, remove it from both dicts and move it to update, or before adding a rule to delete, check if it is already in insert since you create all the inserts first. If it is remove it from insert and move it to update. A rule with the same name needs to use update, otherwise you would get an exception from the API when trying to insert it again with the same name.

Down the road when rule.to_json() is defined, you have all the data you'll need to update the recommended action strings.

Then self._create_recommendation() should take a dict like above and parse it out into the list of strings, in order of INSERT, DELETE, UPDATE and sorted by rule.name.

google/cloud/security/scanner/audit/fw_rules_engine.py
```diff
@@ -0,0 +1,292 @@
...
+          firewall_policies (list): A list of FirewallRules to check.
+
+        Yields:
+          iterable: A generator of RuleViolations.
+        """
+        for rule in self.match_rules:
+            if is_rule_exists_violation(rule, firewall_policies,
+                                        self._exact_match):
+                recommended_action = self._create_recommendation(
+                    'INSERT_FIREWALL_RULE', self)
```
ahoying: Shouldn't this be rule, not self? Should a test have failed on this?

google/cloud/security/scanner/audit/fw_rules_engine.py
mwwolters: Done.

google/cloud/security/scanner/audit/fw_rules_engine.py
mwwolters: Mostly done.
to_json: for next PR
self._create_recommendation: was removed. I think it's better to pass this as a dict and other code can handle parsing it at a different stage.

google/cloud/security/scanner/audit/fw_rules_engine.py
mwwolters: Done.

google/cloud/security/scanner/audit/fw_rules_engine.py
mwwolters: Done.

PR #717, opened 2017-10-16
Fix email error

### LINKED ISSUE #602: If no sender/recipient or API key are provided to email util, don't raise the Error
Right now we raise an Error which customers get concerned about because it's an error.

carise:
Don't raise an EmailError if no sender/recipient on the email, just log a warning and move on.

resolves #602

### DIFF
--- a/google/cloud/security/common/util/email_util.py
+++ b/google/cloud/security/common/util/email_util.py
@@ -18,7 +18,6 @@ import base64
 import os
 import urllib2
 
-import gflags as flags
 import jinja2
 
 from retrying import retry
--- a/google/cloud/security/common/util/email_util.py
+++ b/google/cloud/security/common/util/email_util.py
@@ -103,7 +103,7 @@ class EmailUtil(object):
         if not email_sender or not email_recipient:
             LOGGER.warn('Unable to send email: sender=%s, recipient=%s',
                         email_sender, email_recipient)
-            return
+            raise util_errors.EmailSendError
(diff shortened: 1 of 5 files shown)


google/cloud/security/common/util/email_util.py
```diff
@@ -116,7 +103,7 @@ def send(self, email_sender=None, email_recipient=None,
         if not email_sender or not email_recipient:
             LOGGER.warn('Unable to send email: sender=%s, recipient=%s',
                         email_sender, email_recipient)
-            raise util_errors.EmailSendError
+            return
```
blueandgold: This is handled upstream to capture a bit more detail of what is not able to be sent.  If you want to keep the logging here, can you please add the detail here, perhaps logging the email subject?

https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/security/notifier/pipelines/email_inventory_snapshot_summary_pipeline.py#L92

If the EmailSendError will not be used / raised, can you please clean remove the exception and from upstream?

carise: Hm, interesting -- I think I'll tweak the code so that we keep the EmailSendError and handle it upstream for the Scanner summary. That's where the stacktrace is showing up.

carise: Tweaked the scanner and violation email notification pipelines so that they catch the EmailSendError.

PR #728, opened 2017-10-18
Add more setup options, gsuite superadmin, pin default version to local

### LINKED ISSUE #708: DOMAIN_SUPER_ADMIN_EMAIL is not populated by setup wizard
Deploying with the setup wizard, the DOMAIN_SUPER_ADMIN_EMAIL field is not being populated in the forseti_conf.yaml file.  This is causing an extra step for the user to edit the forseti_conf.yaml file and upload to GCS bucket if they need gsuite data.

### LINKED ISSUE #733: Tie default deployments via installer to the local Forseti version
There's an issue with drift between the local Forseti code and what gets deployed to GCE, mainly in the Deployment Manager startup script.

For example, if the user downloads 1.1.3, the installer (as of 1.1.6 and prior) will deploy the "master" branch of Forseti. This is fine if HEAD on master is pointing to 1.1.3, but if the GCE instance gets restarted or there was a lag between when the user downloaded Forseti and when the user ran the deployment, there could be inconsistencies in the startup script. The 1.1.3 startup script will be deployed to the GCE instance, but later releases could require dependencies that are not found in the 1.1.3 startup script.

To fix this, the installer should, by default, tie the deployment template "release-version" to the current `__version__` found in `google/cloud/security/__init__.py`. That way, whatever version the user is running the installer from will also run that same version on the GCE instance.

Also, print the Forseti version that will be/has been installed -- once in the beginning of installation, and then at the end in the summary.

carise:
WIP, needs additional discussion/documentation

Changes:

* ~Add option for `--unattended-install`~
* G Suite domain super admin email
  * Prompt for the super admin email
  * Add a flag for the super admin email
* Add flag for Sendgrid API key
* Add flag for notification email recipient
* Delete the gsuite private key once after attempting to copy it to the GCE instance
* Pin deployments done with setup wizard to the local version, by default (i.e. what's found in `g/c/s/__init__.py`)

resolves #708 
resolves #733

### DIFF
--- a/scripts/gcp_setup/environment/gcloud_env.py
+++ b/scripts/gcp_setup/environment/gcloud_env.py
@@ -114,7 +114,6 @@ class ForsetiGcpSetup(object):
         self.force_no_cloudshell = kwargs.get('no_cloudshell')
         self.skip_iam_check = kwargs.get('no_iam_check')
         self.branch = kwargs.get('branch')
-        self.unattended_install = kwargs.get('unattended_install')
 
         self.is_devshell = False
         self.authed_user = None
@@ -714,9 +713,6 @@ class ForsetiGcpSetup(object):
 
     def _get_user_input(self):
         """Ask user for specific setup values."""
-        if self.unattended_install:
-            return
-
         if not self.sendgrid_api_key:
             # Ask for SendGrid API Key
             print('Forseti can send email notifications through SendGrid '
@@ -738,9 +734,9 @@ class ForsetiGcpSetup(object):
 
         if not self.gsuite_superadmin_email:
             # Ask for G Suite super admin email
-            print('\nTo read G Suite Groups data (e.g. if you are want to use '
-                  'IAM Explain), please provide a G Suite super admin email '
-                  'address. '
+            print('\nTo read G Suite Groups data, for example, if you want to '
+                  'use IAM Explain, please provide a G Suite super admin '
+                  'email address. '
                   'This step is optional and can be configured later.')
             self.gsuite_superadmin_email = raw_input(
                 'What is your organization\'s G Suite super admin email? '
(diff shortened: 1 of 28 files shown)


carise:
what's up with the cla?

scripts/gcp_setup/environment/gcloud_env.py
```diff
@@ -647,39 +689,20 @@ def generate_forseti_conf(self):
...
 
-        # Ask for SendGrid API Key
-        print('Forseti can send email notifications through SendGrid '
-              'via an API key. '
-              'This step is optional and can be configured later.')
-        sendgrid_api_key = raw_input(
-            'What is your SendGrid API key? (press [enter] to skip) ').strip()
-        if sendgrid_api_key:
-            self.sendgrid_api_key = sendgrid_api_key
+        self._get_user_input()
```
blueandgold: I would like to see the ```if self.unattended_install:``` condition to be outside here, instead of being inside ```_get_user_input()```.  This way, it's easier to understand the workflow that if this is not an unattended_install, then prompt the user for input.

```
if not self.unattended_install:
    _get_user_input()
```

scripts/gcp_setup/setup_forseti.py
```diff
@@ -33,12 +33,24 @@ def run():
                         help='Bypass IAM check for user running script')
     parser.add_argument('--branch',
                         help='Which Forseti branch to deploy')
+    parser.add_argument('--unattended-install',
```
blueandgold: Can you please explain why we are adding this unattended install option now?  Did anyone specifically ask for it?   The reason I am asking is because now we will have to maintain & test 2 different flows here.  So, I really want to make sure there is a real benefit to add this.

scripts/gcp_setup/environment/gcloud_env.py
```diff
@@ -137,15 +139,19 @@ def __init__(self, **kwargs):
         self.deploy_tpl_path = None
         self.forseti_conf_path = None
 
+        # forseti_conf.yaml.in properties
         self.skip_email = False
-        self.sendgrid_api_key = '""'
-        self.notification_sender_email = '""'
-        self.notification_recipient_email = '""'
+        self.sendgrid_api_key = kwargs.get('sendgrid_api_key')
```
blueandgold: For unattended install mode, are there any checks or validations that certain flags have to be set?   I think that is really important to have.

carise:
I'll take out unattended install, as no one asked for it.

PR #755, opened 2017-11-01
Fix resource ancestry lookup

carise:
The previous code assumed that the resource to lookup ancestry for already "knew" whether it had a parent. This fix checks if the resource has a parent supplied and if not, performs a lookup (just in case we didn't preload the data).

### DIFF
--- a/google/cloud/security/common/data_access/org_resource_rel_dao.py
+++ b/google/cloud/security/common/data_access/org_resource_rel_dao.py
@@ -64,24 +64,42 @@ class OrgResourceRelDao(object):
         while curr_resource is not None:
             parent_resource = None
 
+            # If we don't have parent information for the current
+            # resource, try to look it up.
+            if not curr_resource.parent:
+                curr_resource = self._load_resource(
+                    curr_resource, snapshot_timestamp)
+
             if (curr_resource.parent and
                     curr_resource.parent.type and
                     curr_resource.parent.id):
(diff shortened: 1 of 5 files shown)


google/cloud/security/common/data_access/org_resource_rel_dao.py
```diff
@@ -64,24 +64,42 @@ def find_ancestors(self, org_resource, snapshot_timestamp=None):
...
+        """Load the resource from the database.
+
+        Args:
+            unloaded_resource (Resource): The Resource to load.
+            snapshot_timestamp (str): The timestamp to use for data lookup.
+
+        Returns:
+            Resource: The resource.
+        """
+        loaded_resource = unloaded_resource
```
blueandgold: nit: the code can be a bit more clear, as well as less nesting with the below.

```
if not unloaded_resource :
    return unloaded_resource

resource_lookup = self._resource_db_lookup.get(
    unloaded_resource.type, {})
# Invoke the dao.get_*() method, to get the resource
if resource_lookup.get('dao'):
    loaded_resource = getattr(
        resource_lookup.get('dao'),
        resource_lookup.get('get'))(
            unloaded_resource.id, snapshot_timestamp)
return loaded_resource
```

carise: Done

carise:
Thanks for the review!

PR #756, opened 2017-11-01
Tweak names

carise:
Renaming the fw_rules* to firewall_rules* for clarity.

Also fixed a typo in the forwarding rules tests.


google/cloud/security/scanner/scanners/firewall_rules_scanner.py
```diff
@@ -25,13 +25,13 @@
...
 from google.cloud.security.scanner.scanners import base_scanner
 
 LOGGER = log_util.get_logger(__name__)
 
 
-class FwPolicyScanner(base_scanner.BaseScanner):
+class FirewallPolicyScanner(base_scanner.BaseScanner):
     """Scanner for firewall data."""
 
     SCANNER_OUTPUT_CSV_FMT = 'scanner_output_fw.{}.csv'
```
blueandgold: What about this?  If we are doing this renaming, we should get everything.

carise: Good point, I'll fix that.

carise:
Thanks for the review!

PR #758, opened 2017-11-01
Notifier violations

carise:
Tweak the setup wizard default conf to upload violations to GCS for the violations that are not in (IAM, firewall, IAP).

Also call flush() on the temp file before uploading to GCS. For very small files (don't know the exact size, but anything under 1KB seemed to be fair game), we were uploading 0B files because the data hadn't been flushed. [StackOverflow link](https://stackoverflow.com/questions/9422478/why-does-tempfile-namedtemporaryfile-truncate-my-data)


carise:
Thanks for the review!

PR #762, opened 2017-11-02
Add firewall rule violations to database

### LINKED ISSUE #760: Firewall violations are not written to the violations table.
This is for 1.1.7 release.  The email with the csv attachment of firewall violations does get sent, but the violations are not saved to the violations table.

carise:
resolves #760


carise:
Thanks for the review!

PR #804, opened 2017-11-14
Copy the scanner service (by cherrypick).

blueandgold:
The scanner service was not copied into this branch.  So, cherrypick the original commit back in.
https://github.com/GoogleCloudPlatform/forseti-security/commit/05c29d4062e64ae2f586d37191964926f561e00b


blueandgold:
@felixbb there is a whole bunch of ```_Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.UNIMPLEMENTED, Method not found!)>``` errors in the tests.  Would you know why that is?

PR #807, opened 2017-11-14
lowercase string comparison

angelomellos:
Usernames are case insensitive, but the Forseti permissions check is case sensitive. So certain accounts will get an error when trying to run Forseti, saying that they don't have the Org Admin role, even though they do. That was the cause of that error on Friday.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/forseti-security/807)


carise:
Hey @angelomellos thanks for your PR! When you get the chance, could you also sign the CLA?

angelomellos-nccgroup:
Ok the CLA is signed now. I added the gmail account angelomellos.nccgroup@gmail.com, which is now in the Google Group that signed the CLA, to my Github account.

PR #838, opened 2017-11-22
Integrate the Firewall Scanner with Inventory 2.0

blueandgold:
This makes the firewall scanner to work with the new data model provided by Inventory 2.0.  Test is pending until the overall work here is LGTM.

One main change is that I added a custom function to do ancestor lookup by parsing the hierarchical name provided in the data model, since the the data access is not available to look up individual resource data model.  Hope this is an acceptable compromise until auditor comes online.

Example of a hierarchical name:
```organization/88888/project/myproject/firewall/99999/```

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/forseti-security/838)

### CODE REVIEW

File: google/cloud/forseti/common/gcp_type/firewall_rule.py [hunk 1]
```diff
@@ -66,6 +66,7 @@ def __init__(self, validate=False, **kwargs):
         self.resource_id = kwargs.get('id')
         self.create_time = kwargs.get('firewall_rule_create_time')
         self.name = kwargs.get('firewall_rule_name')
+        self.hierarchical_name = kwargs.get('firewall_rule_hierarchical_name')
```
mwwolters: This should probably default to a string, as get_resource_ancestors is causing the tests to fail due to the rsplit on a None hierarchical_name.

blueandgold: Done

google/cloud/forseti/scanner/scanners/firewall_rules_scanner.py
```diff
@@ -176,10 +181,22 @@ def _retrieve(self):
...
+            for i in data_access.scanner_iter(session, "firewall"):
+                firewall_data_for_scanner = ast.literal_eval(i.data)
+                firewall_data_for_scanner['project_id'] = i.parent.display_name
+                firewall_data_for_scanner['hierarchical_name'] = i.full_name
+
+                firewall_policies.append(
+                    firewall_rule.FirewallRule.from_dict(
+                        firewall_data_for_scanner,
+                        i.parent.display_name,
+                        True))
```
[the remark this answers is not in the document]
blueandgold: Done

blueandgold:
/pr test is failing because ```tests.iam.utils.db``` has moved to ```tests/services/utils```.  I'll fix this when I sync before merging.

blueandgold:
Review status: 0 of 26 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[google/cloud/forseti/common/data_access/org_resource_rel_dao.py, line 129 at r4](https://reviewable.io:443/reviews/googlecloudplatform/forseti-security/838#-L-YV_tv0i83k2W7Fl_K-r4-129:-L-bWFiW6nqAPOMrnFDX:bb05dve) ([raw file](https://github.com/googlecloudplatform/forseti-security/blob/d637ebb797c692f2e251e879902d2e5179bab86a/google/cloud/forseti/common/data_access/org_resource_rel_dao.py#L129)):*
<details><summary><i>Previously, mwwolters wrote…</i></summary><blockquote>

This method feels like it should be more generic than for just firewall rules, perhaps pass in the hierarchical name instead?
</blockquote></details>

Excellent prescriptive advice: Done!

---

*[google/cloud/forseti/common/data_access/org_resource_rel_dao.py, line 139 at r4](https://reviewable.io:443/reviews/googlecloudplatform/forseti-security/838#-L-YV_tv0i83k2W7Fl_K-r4-139:-L-bWMsYmRFlaHBZokV9:biukace) ([raw file](https://github.com/googlecloudplatform/forseti-security/blob/d637ebb797c692f2e251e879902d2e5179bab86a/google/cloud/forseti/common/data_access/org_resource_rel_dao.py#L139)):*
<details><summary><i>Previously, mwwolters wrote…</i></summary><blockquote>

Why not just:
hierarchical_name_parts = hierarchical_name.split('/')[:-1]
since you don't use hierarchical_name after that.
</blockquote></details>

I broke this into a separate line, so that the operation to strip the trailing '/' is clear on its own.  I could combine the 2 steps, but it will be harder to decipher.  So, I will keep the operations separate.

---

*[google/cloud/forseti/services/scanner/service.py, line 68 at r4](https://reviewable.io:443/reviews/googlecloudplatform/forseti-security/838#-Kzzg1VJ1GT_evQEC-bM-r4-68:-L-bWbCkaC0eXRk4nmTX:b-896fix) ([raw file](https://github.com/googlecloudplatform/forseti-security/blob/d637ebb797c692f2e251e879902d2e5179bab86a/google/cloud/forseti/services/scanner/service.py#L68)):*
<details><summary><i>Previously, mwwolters wrote…</i></summary><blockquote>

Having SERVICE_CONFIG as None and requiring it to be set externally feels weird to me. Have you considered making a scanner object where you have to set the config?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/googlecloudplatform/forseti-security/838)*

google/cloud/forseti/scanner/scanners/base_scanner.py
```diff
@@ -64,15 +66,10 @@ def _output_results_to_db(self, violations):
...
-        try:
-            vdao = violation_dao.ViolationDao(self.global_configs)
-            (inserted_row_count, violation_errors) = vdao.insert_violations(
-                violations,
-                snapshot_timestamp=self.snapshot_timestamp)
-        except db_errors.MySQLError as err:
-            LOGGER.error('Error importing violations to database: %s\n%s',
-                         err, violations)
 
+        violation_access = self.service_config[0].violation_access(
```
ahoying: This code is untested, I recommend ensuring there is a test using the sqlite db engine.

blueandgold: ```scanner_dao_test.test_save_violations()``` should provide the test coverage here, since the code here is just calling the violation access.  I have, however, updated the TODO to add unit test when violation error is handled, which would be interested to test.

google/cloud/forseti/scanner/scanners/firewall_rules_scanner.py
```diff
@@ -176,10 +181,22 @@ def _retrieve(self):
...
-                             .FirewallRuleDao(self.global_configs)
-                             .get_firewall_rules(self.snapshot_timestamp))
 
+        model_manager = self.service_config[0].model_manager
+        scoped_session, data_access = model_manager.get(self.model_name)
+        with scoped_session as session:
+            firewall_policies = []
+
+            for i in data_access.scanner_iter(session, "firewall"):
+                firewall_data_for_scanner = ast.literal_eval(i.data)
```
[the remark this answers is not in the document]
ahoying: That won't work, it will die for data other than the firewall (which happens to not have any Boolean values so doesn't crash). It is JSON data, it just isn't a json type. You have to load it with json.loads.

tests/services/scanner/scanner_dao_test.py
```diff
@@ -67,33 +68,18 @@ def test_save_violations(self):
...
+                'violation_data']
+        for fake, saved in izip(FAKE_VIOLATIONS, saved_violations):
+            for key in keys:
+                if key != 'violation_data':
+                    self.assertEquals(
+                        fake.get(key), getattr(saved, key),
+                        'key %s differs' % key)
+                else:
+                    self.assertEquals(
+                        fake.get(key), ast.literal_eval(getattr(saved, 'data')),
```
ahoying: This needs to use json.loads as well. The only reason ast.literal_eval works is because the violation data doesn't contain boolean values. You can test this by setting violation_data to something like this:

```
          {'policy_names': ['fw-tag-match_222'],
           'action_required': true,
           'recommended_actions':
               {'DELETE_FIREWALL_RULES': ['fw-tag-match_222']}},
```

blueandgold:
Review status: 0 of 30 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[tests/services/scanner/scanner_dao_test.py, line 81 at r10](https://reviewable.io:443/reviews/googlecloudplatform/forseti-security/838#-L-gY0Ur4796DkVy7YbH-r10-81:-L-gZJ4sErLgQLxB99-Y:b-31crth) ([raw file](https://github.com/googlecloudplatform/forseti-security/blob/90b020082a36ae98406ef0ba3aa5018e206b971d/tests/services/scanner/scanner_dao_test.py#L81)):*
<details><summary><i>Previously, ahoying (Andrew Hoying) wrote…</i></summary><blockquote>

This needs to use json.loads as well. The only reason ast.literal_eval works is because the violation data doesn't contain boolean values. You can test this by setting violation_data to something like this:

```
          {'policy_names': ['fw-tag-match_222'],
           'action_required': true,
           'recommended_actions':
               {'DELETE_FIREWALL_RULES': ['fw-tag-match_222']}},
```

</blockquote></details>

Ah, done.  I think that I didn't catch what you're saying earlier (long day, and late night).  But you're right, done!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/googlecloudplatform/forseti-security/838)*

PR #836, opened 2017-11-21
2.0 merge gcp api

ahoying:
Thanks for opening a Pull Request!

Here's a handy checklist to ensure your PR goes smoothly.

- [X] I signed Google's [Contributor License Agreement](https://opensource.google.com/docs/cla/)
- [X] My code conforms to Google's [python style](https://google.github.io/styleguide/pyguide.html).
- [ ] My PR at a minimum doesn't decrease unit-test coverage (if applicable).
- [ ] My PR has been functionally tested.
- [ ] All of the [unit-tests](http://forsetisecurity.org/docs/development/#executing-tests) still pass.
- [ ] Running `pylint --rcfile=pylintrc` passes.

These guidelines and more can be found in our [contributing guidelines](https://github.com/GoogleCloudPlatform/forseti-security/blob/master/.github/CONTRIBUTING.md).

PR #839, opened 2017-11-22
Merge untracked progress on branch folder_project_inventory to 2.0-dev

jiyunyao:
Progress included:
Fixes for deployments
Fix docker tests and precommit hook
Fix proxy binary check
Update gcloud commands in iam_auto_deployment.sh
Refactor the fix on inventory when APIs are disabled for some projects
Fix the inventory when no group in GSuite is created
Test fix when port number is not released in time in case of SQLite

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/forseti-security/839)


google/cloud/forseti/services/inventory/crawler.py
```diff
@@ -91,24 +91,20 @@ def get_client(self):
...
-        if not resource._warning:
-            resource._warning = warning_message
-        else:
-            resource._warning += warning_message
         self.config.storage.warning(warning_message)
         self.config.progresser.on_warning(error)
 
     def update(self, resource):
-        """Update the error message in the row of the resource
+        """Update the row of an existed resource
```
blueandgold: nit: do you mean ```existing``` rather than ```existed```?

ahoying: Can those files be left out of this PR and merged in post #836? In general I don't think there will be conflicts with my changes, but I'd like to understand the changes in context, and not in with the other unrelated changes in this PR.

jiyunyao: Thanks, I just checked dictionary. I mean "existing", let me fix this

jiyunyao: Sure. We can do that. I don't think there will be any conflicts though.

scripts/iam_auto_deployment.sh
```diff
@@ -13,20 +13,58 @@
...
+BRANCHNAME=""
+SQLINSTANCE=""
+PROJECT_ID=""
+DEPLOYMENTNAME=""
+
+# WARNING: THESE VALUES ARE NOT CONFIGURABLE!
+# These defaults match values contained in the example deployment YAMLs.
+# These should not be considered to be configurable but they are placed
+# here to avoid repetition.
+FORSETI_DB_NAME="forseti_security"
```
ahoying: This variable is no longer used. Now that we have combined the databases, I recommend just leaving in FORSETI_DB_NAME and replacing all uses of EXPLAIN_DB_NAME with FORSETI_DB_NAME.

scripts/iam_auto_deployment.sh
```diff
@@ -331,25 +368,18 @@ then
 	$CmdPrefix add-iam-policy-binding $ROOTID \
 	 --member=serviceAccount:$SCRAPINGSA \
 	 --role=roles/storage.admin
-	
+
+	gcloud projects add-iam-policy-binding $PROJECT_ID \
```
ahoying: You also need to add --role=roles/logging.logWriter to the project IAM bindings.

deployment-templates/compute-engine/explain-instance.py
```diff
@@ -46,6 +48,18 @@ def GenerateConfig(context):
 
     resources = []
 
+    EXPORT_INITIALIZE_VARS = (
+        'export SQL_PORT={0}\n'
+        'export SQL_INSTANCE_CONN_STRING="{1}"\n'
+        'export EXPLAIN_DB_NAME="{2}"\n'
```
ahoying: I recommend replacing the use of EXPLAIN_DB with FORSETI_DB.

google/cloud/forseti/services/inventory/base/resources.py
```diff
@@ -107,6 +106,16 @@ def parent(self):
     def key(self):
         raise NotImplementedError('Class: {}'.format(self.__class__.__name__))
 
+    def add_warning(self, warning):
+        warning_message = '{}\n'.format(warning)
```
ahoying: What about making self._warning a list that you append to, and then in get_warning() return '\n'.join(self._warning) ? Then you don't need to build a new string for each new warning.

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -571,25 +606,24 @@ def update(self, resource):
...
+
+        try:
+            new_rows = Inventory.from_resource(self.index, resource)
+            old_rows = self._get_resource_rows(resource.key())
+
+            new_dict = {row.type_class : row for row in new_rows}
+            old_dict = {row.type_class : row for row in old_rows}
+
+            for type_class in InventoryTypeClass.SUPPORTED_TYPECLASS:
+                if type_class in new_dict:
```
ahoying: Is there ever a case where there are rows in old_dict that are not in new_dict that should be deleted?

scripts/iam_auto_deployment.sh
```diff
@@ -376,13 +406,10 @@ then
...
 	BRANCHNAME=$DownloadBranch
 fi
-sed -i -e 's/BRANCHNAME/'$BRANCHNAME'/g' \
-$repodir/deployment-templates/deploy-explain.yaml
 
 # sql instance name
 echo -e "${TYELLOW}Choosing SQL Instance name${TNC}"
-timestamp=$(date --utc +%Ft%Tz | sed -e 's/:/-/g')
-SQLINSTANCE="iam-explain-no-external-"$timestamp
+SQLINSTANCE="iam-explain-no-external-"$LABEL_SAFE_TIMESTAMP
```
ahoying: replace iam-explain reference(s) with forseti-security

deployment-templates/compute-engine/explain-instance.py
jiyunyao: I fixed this one. Can we open issues if there are other names needs to be changed? I don't want to accumulate too much fixes and improvements in this PR.

ahoying: I'm fine with putting the variable rename changes in a future PR.

google/cloud/forseti/services/inventory/base/resources.py
jiyunyao: Fixed this one.

google/cloud/forseti/services/inventory/storage.py
jiyunyao: We don't have any case that there are rows in old_dict should be deleted. This is why we want delete information here.
There are 3 rows returns at most:
resource, iam_policy, gcs_policy
Right now the update is only used when adding error logs. 
Only resource has been updated when adding error logs.

scripts/iam_auto_deployment.sh
jiyunyao: Fixed

scripts/iam_auto_deployment.sh
jiyunyao: fixed

scripts/iam_auto_deployment.sh
jiyunyao: fixed

PR #840, opened 2017-11-24
Setup wizard can deploy Forseti to a shared VPC

bgleonard:
This PR is designed to allow users who wish to run the Forseti Compute VM in a shared VPC to use the setup wizard. The user must specify the host project id, shared vpc name, and subnetwork name as arguments.

The generated compute instance template section networkInterfaces.network will now contain a host-project and vpc-name. The host-project can either be the host project id supplied by the user or the current project id as discovered by the script. The vpc-name can either be the vpc name supplied by the user or the name string 'default'.

The generated compute instance template section networkInterfaces.subnetwork is new. It will contain a host-project defined the same as above, region defined as it was previously, and subnetwork supplied by the user or the name string 'default'.


bgleonard:
Rebased, amended commit with correct author for CLA.

PR #855, opened 2017-12-01
Rollback when session cannot be commit

jiyunyao:

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/forseti-security/855)


google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -192,6 +192,11 @@ def run(self):
             traceback.print_exc(file=buf)
             buf.seek(0)
             message = buf.read()
+            try:
+                self.session.commit()
+            except Exception as e:
```
ahoying: There should not be any need for such a broad except catch here, please catch the exception(s) that session.commit() could raise. That will also fix the lint error.

Issue #848, opened 2017-11-29
Slack notification cause traceback errors


brandon-height:
-  Forseti Security version: "1.1.7"
 -  Notifier is the problem module
 - Setup Wizard (Forseti is in an isolated project)

In notifier section of forseti_conf.yaml:
```
            - name: slack_webhook_pipeline
              configuration:
                webhook_url: '<excluded>'
```
If the above is included in any resource under the notifier section I receive the following error:

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-1.1.7-py2.7.egg/google/cloud/security/notifier/notifier.py", line 225, in <module>
    app.run()
  File "/usr/local/lib/python2.7/dist-packages/google/apputils/app.py", line 238, in run
    return _actual_start()
  File "/usr/local/lib/python2.7/dist-packages/google/apputils/app.py", line 267, in _actual_start
    really_start()
  File "/usr/local/lib/python2.7/dist-packages/google/apputils/app.py", line 220, in really_start
    sys.exit(main(argv))
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-1.1.7-py2.7.egg/google/cloud/security/notifier/notifier.py", line 221, in main
    pipeline.run()
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-1.1.7-py2.7.egg/google/cloud/security/notifier/pipelines/slack_webhook_pipeline.py", line 96, in run
    webhook_payload = self._compose(violation=violation)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-1.1.7-py2.7.egg/google/cloud/security/notifier/pipelines/slack_webhook_pipeline.py", line 70, in _compose
    'details': json.loads(violation.get('violation_data'))
  File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
TypeError: expected string or buffer
```
I dropped a print statement into google/cloud/security/notifier/pipelines/slack_webhook_pipeline.py
in the _compose() method to print out the violation_data it attempts to json.loads()
```print(violation.get('violation_data'))```

After installing my change and manually running forseti I can see the offending violation_data:
[...]

PR #866, opened 2017-12-05
Fix json breakage in slack notifiiction.

blueandgold:
The issue is that [json.loads() is called upstream] (https://github.com/GoogleCloudPlatform/forseti-security/pull/762/files#diff-35e1e2967889bcc1d9049e770706aee7R164), which then causes a problem when json.loads() is called here again.

Best way to make sure this doesn't regress in the future is to test this functionally, when we have an in-memory database.  Will open an issue for this.


blueandgold:
Fixes issue #848

PR #879, opened 2017-12-07
Scanner tweaks

carise:
* Change sys.exit(1) to return some int value, just so the client doesn't hang if there's an error
* Don't use the snapshot_timestamp

### DIFF
--- a/google/cloud/forseti/scanner/scanner.py
+++ b/google/cloud/forseti/scanner/scanner.py
@@ -92,14 +92,14 @@ def run(forseti_config, model_name=None, service_config=None):
 
     if forseti_config is None:
         LOGGER.error('Path to Forseti Security config needs to be specified.')
-        sys.exit()
+        return 1
 
     try:
         configs = file_loader.read_and_parse_file(forseti_config)
     except IOError:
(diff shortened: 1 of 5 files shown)


google/cloud/forseti/scanner/scanner.py
```diff
@@ -92,33 +91,33 @@ def run(forseti_config, model_name=None, service_config=None):
...
+    #    sys.exit()
 
     violation_access = scanner_dao.define_violation(
         model_name, service_config.engine)
     service_config.violation_access = violation_access
 
     runnable_scanners = scanner_builder.ScannerBuilder(
         global_configs, scanner_configs, service_config, model_name,
-        snapshot_timestamp).build()
+        model_name).build()
```
blueandgold: This model_name should not be repeated.

carise: Yes, I tested the functionality -- the reason for this PR is that I was having trouble getting scanner to work on my side.

The snapshot timestamp is not used as far as I can tell, for the firewall scanner. However, because we've moved over one scanner, the other scanners still use the snapshot timestamp (it seems) and I didn't want to totally drop it.

PR #880, opened 2017-12-07
Use a temporary .forseti that gets cleaned up after unit tests for cli

carise:
Instead of overwriting the default .forseti in $HOME, create a temp directory and temp file for the Forseti configuration and clean it up after running the CLI tests.

### DIFF
--- a/tests/services/cli_test.py
+++ b/tests/services/cli_test.py
@@ -17,8 +17,10 @@ from argparse import ArgumentParser
 from copy import copy
 import json
 import os
+import shutil
 import shlex
 import StringIO
+import tempfile
 import unittest
 import mock
 from tests.unittest_utils import ForsetiTestCase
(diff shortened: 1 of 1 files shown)


tests/services/cli_test.py
```diff
@@ -191,36 +197,39 @@ def tearDown(self):
...
-                            attribute,
-                            value))
-
-            except ArgumentParserError as e:
-                self.fail('Argument parser failed on {}, {}'.format(
-                    commandline,
-                    e.message))
+        tmp_config = os.path.join(self.test_dir, '.forseti')
+        with mock.patch.dict(
+            os.environ, {'FORSETI_CLIENT_CONFIG': tmp_config}) as mock_config:
```
ahoying: nit: you don't need to use 'as var' if you don't need to access the var in the context.

carise:
Thanks for the review!

PR #888, opened 2017-12-12
Add Forseti overview

Tabrarian:
Adding Forseti overview from doc draft.


_docs/overview/overview.md
```diff
@@ -0,0 +1,96 @@
+---
+title: Forseti conceptual overview
+order: 001
+hide: 
+  left_sidebar: true
+---
+
+# {{ page.title }}
+
```
mirons-google: Nit, I think we've standardized on no space between the heading and its content.

_docs/overview/overview.md
```diff
@@ -0,0 +1,96 @@
...
+Forseti Security is a collection of community-driven, open-source tools to help
+you improve the security of your Google Cloud Platform (GCP) environments.
+Forseti includes core modules that you can enable, configure, and execute
+independently of each other. Community contributors can also develop add-on
+modules to offer unique capabilities. Forseti's core modules work together to
+provide their respective features, and provide a foundation that addons can
+build upon.
+
+Get started with
+[Forseti Security]({% link _docs/quickstarts/forseti-security/index.html %}).
```
mirons-google: This link and every other one on this page should be corrected to point to the .md extension .html.

_docs/overview/overview.md
```diff
@@ -0,0 +1,96 @@
...
+make sure that there isn't any unauthorized access to your GCP resources.
+
+## How Forseti Security works
+
+When you install Forseti Security, you can configure and deploy core Forseti
+Security modules. The core Forseti Security modules work together to take a
+snapshot of your GCP resources, monitor those resources, and notify you of
+access policy changes.
+
+![Forseti architecture diagram](/images/docs/overview/diagram.png)
```
mirons-google: Two things.

1) This overview only applies to 2.0 (Q1-2018). We either need to remove it or note as such.
2) We should reference the (#) annotations in the image with details.

_docs/overview/overview.md
Tabrarian: I'm not sure what you mean. On other pages, I see # {{ page.title }} followed by a blank line before the content. Can you clarify?

mirons-google: Ah, you're correct, I was wrong. Disregard.

_docs/overview/overview.md
Tabrarian: Fixed, thanks, and apologies for the oversight!

_docs/overview/overview.md
Tabrarian: I'll remove it for now and work offline on annotation notes for when 2.0 is ready.

PR #893, opened 2017-12-13
Implement dispatch logic to inventory each project on a seperate thread.

ahoying:
* Use concurrent.futures to setup a threadpool with 10 workers.
* Dispatch project resources to the threadpool.
* Implement write locking on the storage.

Baseline run time for our test organization:
5m31s

Parallel run time for our test organization using this branch:
1m12s

Data in inventory identical.

### DIFF
--- a/google/cloud/forseti/services/inventory/base/resources.py
+++ b/google/cloud/forseti/services/inventory/base/resources.py
@@ -111,6 +111,15 @@ class Resource(object):
     def get_warning(self):
         return '\n'.join(self._warning)
 
+    def try_accept(self, visitor, stack=None):
+        """Handle exceptions on the call the accept."""
+        try:
+            self.accept(visitor, stack)
+        except Exception as e:
+            self.parent().add_warning(e)
+            visitor.update(resource.parent)
+            visitor.on_child_error(e)
+
     def accept(self, visitor, stack=None):
         stack = [] if not stack else stack
         self._stack = stack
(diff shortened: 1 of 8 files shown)


google/cloud/forseti/services/inventory/crawler.py
```diff
@@ -12,15 +12,18 @@
...
-from google.cloud.forseti.services.inventory.base import gcp
+import threading
+from concurrent.futures import ThreadPoolExecutor
+
 from google.cloud.forseti.services.inventory.base import crawler
+from google.cloud.forseti.services.inventory.base import gcp
+from google.cloud.forseti.services.inventory.base import resources
 
 
 class CrawlerConfig(crawler.CrawlerConfig):
```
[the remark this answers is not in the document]
ahoying: Done.

google/cloud/forseti/services/inventory/crawler.py
```diff
@@ -116,15 +135,110 @@ def update(self, resource):
...
+        """Process items in the queue until the shutdown event is set."""
+        while not self._shutdown_event.is_set():
+            try:
+                (callback, resource) = self._dispatch_queue.get(timeout=1)
+            except Empty:
+                continue
+
+            try:
+                callback()
+            except Exception as e:
```
felixbb: Is it possible to have this handler in the callback?

The reason I'm asking is that I believe what the handler does is a implementation detail of the resource so it should be part of the callback closure. This way, the crawler just schedules callables and does really not have to care what these callables do.

google/cloud/forseti/services/inventory/base/resources.py
```diff
@@ -124,19 +121,19 @@ def accept(self, visitor, stack=None):
...
                 for resource in yielder.iter():
                     res = resource
+                    new_stack = stack + [self]
 
-                    def call_accept():
-                        res.accept(visitor, stack + [self])
-
-                    if res.is_leaf():
-                        call_accept()
+                    if res.type() != 'project':
```
felixbb: Nit: If we could make this a res.should_dispatch() with a base implementation of "return False" and a Project class implementation of "return True", that would be better documentation of what's going on.

ahoying: Fixed.

google/cloud/forseti/services/inventory/crawler.py
ahoying: Fixed.

ahoying:
PTAL.

PR #919, opened 2018-01-04
Migrate Instance Network Scanner to work with 2.0 data model

### LINKED ISSUE #917: Integrate instance network inferface scanner with 2.0 data model

blueandgold:
There are dependencies here (the instance type and dao) that would impact the IAP scanner.  Since the IAP is more complicated, I think the changes should not be done until the IAP scanner is being worked on.

Fixes #917


blueandgold:
After discuss with Andrew, will put this PR on hold until after IAP dependencies are updated when the IAP scanner is migrated to 2.0 data model.  The code in this PR will then be much cleaner.

PR #928, opened 2018-01-05
Update LICENSE

michaelawyu:
As requested in b/70893145, any repository that is expected to be Apache 2.0 should have the exact license file from Apache with no modifications.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,3 @@
-
                                  Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -173,3 +172,30 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
blueandgold: The year 2018 needs to be filled in?
Also, what should be the copyright owner?  Please see this AUTHORS file:
https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/AUTHORS

tswast: No, actually it shouldn't be filled in. According to the OSPO office the license file should be text version of https://www.apache.org/licenses/LICENSE-2.0, unchanged.

tswast: It should be filled in in the source code headers, though.

blueandgold: I see, we have those.  Thanks for responding!

blueandgold:
The travis errors are not related to this PR, so merging.

PR #934, opened 2018-01-09
Allow users to choose a log level when running the application, added more logging to explain

joecheuk:


google/cloud/forseti/services/cli.py
```diff
@@ -23,14 +23,20 @@
...
 
 def define_playground_parser(parent):
     """Define the playground service parser.
 
     Args:
         parent (argparser): Parent parser to hook into.
     """
+
+    if parent is None:
+        LOGGER.warn("parent is None")
```
ahoying: In general, I'm not sure it is worthwhile to log a missing parameter if the function or method will throw an AttributeError exception due to the missing parameter. It may be better to use assert(parent is not None,  'parent is None') so these can be caught during testing, but are not checked when python is started with -O if we want to run without assertions.

This applies to all the locations below where the log message is just noting missing parameters.

google/cloud/forseti/services/cli.py
```diff
@@ -163,6 +169,10 @@ def define_inventory_parser(parent):
...
```
ahoying: This module uses both single quotes and double quotes around text inconsistently. We should use the same quote type everywhere in a module unless there is a specific case not to, for example 'This isn\'t right' would be better as "This isn't right" even if the rest of the code using single quotes.

I know this file was already messy before you modified it, but it is probably worth cleaning up in the cl.

https://google.github.io/styleguide/pyguide.html#Strings

google/cloud/forseti/services/cli.py
joecheuk: I will go ahead and remove the unnecessary log statements but I don't think it's a good idea to put  assert statements in between the code, do you mind if i just leave that part out?

ahoying: That's fine, we don't make a lot of use of assert today, so no reason to add it here now.

google/cloud/forseti/services/cli.py
joecheuk: I agreed it's better to keep the style consistent. Can we open a separate issue just for cleaning up the codes?

ahoying: SGTM.

google/cloud/forseti/services/dao.py
```diff
@@ -1164,7 +1222,9 @@ def add_member(cls,
             if len(parents) != len(parent_type_names):
                 msg = 'parents: {}, expected: {}'.format(
                     parents, parent_type_names)
-                raise Exception('Parent not found, {}'.format(msg))
+                error_message = 'Parent not found, {}'.format(msg)
+                LOGGER.error(error_message)
+                raise Exception(error_message)
```
jiyunyao: I suggest logger here goes to the exception handling. So all exception get logged.

PR #943, opened 2018-01-11
Fixing the path to the credentials.

mirons-google:


mirons-google:
FYI this might also require more debugging/PRs as I can only trigger the deploy function through a build and not a PR.

PR #946, opened 2018-01-11
Addressing pylint findings.

mirons-google:
This fixes the dev branch and thus our build.

### DIFF
--- a/google/cloud/security/common/util/email_util.py
+++ b/google/cloud/security/common/util/email_util.py
@@ -19,10 +19,8 @@ import os
 import urllib2
 
 import jinja2
-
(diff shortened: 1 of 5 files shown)


google/cloud/security/common/util/email_util.py
```diff
@@ -20,10 +20,11 @@
 
 import jinja2
 
-from retrying import retry
 import sendgrid
 from sendgrid.helpers import mail
 
+from retrying import retry
```
blueandgold: retrying is an installed 3rd party library just like jinja and sendgrid; perhaps try grouping all of them together?

PR #959, opened 2018-01-16
Update jquery version to the latest.

mirons-google:


mirons-google:
I QA'd it by hand with minimal testing. We have no webdriver tests currently.

PR #960, opened 2018-01-16
Website tweaks

mirons-google:
This
- Fixes the site by forcing the appending of / to all directory paths
- Removed liquid code that strips index.html as it's required on app engine
- Tweaks the .travis.yml to disable the assumption of extension as we need to now check it
- Removes the 404.html and CNAME files as they are no longer needed.


mirons-google:
This is now ready for review.

PR #966, opened 2018-01-18
Update IAM Scanner to Use 2.0 Data Model

blueandgold:
This PR also involves adding the IAM policies as resources in the data model.


google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -600,6 +602,30 @@ def _convert_computeproject(self, computeproject):
...
                 parent=parent))
 
+    def _convert_iam_policy(self, iam_policy):
+        """Convert an IAM policy to a database object.
+
+        Args:
+            iam_policy (object): IAM policy to store.
+        """
+        # TODO: Figure out why bucket IAM policy can not be imported.
+        unsupported_iam_policy_type = ['bucket']
```
ahoying: In _convert_bucket, the resource isn't added to the resource cache, if you update _convert_bucket to look more like _convert_dataset then it will work, and you won't need to skip it here.

tests/scanner/audit/iam_rules_engine_test.py
```diff
@@ -57,6 +56,26 @@ def setUp(self):
...
+        self.mock_project2_policy_resource = mock.MagicMock()
+        self.mock_project2_policy_resource.full_name = (
+            'project2/my-project-2')
+
+        self.mock_project3_policy_resource = mock.MagicMock()
+        self.mock_project3_policy_resource.full_name = (
+            'organization/778899/folder/333/project/my-project-3')
+
         # patch the organization resource relation dao
         self.patcher = mock.patch('google.cloud.forseti.common.data_access.org_resource_rel_dao.OrgResourceRelDao')
```
ahoying: This dao should no longer be called, so it should no longer need to be mocked, right?

ahoying:
You should also update the yaml config to turn this scanner back on by default.

ahoying:
Also, are there any more legacy DAO files that can be deleted as they are no longer used post conversion to the model?

google/cloud/forseti/services/model/importer/importer.py
ahoying: I'm okay with keeping this a todo, although the change is small enough it might be nice to include it with this PR.

blueandgold: Done, fixed.  :)

blueandgold:
There is one dao that can be deleted.  Let me do it a separate PR, to make sure this one doesn't break.

PR #968, opened 2018-01-19
Model service FR

jiyunyao:
A new column called description in data model
list_model yields stream instead of single message, warnings are not displayed when listing model
new get_model enable get details of a model by name or handle
use_model is improved to use get_model


jiyunyao:
The new description field is json and looks like:
description: "{\"source\": \"inventory\", \"prestine\": true, \"source_info\": \"<InventoryIndex(id=\'2\', version=\'1\', timestamp=\'2018-01-19 01:17:52\')>\"}"

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -67,6 +68,10 @@ def run(self):
         """Runs the import."""
 
         self.session.add(self.model)
+        self.model.add_description(json.dumps({
+            "source":"empty",
+            "prestine":True
```
blueandgold: Should this be spelled as "pristine" instead?  BTW: what is the significance or usage of this field?

jiyunyao: If a data model is touched by a playground manipulation, pristine will be False, so scanner knows this model has been touched.
In addition, you can know the source of a data model from description and if it is from an inventory, you can tell the exact info of the source.
Since we use json for description, new items can be added into the description for future proof.

PR #979, opened 2018-01-19
Firewall Scanner Bug Fix

mwwolters:


google/cloud/forseti/scanner/scanners/firewall_rules_scanner.py
```diff
@@ -163,13 +164,16 @@ def _find_violations(self, policies):
         policies = itertools.chain(policies)
         all_violations = []
         LOGGER.info('Finding firewall policy violations...')
+        project_policies = defaultdict(list)
```
ahoying: I recommend updating _retrieve to create the dict instead of building it in _find_violations. You could use enumerate() on the scanner_iter and store the count of policies from the value of the counter, rather than the length of the list.

Issue #757, opened 2017-11-01
Use tempfile to create the violations.json


carise:
We currently write the violations.json to /tmp but do not clean it up after sending the email. Probably can leverage tempfile to manage this.

arbeit:
Is this issue still open? If so, I'd ike to take a stab at it.

mirons-google:
@arbeit please do! I've added it to this quarters milestone.

arbeit:
Cool. Thank you!

PR #985, opened 2018-01-21
Clean up temp file after emailing violations

arbeit:
Hello there!

This PR fixes issue #757. The tests ascertain that the temp files after `run()` are the same ones as before.
In order to test that nicely I wanted a specific/isolated temp directory upon which the code under test would operate.
That required some fiddling around and module level setup prior to the first import of `tempfile` which does not pay any attention to `os.environ['TMPDIR']` afterwards.

pylint passes as do all the unit tests. Please take a look :-)

### DIFF
--- a/google/cloud/security/notifier/pipelines/email_violations_pipeline.py
+++ b/google/cloud/security/notifier/pipelines/email_violations_pipeline.py
@@ -57,7 +57,6 @@ class EmailViolationsPipeline(bnp.BaseNotificationPipeline):
                                                       notifier_config,
                                                       pipeline_config)
         self.mail_util = EmailUtil(self.pipeline_config['sendgrid_api_key'])
-        self.tmp_file = tempfile.NamedTemporaryFile(delete=False)
 
     def _get_output_filename(self):
         """Create the output filename.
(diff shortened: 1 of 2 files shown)


arbeit:
Hmm .. I am a Googler, so, the CLA should be fine.

tests/notifier/pipelines/email_violations_pipeline_test.py
```diff
@@ -0,0 +1,97 @@
...
+        self.gvp.run()
+        tmp_files_after = glob.glob(tmp_dir)
+        self.assertEquals(tmp_files_before, tmp_files_after)
+
+    def test_make_attachment_can_read_tempfile(self):
+        """Test that _make_attachment() can read the temp file."""
+        violations = '["violation: abc", "violation: def"]'
+        attachment = self.gvp._make_attachment()
+        decoded_content = base64.b64decode(attachment.content)
+        self.assertEquals(decoded_content, violations)
```
blueandgold: nit: We follow the assert convention of ```assert(expected_value, actual_value)```.  So in this case, we should have ```self.assertEquals(violations, decoded_content)```

blueandgold: Actually, let me take care of this for you.

google/cloud/security/notifier/pipelines/email_violations_pipeline.py
```diff
@@ -70,34 +71,24 @@ def _get_output_filename(self):
...
-        return output_file_name
-
     def _make_attachment(self):
         """Create the attachment object.
 
         Returns:
             attachment: SendGrid attachment object.
         """
-        output_file_name = self._write_temp_attachment()
+        tmp_file = tempfile.NamedTemporaryFile(delete=False)
```
mirons-google: This would be better using the "with" clause.

mirons-google: Also, what if this raises an exception, e.g. disk is full. Can we catch this properly and handle it? Otherwise the references below to tmp_file.name won't work.

arbeit: I can change the code to use a `with` clause. Do I need to open an issue for that or shall I just prepare a quick PR without it?

blueandgold: Since I merged the PR before seeing Michael's comment, please just submit a follow-up PR.  The code should look something like:

```
        with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
            tmp_file.write(parser.json_stringify(self.violations))

        attachment = self.mail_util.create_attachment(
            file_location=tmp_file.name,
            content_type='text/json',
            filename=self._get_output_filename(),
            disposition='attachment',
            content_id='Violations'
        )
        os.unlink(tmp_file.name)
        return attachment
```

Issue #894, opened 2017-12-13
Scanner violation table in 2.0-dev


jiyunyao:
on 2.0-dev, forseti.services.scanner.

Violation table should not use the same tag as the data model if not managed by data model service. The auditor/scanner is an independent service, it should have a different name.

User should be able to delete violation table

blueandgold:
I will move the model-specific violation tables to a centralized, standalone table, with an extra column for the model name.

blueandgold:
Fixed by PR #1071

PR #993, opened 2018-01-22
Store all the violations in a centralized table.

blueandgold:
This is to address the main item in #894, by moving the violation tables from a per-model basis, and into a centralized table.  Will update tests if this PR's approach is okay.

Also, there is a secondary item to make it clear where the violations is coming from.  Do you have any suggestions?  I was thinking to use the ```full_name```, that has the full hierarchy path of the resource.  How do you think?

### DIFF
--- a/google/cloud/forseti/services/scanner/dao.py
+++ b/google/cloud/forseti/services/scanner/dao.py
@@ -26,13 +26,12 @@ from google.cloud.forseti.services import db
 
 # pylint: disable=no-member
 
-def define_violation(model_name, dbengine):
+def define_violation(dbengine):
     """Defines table class for violations.
 
     A violation table will be created on a per-model basis.
 
     Args:
-        model_name (str): name of the current model
         dbengine (engine): sqlalchemy database engine
 
     Returns:
--- a/google/cloud/forseti/services/scanner/dao.py
+++ b/google/cloud/forseti/services/scanner/dao.py
@@ -40,9 +40,7 @@ def define_violation(model_name, dbengine):
     """
 
     base = declarative_base()
-    # pylint: disable=too-many-format-args
-    violations_tablename = 'violations'.format(model_name)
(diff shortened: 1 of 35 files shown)


google/cloud/forseti/services/scanner/dao.py
```diff
@@ -40,14 +40,17 @@ def define_violation(model_name, dbengine):
     """
 
     base = declarative_base()
-    violations_tablename = '{}_violations'.format(model_name)
+    # pylint: disable=too-many-format-args
```
ahoying: Remove the pylint disable and remove the .format() from the string. The tablename is just 'violations'

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -40,14 +40,17 @@ def define_violation(model_name, dbengine):
...
+    # pylint: enable=too-many-format-args
 
     class Violation(base):
         """Row entry for a violation."""
 
         __tablename__ = violations_tablename
 
         id = Column(Integer, primary_key=True)
+        model_handle = Column(String(256))
         resource_type = Column(String(256), nullable=False)
```
ahoying: Please add resource_id back into the model, this got dropped. Also I agree that adding resource_full_name is a good idea.

google/cloud/forseti/services/scanner/dao.py
blueandgold: Done.

google/cloud/forseti/services/scanner/dao.py
blueandgold: resource_id is added back; also taking the chance to fix how resource_id is being set that will make it more easy to identify which resource is the violation coming from.

I also started to implement the full_name, but it grew to be a sweeping change.  Let me do this in a follow-up PR, as I need to re-think how to make the easiest change.

Issue #615, opened 2017-08-31
group members inventory marked as success when groups failed


mirons-google:
Actual output:

Resource Name | Status | # in Snapshot
-- | -- | --
organizations | SUCCESS | 1
folders | SUCCESS | 12
folder_iam_policies | SUCCESS | 27
projects | SUCCESS | 100
appengine | SUCCESS | 31
backend_services | SUCCESS | 16
bigquery_datasets | SUCCESS | 0
buckets | SUCCESS | 74
buckets_acl | SUCCESS | 224
cloudsql | SUCCESS | 1
firewall_rules | SUCCESS | 434
forwarding_rules | SUCCESS | 65
instance_group_managers | SUCCESS | 25
instance_groups | SUCCESS | 41
instance_templates | SUCCESS | 34
instances | SUCCESS | 167
project_iam_policies | SUCCESS | 744
service_accounts | SUCCESS | 180
groups | FAILURE | Unknown
group_members | SUCCESS | Unknown
org_iam_policies | SUCCESS | 478

mirons-google:
@blueandgold can you confirm if this is fixed?

arbeit:
I would like to work on this if it is still outstanding and nobody else is doing it.

blueandgold:
Thanks @arbeit , please take this!

PR #997, opened 2018-01-24
fixes: group members inventory marked as success when groups failed

arbeit:
Hello there!

This fixes issue #615. I am seeing a weird issue though. When I remove the `import mock` line in `inventory_loader_test.py` and run the tests (via `python -m unittest tests.inventory.inventory_loader_test.InventoryLoaderTest`) I get the following stack trace: https://pastebin.com/hpygZTnK

With the `import mock` in place all tests pass. Any ideas why this might be?

### DIFF
--- a/google/cloud/security/inventory/inventory_loader.py
+++ b/google/cloud/security/inventory/inventory_loader.py
@@ -193,6 +193,9 @@ def _postprocess_statuses(pipelines, run_statuses):
     Example: if the `groups` pipeline failed then the `group_members` pipeline
     should be marked as a failure as well.
 
+    Please note: this function will modify the data passed to it via the
+    arguments below if/as needed.
+
     Args:
         pipelines (list): List of pipelines that were run.
         run_statuses (list): a list of booleans indicating whether each
--- a/google/cloud/security/inventory/inventory_loader.py
+++ b/google/cloud/security/inventory/inventory_loader.py
@@ -201,7 +201,21 @@ def _postprocess_statuses(pipelines, run_statuses):
         list: a (possibly adjusted) list of booleans indicating whether each
         pipeline completed successfully or not.
     """
-    pass
+    indices = dict(
+        (v, i) for (i, v) in enumerate(p.RESOURCE_NAME for p in pipelines))
+    grps_idx = indices.get('groups')
+    grp_members_idx = indices.get('group_members')
+    # Do nothing unless the status for both 'group_members' and 'groups' is
+    # present and the latter is `False`.
+    if (grps_idx is None or grp_members_idx is None or run_statuses[grps_idx]):
+        return run_statuses
+
+    if run_statuses[grp_members_idx]:
+        run_statuses[grp_members_idx] = False
+        pipelines[grp_members_idx].status = 'FAILURE'
+
+    return run_statuses
+
 
 def _complete_snapshot_cycle(inventory_dao, cycle_timestamp, status):
     """Complete the snapshot cycle.
--- a/google/cloud/security/inventory/inventory_loader.py
+++ b/google/cloud/security/inventory/inventory_loader.py
@@ -163,7 +163,7 @@ def _run_pipelines(pipelines):
         pipelines (list): List of pipelines to be run.
 
     Returns:
-        list: a list of booleans whether each pipeline completed
+        list: a list of booleans indicating whether each pipeline completed
             successfully or not.
     """
     # TODO: Define these status codes programmatically.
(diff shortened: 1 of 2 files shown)


google/cloud/security/inventory/inventory_loader.py
```diff
@@ -185,8 +185,43 @@ def _run_pipelines(pipelines):
                          exc_info=True)
             pipeline.status = 'FAILURE'
         run_statuses.append(pipeline.status == 'SUCCESS')
+    _postprocess_statuses(pipelines, run_statuses)
```
blueandgold: Should the result that is returned from ```_postprocess_statuses()``` be used here?  Otherwise, what is the state of ```run_statues``` and ```pipelines``` here?

google/cloud/security/inventory/inventory_loader.py
```diff
@@ -185,8 +185,43 @@ def _run_pipelines(pipelines):
...
+    # present and the latter is `False`.
+    if (grps_idx is None or grp_members_idx is None or run_statuses[grps_idx]):
+        return result
+
+    if run_statuses[grp_members_idx]:
+        result = True
+        run_statuses[grp_members_idx] = False
+        pipelines[grp_members_idx].status = 'FAILURE'
+
+    return result
```
blueandgold: The modified ```pipelines``` might need to be returned, so that it can be used in the outer execution for email template.

for example:
```
    return pipelines, run_statuses
````
Then, your caller can be:
```
pipelines, run_statuses = _postprocess_statuses(pipelines, run_statuses)
```

File: google/cloud/security/inventory/inventory_loader.py [hunk 3]
```diff
@@ -185,8 +185,43 @@ def _run_pipelines(pipelines):
                          exc_info=True)
             pipeline.status = 'FAILURE'
         run_statuses.append(pipeline.status == 'SUCCESS')
+    _postprocess_statuses(pipelines, run_statuses)
     return run_statuses
 
+def _postprocess_statuses(pipelines, run_statuses):
```
blueandgold: This method is very specific to processing group and group members.  So, I would suggest to change it to:
```
def _adjust_group_member_run_status_if_necessary(pipelines, run_statuses)
```

google/cloud/security/inventory/inventory_loader.py
blueandgold: ```result``` is not a list as currently used.  If you return the updated ```run_statuses```, the it will be a list.

google/cloud/security/inventory/inventory_loader.py
arbeit: `_postprocess_statuses()`  modifies the actual data/parameters passed to it.

google/cloud/security/inventory/inventory_loader.py
arbeit: There's no need to return, the actual data passed to the function is modified (using python's call-by-object mechanism)

arbeit: `_postprocess_statuses()' (as it stands) is actually a procedure and not a function that needs to return a result. If you like I can remove the result value altogether.

google/cloud/security/inventory/inventory_loader.py
arbeit: That's a bit long :) but can do. I guess the idea was that the current `_postprocess_statuses()` implementation is just a first step but maybe we will need to add more to it later.

google/cloud/security/inventory/inventory_loader.py
arbeit: See e.g. http://cryptroix.com/2016/10/25/python-call-object/

google/cloud/security/inventory/inventory_loader.py
arbeit: The `bool` test result (indicating whether any adjustments to the passed date were made) just made it a bit easier to test.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Ah, interesting; good thing to learn.  :)  So, yes, in this case, this is fine and no change necessary.

google/cloud/security/inventory/inventory_loader.py
blueandgold: I think this method should be renamed.  If there are any other postprocess adjustments to be in separate helper methods.

google/cloud/security/inventory/inventory_loader.py
blueandgold: Yah, then just remove this return.

blueandgold: It looks like the ```result``` is not used and can be dropped.  Instead of ```return True```, you can just do a simple ```return```?

google/cloud/security/inventory/inventory_loader.py
arbeit: Looking for a short name, pylint is complaining :-P

    C:191, 0: Function name "_adjust_group_members_run_status" doesn't conform to u'[a-z_][a-z0-9_]{2,30}$' pattern (invalid-name)

arbeit: Done with all changes. Please take a look. Thanks!

Issue #964, opened 2018-01-17
Consolidate build and test dependencies


mirons-google:
Today we have dependencies scattered across various sources
- setup.py (build, test, install)
- pip list in scripts/docker/base
- apt-get list in scripts/docker/base
- and the dev environment [docs](https://forsetisecurity.org/docs/howto/deploy/dev-setup.html#setting-up-local-environment)

These should be consolidated into one list that supports consistent versions across local, remote (travis), and in VM environments (docker).

mirons-google:
There is still work to be done here in setup.py (one list of requirements) and requirements.txt

mirons-google:
Related to #1420 and #1418

PR #1000, opened 2018-01-24
Consolidate install, build, and test functionality and dependencies

### LINKED ISSUE #940: check_libyaml.c:2:18: fatal error: yaml.h: No such file or directory compilation terminated
When running ```python setup.py install``` on ubuntu, installation of pyyaml is throwing this error from check_libyaml.c file.

After looking into this issue, this is not critical as pyyaml will fall back to the native python parser.

But to fix this, it looks like either ```libpython-dev``` or ```python-dev``` needs to be installed.
https://github.com/aws/aws-cli/issues/2290
(Our documentation already mentions python-dev.)

Or perhaps this:
https://github.com/NVIDIA/DIGITS/issues/774
sudo apt-get install libyaml-cpp-dev

How to test:
http://rmcgibbo.github.io/blog/2013/05/23/faster-yaml-parsing-with-libyaml/

```
Searching for PyYAML==3.12
Reading https://pypi.python.org/simple/PyYAML/
Downloading https://pypi.python.org/packages/6b/f0/a0250248ea260d55748fff586d89a32afbb22656f4498b08d2636a48d4ec/PyYAML-3.12.zip#md5=aa5e762e79b19bb80dbcf04f82178832
Best match: PyYAML 3.12
Processing PyYAML-3.12.zip
Writing /tmp/easy_install-6j71iG/PyYAML-3.12/setup.cfg
Running PyYAML-3.12/setup.py -q bdist_egg --dist-dir /tmp/easy_install-6j71iG/PyYAML-3.12/egg-dist-tmp-qzMjPm
build/temp.linux-x86_64-2.7/check_libyaml.c:2:18: fatal error: yaml.h: No such file or directory
compilation terminated.

libyaml is not found or a compiler error: forcing --without-libyaml
(if libyaml is installed correctly, you may need to
 specify the option --include-dirs or uncomment and
 modify the parameter include_dirs in setup.cfg)
zip_safe flag not set; analyzing archive contents...
Moving PyYAML-3.12-py2.7-linux-x86_64.egg to /home/ep/.environments/forseti-security/lib/python2.7/site-packages
Adding PyYAML 3.12 to easy-install.pth file

Installed /home/ep/.environments/forseti-security/lib/python2.7/site-packages/PyYAML-3.12-py2.7-linux-x86_64.egg
```

mirons-google:
This PR does many significant things
- Resolves #940 and #964 by creating a standard list of pip, host (apt), and package requirements
- This creates a canonical way for developers (that matches our Travis build) leveraging Docker
  - To build a base docker image
  - Install forseti and it dependencies
  - Perform unittest and pylint with consistency
- This consolidates various build, install, deployment, scripts into one and names it installer
- This removes the unnecessary iam-explain deployment template
- Addresses various lint issues


deployment-templates/compute-engine/forseti-instance.py
```diff
@@ -205,8 +207,8 @@ def GenerateConfig(context):
 )"
 echo "$FORSETI_ENV" > $USER_HOME/forseti_env.sh
 
-sudo su $USER -c "$FORSETI_HOME/scripts/gcp_setup/bash_scripts/run_forseti.sh"
-(echo "{run_frequency} $FORSETI_HOME/scripts/gcp_setup/bash_scripts/run_forseti.sh") | crontab -u $USER -
+sudo su $USER -c "$FORSETI_HOME/setup/installer/scripts/forseti_run.sh"
```
blueandgold: Is there a reason to change ```run_forseti.sh``` to ```forseti_run.sh```.  It was actually pretty nice to read before.  If anything, this change should be to a noun, like ```forseti_runner.sh```. 

This also has been done elsewhere, and the reading of those files seems awkward.

mirons-google: To me it made more sense (especially in one directory) to organize them based on the object then the verb.

blueandgold: Then for organization purposes, we should go with all noun as with (most) of our module namings.  The object/verb style is really unnatural in English.  So, being easy to read & say will be super-awesome and help us so much down the road.

 ```forseti_runner.sh```.

Similiarly, use
```docker_builder.sh```
```docker_runner.sh```
```forseti_services_initializer.sh```

Comment (mirons-google) [reply deployment-templates/compute-engine/forseti-instance.py]: Sure, this should mostly be the case. I'm happy to switch conventions if we think it's best.

File: setup/utils/build_protos.py [hunk 2]
```diff
@@ -21,16 +21,27 @@
 
 
 def is_grpc_service_dir(files):
-    """Returns true iff the directory hosts a gRPC service."""
+    """Returns true iff the directory hosts a gRPC service.
```
blueandgold: If we are taking the opportunity to tidy up things, then this file would be better named as ```protos_builder.py```.

Then, the following code usage would be so much cleaner.  i.e. no import as, and the caller would be much more readable, as protos_builder.clean().
```
from setup.utils import build_protos as forseti_protos

forseti_protos.clean(abs_path)
forseti_protos.make_proto(abs_path)
```

deployment-templates/compute-engine/forseti-instance.py
```diff
@@ -155,35 +149,43 @@ def GenerateConfig(context):
...
+sudo apt-get install -y $(cat setup/dependencies/apt_packages.txt | grep -v "#" | xargs)
+
+# Forseti dependencies
+pip install --upgrade pip
+pip install --upgrade setuptools
+pip install -r setup/dependencies/pip_packages.txt
+
 # Set ownership of config and rules to $USER
-chown -R $USER {forseti_home}/configs {forseti_home}/rules {forseti_home}/scripts/gcp_setup/bash_sripts/run_forseti.sh
+chown -R $USER {forseti_home}/configs {forseti_home}/rules {forseti_home}/setup/installer/scripts/forseti_runner.sh
```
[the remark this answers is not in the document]
mirons-google: My thinking was:

`setup/installer/scripts` are scripts needed by the installer on the host machine.

`setup/scripts` are general purpose and used in the setup (travis, docker, future builders).

setup/utils/build_protos.py
mirons-google: done!

mirons-google:
I just committed a change to create a stub for running the installer.

Running it would now be `python setup/forseti_install.py`

mirons-google:
Another quick change to revert back to the suggested naming style.

setup/install.py
```diff
@@ -12,4 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""A setup_forseti wizard for assistance in installation."""
+""" Forseti Installer.
```
blueandgold: A couple of comments:
1: the name of this module should be a noun, i.e. instead of `install.py`, it should be `installer.py`
2: please add a description that this is for installing to GCP

setup/installer/installer_runner.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Set up the gcloud environment and Forseti prerequisites.
+""" GCP Installer runner.
```
blueandgold: The name `installer_runner.py` might be better as just `installer.py`.

setup/scripts/build_docker.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
blueandgold: Perhaps it's not a big deal, but the file naming here is not consistent.  Do you want to make it so as `docker_builder.sh`?

setup/scripts/run_docker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
blueandgold: Perhaps it's not a big deal, but the file naming here is not consistent.  Do you want to make it so as `docker_runner.sh`?

setup/scripts/run_docker_tests.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
blueandgold: ditto, here; Perhaps it's not a big deal, but the file naming here is not consistent.  Do you want to make it so as `docker_runner_tests.sh`?

setup/scripts/build_docker.sh
blueandgold: Thanks for explaining `setup/scripts` and `setup/installer/scripts`.  Do you want to clarify these, such as:
`setup/gcp`, so that you can have `setup/gcp/installer.py` and `setup/gcp/util/`
`setup/build`, and just put all the scripts there.  Do you think that we can drop the `scripts` as a directory?

setup/install.py
mirons-google: done

setup/installer/installer_runner.py
mirons-google: That creates a conflict between `setup/installer.py `and `setup/installer/installer_runner.py`

setup/scripts/build_docker.sh
mirons-google: Reverted to your original suggestion: `build_docker.sh` matching the naming convetion of the other scripts.

mirons-google: I think we should have a separation of scripts used on GCP deployment from those used to build and test on Travis. But I agree with the suggestion on renaming `setup/installer/installer_runner.py` to `setup/gcp/installer.py`

setup/scripts/run_docker.sh
mirons-google: Reverted to your original suggestion: `run_docker.sh` matching the naming convetion of the other scripts (verb_noun.sh)

setup/scripts/run_docker_tests.sh
mirons-google: Reverted to your original suggestion: `run_docker_tests.sh` matching the naming convetion of the other scripts (verb_name.sh).

deployment-templates/compute-engine/forseti-instance.py
```diff
@@ -155,19 +149,27 @@ def GenerateConfig(context):
...
+pip install --upgrade setuptools
+pip install -r setup/dependencies/pip_packages.txt
+
 # Set ownership of config and rules to $USER
-chown -R $USER {forseti_home}/configs {forseti_home}/rules {forseti_home}/scripts/gcp_setup/bash_sripts/run_forseti.sh
+chown -R $USER {forseti_home}/configs {forseti_home}/rules {forseti_home}/setup/gcp/scripts/run_forseti.sh
 
 # Build protos.
-python build_protos.py --clean
+python protos_builder.py --clean
```
blueandgold: should this also be reverted back to `build_protos.py`?

setup.py
```diff
@@ -49,17 +50,10 @@
     'SQLAlchemy>=1.1.9',
     'pygraph>=0.2.1',
     'unicodecsv>=0.14.1',
-]
-
-SETUP_REQUIRES = [
+    # Setup.
     'google-apputils>=0.4.2',
     'python-gflags>=3.1.1',
-    'grpcio',
```
ahoying: These should also be in setup.py. If already installed using the script they won't be recompiled from source. But someone just coming and installing from setup.py will get all required dependencies.

setup.py
```diff
@@ -101,9 +97,7 @@ def run(self):
     cmdclass={
         'install': PostInstallCommand,
     },
-    install_requires=SETUP_REQUIRES + INSTALL_REQUIRES,
-    setup_requires=SETUP_REQUIRES,
-    tests_require=INSTALL_REQUIRES + SETUP_REQUIRES + TEST_REQUIRES,
+    install_requires=INSTALL_REQUIRES,
     packages=find_packages(exclude=[
         '*.tests', '*.tests.*', 'tests.*', 'tests']),
     include_package_data=True,
```
ahoying: Below, please remove forseti_scanner and forseti_notifier.

setup/scripts/run_docker.sh
```diff
@@ -0,0 +1,30 @@
...
+    # Docker command does exist.
+    # Check to see if we're on Travis.
+    if [ ${TRAVIS+x} ]; then
+        # We are on Travis.
+        # Required for codecov.io to export coverage within a Docker container.
+        "CI_ENV=`bash <(curl -s https://codecov.io/env)`"
+        # Start the container for testing and codee verification.
+        docker run ${CI_ENV} -it -d --name build forseti/build /bin/bash
+    else
+        # We're not on docker, run without the CI_ENV environment variable.
```
ahoying: Nit: not on Travis.

setup/scripts/run_docker_tests.sh
```diff
@@ -13,12 +13,17 @@
...
-# Delete all running containers
-docker rm -f $(docker ps -a -q)
+echo "Running tests."
 
-docker build -t forseti/base -f scripts/docker/base . || exit 1
-docker build -t forseti/build -f scripts/docker/forseti --no-cache . || exit 1
-docker run -it -d --name build forseti/build /bin/bash || exit 1
+# Check to see if we're on Travis.
+if [ -z ${TRAVIS+x} ]; then
+    # We are on Travis.
```
ahoying: Comments are backwards, right? -z means not on Travis.

ahoying: Really the order of the if should be the same as in other files. Where the on travis is first, and we run coverage, not on travis is second and we only run tests. So remove -z and switch the blocks around.

setup.py
mirons-google: Done and confirmed they're not rebuilt if you're on Travis/Docker.

setup.py
mirons-google: done!

setup/scripts/run_docker.sh
mirons-google: Nice catch, thanks and done.

setup/scripts/run_docker_tests.sh
mirons-google: Nice catch, thanks and done.

mirons-google:
All done. PTAL @ahoying

PR #1020, opened 2018-01-30
Make the notifier into a service.

blueandgold:
This is the first step to integrate the notifier with Forseti 2.0.  There is still more work on the back end, i.e. retrieve the violations from the new centralized violations table, using the inventory index id.  But will save that for another PR.


google/cloud/forseti/services/client.py
```diff
@@ -94,6 +96,39 @@ def run(self, config_dir):
...
+        Args:
+            config_file (str): Path the to the forseti config file.
+            inventory_index_id (int): Inventory Index Id.
+
+        Returns:
+            Any service response.
+        """
+
+        request = notifier_pb2.RunRequest(
+            config_file=config_file,
```
jiyunyao: config file location should not be sent to the server. If we are ever going to host the server and the client on different machines, it will be a problem. Maybe send the file content or necessary parameters.

google/cloud/forseti/notifier/notifier.py
```diff
@@ -151,16 +150,20 @@ def process(message):
             payload.get('email_description'))
         return
 
-def main(_):
-    """Main function.
+# pylint: disable=unused-argument
+def run(forseti_config, inventory_index_id, service_config=None):
```
jiyunyao: The args here assume forseti_config is a dict. But from the cli to service and notifier it is all referred as config_file and in cli is referred to as a path. If the customer input "forseti notifier --config_file ./PATH/ config" it will cause problem.

blueandgold: Dropping this.  This is now server-side flag.

google/cloud/forseti/services/client.py
blueandgold: Good catch.  But instead of sending the file content, I'm going to keep this config file on the server-side, and reference it via flag on server startups.

PR #1039, opened 2018-02-02
Associate Violations to the Inventory Index Id, full name, rather than Model Handle

blueandgold:
I also had to tweak the model description.  Otherwise, the InventoryIndex was a string repr, which I didn't want to parse.

This is only for bigquery dataset only.  I will tackle the other resources next.


google/cloud/forseti/services/scanner/dao.py
```diff
@@ -47,7 +47,7 @@ class Violation(base):
         __tablename__ = violations_tablename
 
         id = Column(Integer, primary_key=True)
-        model_handle = Column(String(256))
+        inventory_id = Column(String(256))
```
joecheuk: Should this be inventory_id or inventory_index_id?  I think it's better to have this as a foreign key and reference back to the inventory/inventory_index table.

blueandgold: This is a good point.  I originally had it as `inventory_index_id`, but then thought to be consistent with the rest of the usage of `inventory_id` elsewhere in the code base.  Do you think that we should change this everywhere?  As there is a difference between the two.

blueandgold: After more discussions, this should be better as ```inventory_index_id```, so that we do not confuse it with ```inventory_id```

google/cloud/forseti/scanner/scanners/base_scanner.py
```diff
@@ -65,13 +65,18 @@ def _output_results_to_db(self, violations):
         Returns:
             list: Violations that encountered an error during insert.
         """
+        model_description = (
+            self.service_config.model_manager.get_description(self.model_name))
+        inventory_index_id = (
```
jiyunyao: This can be changed to source_id = "{}-{}".format(model_description.get(source), model_description.get('source_info').get('inventory_index_id'))), if we want to have a generic id-name for scanning result.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -47,13 +47,15 @@ class Violation(base):
         __tablename__ = violations_tablename
 
         id = Column(Integer, primary_key=True)
-        model_handle = Column(String(256))
+        inventory_index_id = Column(String(256))
```
jiyunyao: Maybe we can change it to a more generic name like source_id.

google/cloud/forseti/scanner/scanners/base_scanner.py
blueandgold: Please see comment below.

google/cloud/forseti/services/scanner/dao.py
blueandgold: So, from our discussions with joe, no change here.  We will keep this specific column for `inventory_index_index`, and add more columns if we have other id's to store.

PR #1050, opened 2018-02-07
Enable calling build_protos through setup.py

### LINKED ISSUE #1049: 2.0 - Developers should be able to build protos within installing using setup.py
When build_protos.py was moved into a library, the ability to just build the protos was lost. This functionality is needed for local development and testing. I recommend adding a new command to setup.py:

...

class BuildProtosCommand(install):
    """Post installation command."""

    def run(self):
        build_protos()

    cmdclass={
        'install': PostInstallCommand,
        'build_protos': BuildProtosCommand,
    },

mirons-google:
This adds to helper commands through setuptools that cleans and builds protos. This resolves #1049


setup/utils/build_protos.py
```diff
@@ -120,12 +107,10 @@ def make_proto(path):
 
 def main():
```
ahoying: Is main() still needed if it can't be run directly? Maybe return an error that says run from setup.py?

setup/utils/build_protos.py
```diff
@@ -40,13 +42,13 @@ def clean(path):
...
     # this scripts's location as __file__.
-    logging.info("Cleaning out compiled protos.")
+    LOGGER.info("Cleaning out compiled protos.")
     cwd = os.path.dirname(path)
 
     # Find all the .proto files.
     for (root, dirs, files) in os.walk(cwd):
         if is_grpc_service_dir(files):
-            logging.info('Skipping grpc service directory: %s', root)
+            LOGGER.info('Skipping grpc service directory: %s', root)
```
ahoying: Is there a reason that clean doesn't clean up the service directories?

setup/utils/build_protos.py
mirons-google: I opted to remove it for simplicity sake.

setup/utils/build_protos.py
mirons-google: This I'm not sure about. @jiyunyao do you know why we don't clean service directories of built protos?

PR #1062, opened 2018-02-09
Update references to IAM Explain

mirons-google:
More simply it's just Explain now. This PR is devoid of applying any logic to whether the changed lines are still needed, it's simply a renaming exercise.


mirons-google:
This resolves #996

PR #1064, opened 2018-02-09
Make the Notifier work with the new 2.0 Violations

blueandgold:


google/cloud/forseti/notifier/notifier.py
```diff
@@ -77,25 +76,26 @@ def find_pipelines(pipeline_name):
...
-    except db_errors.MySQLError as err:
-        LOGGER.error('Error getting latest snapshot timestamp: %s', err)
-
-    return latest_timestamp
+    allowed_status = ('SUCCESS', 'PARTIAL_SUCCESS')
+    inventory_index_id = ''
+    with service_config.scoped_session() as session:
+        for item in DataAccess.list(session):
+            if item.status in allowed_status:
+                inventory_index_id = item.id
```
joecheuk: Do we need to break out of the loop once we found the inventory_index_id?

blueandgold: No, because we are looking for the last record here, so the loop must continue to the end.  This is not my preferred way, but sqlalchemy doesn't support finding the last record, and I don't see an easy way to sort the datetime which is a string.

blueandgold: I tested the string sort for the datetime stamp and it works.  So, following up with PR #1070 to use sort, rather than the loop.

PR #1069, opened 2018-02-09
Install and expose glcoud on the docker testing image.

mirons-google:


setup/scripts/build_docker.sh
```diff
@@ -20,7 +20,7 @@ if [ -z ${TRAVIS+x} ]; then
     docker -l error rm -f $(docker ps -a -q)
 fi
 
-# Update docker only on Travis.
+# Install docker only on Travis.
 if [ ${TRAVIS+x} ]; then
     # We are on Travis.
     echo "Updating docker to the latest on Travis."
```
joecheuk: Just to make sure, are we updating or installing docker?

mirons-google: Thanks. I clarified it.

PR #1071, opened 2018-02-10
Update all scanners to output violations with inventory_index_id, full_name, inventory_data

blueandgold:
This is for all the remaining scanners.


joecheuk:
Some of the rules have `full_name` as part of the rule definition and some don't, any reason why is that? E.g. INN, IAP, IAM Firewall rules don't have that as part of the rule definition.

PR #1099, opened 2018-02-15
Remove the 1.0 DAO

mirons-google:


google/cloud/forseti/common/data_access/dao.py
```diff
@@ -27,90 +27,12 @@
 from google.cloud.forseti.common.data_access import load_data_sql_provider
```
blueandgold: Can this ```dao.py``` be deleted completely?  Or what else is still using this?

Issue #438, opened 2017-07-06
Purge Database Tables


blueandgold:
Currently, the inventory database tables are not purged.  So, they stay in place, and grows over time.  Which makes it harder to navigate and exceeds the database's capacity.  So, we need to provide a way for users to purge these tables.  Possibly by a config value that users can specify.

gbrindisi:
This is a problem we would like to solve as well :) I'm commenting to remember to get back with ideas later

victorvess:
@blueandgold I am interested in picking this up - my thought is to provide a **retention_days** property in the inventory section of forseti config, that will cause inventory scan to clean up any snapshot tables older than that time period (basically delete tables like '%_cycle_timestamp')

Is there any other dimension than age in days that should be considered?

Also - do we think the corresponding rows within snapshot_cycles should be cleaned up too? I am not sure if there is any kind of audit benefit of leaving those around, I was leaning toward leaving them just in case because it shouldn't take up much space (<9000 rows per year).

This would be my first contribution to Forseti, let me know if you need any additional artifacts

blueandgold:
Hi Victor, thank you for your input here!  Yes, I think that your approach to using a retention_days is more flexible and comprehensive, and will also cover angelo's use case as well.  Please make sure to default it to a state where no data will be purged (perhaps with `-1`).

Keeping the snapshot data makes sense to me, and preferable as it might be helpful for people to have a sense of when forseti has run previously.  As you pointed out, this is very inexpensive to do as well.

Looking forward to your PR, and please ask any questions!

blueandgold:
This also looks good to @angelomellos in PR #1103 .

victorvess:
Already submitted #1129 - per your comment around defaulting, I do this if the directive is missing, but I put a non-zero default in the forseti_conf.yaml.in - would you prefer if I set that to 0 instead of 30 days?

blueandgold:
Fixed by #1129

PR #1103, opened 2018-02-16
clear DBs

angelomellos:
This is for issue https://github.com/GoogleCloudPlatform/forseti-security/issues/438

It lets the user add a flag to drop all DB tables from previous inventory runs before running the current one. Like this:
forseti_inventory -clear_previous

Redone to change the Git email so the CLA will be signed.

### DIFF
--- a/google/cloud/security/inventory/inventory_loader.py
+++ b/google/cloud/security/inventory/inventory_loader.py
@@ -64,6 +64,8 @@ FLAGS = flags.FLAGS
 
 flags.DEFINE_boolean('list_resources', False,
                      'List valid resources for inventory.')
+flags.DEFINE_boolean('clear_previous', False,
+                     'Drop all database tables.')
 
 # Hack to make the test pass due to duplicate flag error here
 # and scanner, enforcer.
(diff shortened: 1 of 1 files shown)


blueandgold:
Thanks for submitting this PR!  But before we accept it, can we have a discussion first for all the possible use cases, so that we have a coherent way for purging the tables?  Since this is the start of a long weekend, will follow-up with you next week.  Thanks again!

angelomellos:
Ok sure, sounds good.

victorvess:
@angelomellos I was starting to look into an approach for this as well, I am wondering instead of deleting every single table in the database, should we allow a retention time interval and base the deletes on a query of **snapshot_cycles** table.  Do you think that would meet your needs?  I was going to start to work my own branch/PR, maybe we can reconcile them?

blueandgold:
@angelomellos, there is a proposal by @victorvess to handle this in a more flexible and comprehensive way in issue #438 .  Can you please take a look?

angelomellos:
Sounds like a good idea to me.

blueandgold:
@angelomellos, PR #1129 has been merged and will cover your use case here, i.e. use ```retention_days = 0```  So, I am going to close out this PR.  Thanks for motivating the work here!

PR #1121, opened 2018-02-17
updated cloudsql_region in the server_config class to have default value

joecheuk:


setup/gcp/installer/configs/server_config.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, **kwargs):
         self.cloudsql_instance = '{}-{}'.format(
             'forseti-security',
             self.datetimestamp)
-        self.cloudsql_region = kwargs.get('cloudsql_region', 'us-central1')
+        self.cloudsql_region = kwargs.get('cloudsql_region') or 'us-central1'
```
blueandgold: After discussion with Joe, we should set these defaults in the flags.  It will be easier to reference and centralized, rather than scattered throughout out the code.

blueandgold: https://stackoverflow.com/a/15301183/2830207

PR #1093, opened 2018-02-14
Support iam policy member of domain type in iam scanner's required mode,

arbeit:
Hello there!

This fixes issue #799. Please take a look and let me know what you think. Thanks!

PR #1123, opened 2018-02-19
port of PR #1093 to the 2.0-dev branch

arbeit:
Hello there!

This is the port of PR #1093 to the `2.0-dev` branch. Please take a look. Thanks!

### DIFF
--- a/google/cloud/forseti/common/gcp_type/iam_policy.py
+++ b/google/cloud/forseti/common/gcp_type/iam_policy.py
@@ -288,6 +288,28 @@ class IamPolicyMember(object):
             member_name = identity_parts[1]
         return cls(identity_parts[0], member_name=member_name)
 
+    def _is_matching_domain(self, other):
+        """Determine whether IAM policy member belongs to domain.
+
+        This applies to a situation where a rule has a `domain` style `members`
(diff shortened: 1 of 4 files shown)


google/cloud/forseti/common/gcp_type/iam_policy.py
```diff
@@ -288,6 +288,28 @@ def create_from(cls, member):
...
+        This applies to a situation where a rule has a `domain` style `members`
+        specification and the policy to check specifies users.
+
+        Args:
+            other (IamPolicyMember): The policy binding member to check.
+
+        Returns:
+            bool: True if `other` is a member of the domain, False otherwise.
+        """
+        if self.type != 'domain' or other.type != 'user':
```
blueandgold: Should this be `other.type != 'domain'`, instead of self.type?

arbeit: The rule member (`self`) has to be of type `domain` and the policy member (`other`) of type `user` in order for this to work.
IOW, the code is correct. Please take another look at `test_policy_all_projects_must_have_owners_from_domain_type` for an example.

Issue #1025, opened 2018-01-31
Revisit run_forseti.sh


mirons-google:
This [script](https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/scripts/gcp_setup/bash_scripts/run_forseti.sh) is used on a successful installation using the wizard and deployment manager. It looks incomplete and/or could be updated.

joecheuk:
closed: #1134

PR #1134, opened 2018-02-21
Updated run_forseti.sh

joecheuk:
Issue: #1025


mirons-google:
Thanks for improving this!

You can do a `set -x` early on in the script, then it will log the commands it's running to `stdout` (console in this case). That way you won't have to do the `echo running command x -y=a`.

I'd also switch to using `logger` instead of `echo` that way the messages get logged to syslog/messages and can be picked up by stackdriver (agent).

PR #1137, opened 2018-02-21
Using set -x to print commands used in the script.

joecheuk:


setup/gcp/scripts/run_forseti.sh
```diff
@@ -15,6 +15,8 @@
 
 source /home/ubuntu/forseti_env.sh
 
+set -x
```
mirons-google: Please add a comment here that because of this setting we should not put anything private/secret in the commands called here because they will be logged.

joecheuk: Comment added, does it look ok to you?

PR #1138, opened 2018-02-22
Violation hashing

### LINKED ISSUE #1059: Add Hash to Uniquely Identify Violations (i.e. prevent duplication from multipe scanning)
IOW, we should not create the violation again on new scanning, if the violation has already been created from previous scanning.

A hash of  resource data, violation data, full_name would serve the purpose here.

mirons-google:
This implements hashing of each violation and stores it along side. Thus, allowing for de-duplication of violations for other systems. It resolves #1059  

---


google/cloud/forseti/services/scanner/dao.py
```diff
@@ -190,3 +199,34 @@ def map_by_resource(violation_rows):
...
+        algorithm (str): The algorithm used to create the hash.
+
+    Returns:
+        str: The resulting hex digest or '' if we can't successfully create
+        a hash.
+    """
+    try:
+        violation_hash = hashlib.new(algorithm)
+    except ValueError as e:
+        LOGGER.error('Unable to use the specified hash algorithm: %s',
```
blueandgold: It's hard to see where someone can specify a different algorithm, without changing the code.
So, it seems this hashing would run twice on error, and fail twice.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -190,3 +199,34 @@ def map_by_resource(violation_rows):
             v_by_type[v_resource].append(v_data)
 
     return dict(v_by_type)
+
+def _create_violation_hash(violation, inventory_index_id,
```
blueandgold: Per description of issue #1059, can you use ```inventory_data, violation data, full_name``` to create the hash?  This was something that andrew, joe, and i `hashed out` to create the uniqueness that we want.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -190,3 +199,34 @@ def map_by_resource(violation_rows):
...
+        a hash.
+    """
+    try:
+        violation_hash = hashlib.new(algorithm)
+    except ValueError as e:
+        LOGGER.error('Unable to use the specified hash algorithm: %s',
+                     algorithm)
+        LOGGER.info('Using the fallback hash algorithm: %s',
+                    FALLBACK_HASH_ALGORITHM)
+        violation_hash = hashlib.new(FALLBACK_HASH_ALGORITHM)
```
blueandgold: What happens if this hashing fails?

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -53,16 +55,17 @@ class Violation(base):
...
         resource_id = Column(String(256), nullable=False)
         resource_type = Column(String(256), nullable=False)
-        full_name = Column(String(1024))
         rule_name = Column(String(256))
         rule_index = Column(Integer, default=0)
-        violation_type = Column(String(256), nullable=False)
         violation_data = Column(Text)
-        inventory_data = Column(Text(16777215))
+        violation_type = Column(String(256), nullable=False)
+        violation_hash = Column(String(256))
```
blueandgold: I think we need to make this column to be unique, if the hash already exists, the insert will fail.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -120,8 +127,10 @@ def create(self, violations, inventory_index_id):
                         violation_type=violation.get('violation_type'),
                         violation_data=json.dumps(
                             violation.get('violation_data')),
-                        inventory_data=violation.get('inventory_data')
+                        inventory_data=violation.get('inventory_data'),
+                        violation_hash=violation_hash
                     )
+
                     session.add(violation)
```
blueandgold: This might to handle if violation can not be added due to hash is not unique.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -53,16 +55,17 @@ class Violation(base):
 
         __tablename__ = violations_tablename
 
+        full_name = Column(String(1024))
```
blueandgold: not a big deal, but it's nicer to see the key column `id` as the first one, and every other column can be sorted after.

google/cloud/forseti/services/scanner/dao.py
mirons-google: The `FALLBACK_HASH_ALGORITHM` is known to work and will be used if the caller specifies an algorithm that fails.

google/cloud/forseti/services/scanner/dao.py
mirons-google: Done!

google/cloud/forseti/services/scanner/dao.py
mirons-google: As indicated in the method doc it will return an empty string, which is a safe way to fail in this scenario.

google/cloud/forseti/services/scanner/dao.py
mirons-google: Done.

google/cloud/forseti/services/scanner/dao.py
mirons-google: Done.

google/cloud/forseti/services/scanner/dao.py
ahoying: The question is, should we keep a record each time scanner finds a violation, even if it is the same one, and rely on downstream systems to filter out violations they've already seen? Or only store new violations?

google/cloud/forseti/services/scanner/dao.py
ahoying: Where would a caller be expected to pass an algorithm in? This is a private method inside of the module, so it should not have any external callers.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -190,3 +202,39 @@ def map_by_resource(violation_rows):
...
+                     algorithm)
+        LOGGER.info('Using the fallback hash algorithm: %s',
+                    FALLBACK_HASH_ALGORITHM)
+        violation_hash = hashlib.new(FALLBACK_HASH_ALGORITHM)
+
+    try:
+        violation_hash.update(
+            violation_full_name +
+            inventory_data +
+            json.dumps(violation_data)
```
ahoying: This must use sort_keys=True so the order of the output is consistent.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -190,3 +202,39 @@ def map_by_resource(violation_rows):
...
+        LOGGER.error('Unable to use the specified hash algorithm: %s',
+                     algorithm)
+        LOGGER.info('Using the fallback hash algorithm: %s',
+                    FALLBACK_HASH_ALGORITHM)
+        violation_hash = hashlib.new(FALLBACK_HASH_ALGORITHM)
+
+    try:
+        violation_hash.update(
+            violation_full_name +
+            inventory_data +
```
ahoying: We should ensure that inventory_data is output with sort_keys=True as well, where it is sourced in google/cloud/forseti/services/inventory/storage.py, in the from_resource() function.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -190,3 +202,39 @@ def map_by_resource(violation_rows):
...
+            json.dumps(violation_data)
+        )
+    except TypeError as e:
```
ahoying: You will also get a NoneType exception if any of the passed in args are None, which they will be if you use .get() in create() above without setting the default value to an empty string.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -110,6 +113,13 @@ def create(self, violations, inventory_index_id):
             """
             with self.violationmaker() as session:
                 for violation in violations:
+
+                    violation_hash = _create_violation_hash(
+                        violation.get('full_name'),
```
ahoying: You should set the default value if the data is not in the dictionary to a string for the string types and an empty dict for violation_data.

google/cloud/forseti/services/scanner/dao.py
ahoying: Nevermind, you will always get a TypeError here. The rest of the comment is still valid.

google/cloud/forseti/services/scanner/dao.py
mirons-google: Great question. I think allowing downstream readers to build functionality around having multiple violations and filtering unwanted ones allows for innovation on this. I support not enforcing uniqueness on this field.

blueandgold: We have decided that we will allow duplicate violations to be stored when scanners are run and find them, as this will allow downstream users the flexibility to de-dupe them or not, based on their use cases.

google/cloud/forseti/services/scanner/dao.py
mirons-google: Done!

google/cloud/forseti/services/scanner/dao.py
mirons-google: Nice catch, fixed.

google/cloud/forseti/services/scanner/dao.py
mirons-google: Done.

google/cloud/forseti/services/scanner/dao.py
mirons-google: Done.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -190,3 +201,40 @@ def map_by_resource(violation_rows):
             v_by_type[v_resource].append(v_data)
 
     return dict(v_by_type)
+
+def _create_violation_hash(violation_full_name, inventory_data, violation_data):
+    """Create a hash of violation data using the requested algorithm.
```
blueandgold: Should remove the comment about the requested algorithm.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -190,3 +201,40 @@ def map_by_resource(violation_rows):
...
+        a hash.
+    """
+
+    # TODO: Intelligently choose from hashlib.algorithms_guaranteed if our
+    # desired one is not available.
+    algorithm = 'sha512'
+
+    try:
+        violation_hash = hashlib.new(algorithm)
+    except ValueError as e:
```
blueandgold: Do we still need this try/except if we use a default ```sha512``` that should just work?

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -54,15 +55,16 @@ class Violation(base):
         __tablename__ = violations_tablename
 
         id = Column(Integer, primary_key=True)
+        full_name = Column(String(1024))
```
blueandgold: Just a note: will you add a `timestamp` or a `violation_timestamp` column in this PR?

google/cloud/forseti/services/scanner/dao.py
mirons-google: It's not _guaranteed__ to work on all systems as noted in the comment on :L218 this can be improved to parse that output of algorithms_guaranteed and choose intelligently from it.

google/cloud/forseti/services/scanner/dao.py
mirons-google: Let me follow-up with a second PR for this.

google/cloud/forseti/services/scanner/dao.py
mirons-google: done

PR #1140, opened 2018-02-22
Added wait time for enable_apis

joecheuk:


setup/gcp/installer/util/utils.py
```diff
@@ -298,3 +300,23 @@ def sanitize_conf_values(conf_values):
...
+    """
+
+    # VT100 control codes, use to remove the last line.
+    erase_line = '\x1b[2K'
+
+    for i in range(0, loading_time*2):
+        dots = '.' * (i % max_number_of_dots)
+        sys.stdout.write('\r{}{}{}'.format(erase_line, message, dots))
+        sys.stdout.flush()
+        time.sleep(0.5)
```
blueandgold: Why multiply ```loading_time``` by 2, and then sleep for 0.5 here?  Seems simpler to leave loading_time as is, and sleep for 1 second?

PR #1146, opened 2018-02-23
Delete data model after each cron run

### LINKED ISSUE #1144: Clean-up Model Tables after each Forseti Cron Run
IIRC, the data model tables need to be clean-up, even if there are violations from it.  So, this means that the `run_forseti.sh` need to do this clean-up.

If I don't recall this correctly, i.e the data models need to be retained because of violations, then we need to develop a way to clean them up later.

joecheuk:
This resolves #1144


setup/gcp/scripts/run_forseti.sh
```diff
@@ -55,3 +55,10 @@ sleep 10s
 echo "Running Forseti notifier."
 forseti notifier run
 echo "Finished running Forseti notifier."
+sleep 10s
+
+# Clean up the model tables
+# Get model_id
+MODEL_ID=$(forseti model get ${MODEL_NAME} | python model_reader.py)
```
blueandgold: Hmmm, this scrapping of the output seems something that can be improved with either a service method that can give you the model handle if you give it the model name. or the delete method should be able to delete by the model name (since it is a primary key).  Please open an issue or a TODO for future improvement.

google/cloud/forseti/services/cli.py
```diff
@@ -252,7 +252,7 @@ def define_model_parser(parent):
         help='Deletes an entire model')
     delete_model_parser.add_argument(
         'model',
-        help='Model to delete')
+        help='Model to delete, either hash or name')
```
blueandgold: Instead of `hash`, use `handle`, as it is more consistent with usage everywhere else.

google/cloud/forseti/services/dao.py
```diff
@@ -1639,7 +1639,7 @@ def delete(self, model_name):
         if model_name in self.sessionmakers:
             del self.sessionmakers[model_name]
         with self.modelmaker() as session:
-            session.query(Model).filter(Model.handle == model_name).delete()
+            session.query(Model).filter(Model.name == model_name).delete()
```
blueandgold: It's hard to see this, but from the comment change above, how do we handle whether to delete by handle or to delete by name?

tests/services/cli_test.py
```diff
@@ -175,7 +175,7 @@ def tearDown(self):
 
         ('model delete foo',
          CLIENT.model.delete_model,
-         ["foo"],
+         ['da39a3ee5e6b4b0d3255bfef95601890afd80709'],
```
blueandgold: From the comment above, do we have a test case for deleting both by handle and by name?

google/cloud/forseti/services/cli.py
joecheuk: Updated.

google/cloud/forseti/services/dao.py
joecheuk: We are no longer changing it in the dao.py, here is what we are doing:

1. get model by handle or name
2. delete that model by handle, this will be reverted

tests/services/cli_test.py
joecheuk: Same as above, since the get_model by name or handle is already tested, we don't need to re test any methods.

blueandgold: Okay.

blueandgold:
Adding `cla: yes` label, as the PR & commit authors are googlers at the time of this PR was created.

Issue #1148, opened 2018-02-24
Differentiate the Client Service Account from Server Service Account


blueandgold:
Both the client & server have service accounts with the naming of forseti-gcp-reader-<TIMESTAMP>, and it's not clear which one is which.  So, the name ideally should be something like `forseti-client-gcp-reader` and `forseti-server-gcp-reader`.

joecheuk:
Closed: #1194

PR #1153, opened 2018-02-25
renamed installer_type to installation_type, replaced timestamp with …

joecheuk:
#1148


setup/gcp/installer/util/utils.py
```diff
@@ -147,32 +147,32 @@ def format_resource_id(resource_type, resource_id):
...
 
     Returns:
-        str: Service account id
+        str: Service account id.
     """
 
     return full_service_acct_email(
         constants.SERVICE_ACCT_FMT.format(
-            prefix, modifier, timestamp), project_id)
+            prefix, modifier, installation_type), project_id)
```
blueandgold: Thanks for making this change!  Just a qq: will we ever have a need for a ```timestamp``` in the service account name?  e.g. someone installing 2 forseti instances into the same project, and thus need the timestamp to differentiate the 2 instances?  What exactly is the character limitation again?

joecheuk: It terms out that service account name (up to 100 characters) can be duplicated and service account id (up to 30 characters) has to be unique. I have separated them out in the new commit, can you please take a look at the new changes?

setup/gcp/installer/util/utils.py
```diff
@@ -147,32 +147,43 @@ def format_resource_id(resource_type, resource_id):
...
+        project_id (str): Id of the project on GCP.
 
     Returns:
-        str: Service account id
+        str: Service account id.
+        str: Service account name.
     """
 
-    return full_service_acct_email(
+    account_id = full_service_acct_email(
```
blueandgold: Can you please change ```account_id``` to ```email```, since that is what it is.

Also, instead of just calling it by `account_id`, can we have `service_account_id` or `service_acct_id`?  Because `service_account` is a very particular type of account that we always call it like that.  So, when I just see `account` in the code, my first impulse is a user account.  No need to backtrack to update everywhere, just update it here locally, and moving forward.  Thanks!

blueandgold: for this kind of methods, please start with verb: e.g. `create_full_service_acct_email()`

setup/gcp/installer/util/utils.py
```diff
@@ -147,32 +147,43 @@ def format_resource_id(resource_type, resource_id):
...
         constants.SERVICE_ACCT_FMT.format(
```
blueandgold: This variable name would read much more clearly as `constants.SERVICE_ACCT_NAME_FMT`.  Then, it's clear that we are creating the service_account_email from service_account_name.

setup/gcp/installer/forseti_installer.py
```diff
@@ -79,10 +79,11 @@ def preflight_checks(self):
                                          is_cloudshell)
         self.organization_id = gcloud.lookup_organization(self.project_id)
         gcloud.check_billing_enabled(self.project_id, self.organization_id)
-        self.format_gcp_service_acct_id()
+        gcp_acct_email, gcp_acct_name = self.format_gcp_service_acct_id()
         self.gcp_service_account = gcloud.create_or_reuse_service_acct(
             'gcp_service_account',
-            self.gcp_service_account,
+            gcp_acct_name,
```
blueandgold: Please make these naming consistent: 
`gcp_service_acct_name`, `gcp_service_acct_id`
or
`gcp_service_account_name`, `gcp_service_account_id`

setup/gcp/installer/forseti_server_installer.py
```diff
@@ -55,13 +55,15 @@ def preflight_checks(self):
...
-        self.format_gsuite_service_acct_id()
         self.should_grant_access()
 
         gcloud.enable_apis(self.config.dry_run)
+        gsuite_acct_email, gsuite_acct_name = (
+            self.format_gsuite_service_acct_id())
         self.gsuite_service_account = gcloud.create_or_reuse_service_acct(
             'gsuite_service_account',
-            self.gsuite_service_account,
+            gsuite_acct_name,
```
blueandgold: ditto as above:
`gsuite_service_acct_name`, `gsuite_service_acct_id`
or
`gsuite_service_account_name`, `gsuite_service_account_id`

setup/gcp/installer/util/gcloud.py
```diff
@@ -431,18 +431,23 @@ def choose_project():
     return target_id
 
 
-def create_or_reuse_service_acct(acct_type, acct_id, advanced_mode, dry_run):
+def create_or_reuse_service_acct(acct_type,
```
blueandgold: Can you please tweak each of these names as below?  Actually the id is really the email, right?  Please update if that's the case.  Because it will make this code easier to understand `acct_id[:acct_id.index('@')]`
`service_acct_type`
`service_acct_name`
`service_acct_id` or `service_acct_email`

setup/gcp/installer/forseti_installer.py
joecheuk: Updated.

setup/gcp/installer/forseti_server_installer.py
joecheuk: Updated.

setup/gcp/installer/util/gcloud.py
joecheuk: Updated acct_it to acct_email. Other argument names for function create_or_reuse_service_account should be self explanatory based on the function context. Since we are creating/reusing service account so arguments like acct_type/acct_name should reference to service account even without the `service` word in the variable.

PR #1155, opened 2018-02-26
Rule files default values

### LINKED ISSUE #1077: Default rules should have their `resource_ids` field replaced with the id of the organization that was chosen during the installation
Default rules should have their `resource_ids` field replaced with the id of the organization that was chosen during the installation.

Example:
https://github.com/GoogleCloudPlatform/forseti-security/blob/4cf7423b581a9d544b77fec7784e81e6b02c57f0/rules/bigquery_rules.yaml#L35-L38

### LINKED ISSUE #1143: Do not run forseti on server startup if the scanner rules are not configured.
Currently, `run_forseti.sh` is invoked on VM startup.  This happens even before rule files are properly configured and creating bad violations in the database.  So, until we are able to pre-configure the rule files as part of the installation, we should remove invoking `run_forseti.sh` on server startup.

Less importantly, this will also affect the initial inventory as gsuite setup will not be done yet by the user.

joecheuk:
Resolves  #1143 
Resolves #1077


joecheuk:
Note: There will not be any default rules populated for forwarding rules. The user have to configure the rule themselves.

setup/gcp/installer/forseti_server_installer.py
```diff
@@ -92,6 +92,10 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
                 self.gcp_service_account,
                 self.user_can_grant_roles)
 
+            # Update rule files with default values
+            rule_vals = self.get_rule_default_values()
+            files.update_rule_files(rule_vals, constants.RULES_DIR_PATH)
```
blueandgold: This block of code can be self-documenting (drop the comment), and read better as:
```
default_rule_values = def get_default rule_values()
files.update_rule_files(default_rule_values)
```

rules/forwarding_rules.yaml
```diff
@@ -12,35 +12,35 @@
...
-    ip_address: "198.51.100.99"
-
-  - name: TCP LB for External VPN
-    target: https://www.googleapis.com/compute/v1/projects/THEPROJECT/regions/us-central1/THELB/FWD_RULE_NAME
-    mode: whitelist
-    load_balancing_scheme: EXTERNAL
-    port_range: 4500-4500
-    ip_protocol: TCP
-    ip_address: "198.51.100.35"
+#rules:
```
blueandgold: By commenting all this out, will it break the forwarding rules scanner?

setup/gcp/installer/forseti_server_installer.py
joecheuk: Updated.

rules/forwarding_rules.yaml
joecheuk: Based on the code, it shouldn't break anything. I will deploy and test it after the data model bug is fixed because at this point we are not able to run scanner without data model.

setup/gcp/installer/forseti_server_installer.py
blueandgold: I don't see the updated change.

joecheuk: Just pushed.

PR #1156, opened 2018-02-26
Standardize on string quoting

### LINKED ISSUE #941: Cleaning up quote types, i.e. use the same quote type everywhere.
We are using both single quotes and double quotes around text inconsistently. We should use the same quote type everywhere in a module unless there is a specific case not to, for example 'This isn't right' would be better as "This isn't right" even if the rest of the code using single quotes.

mirons-google:
This resolves #941 by using a pylint module and enforcing its use. I've also cleaned up all existing infractions of the now `''` single quote policy.


google/cloud/forseti/common/util/parser.py
```diff
@@ -59,17 +59,17 @@ def format_timestamp(timestamp_str, datetime_formatter):
         str: The formatted, stringified timestamp.
     """
     try:
-        if '"' in timestamp_str or "'" in timestamp_str:
+        if '\'' in timestamp_str or '\'' in timestamp_str:
```
joecheuk: I think the logic here needs to be updated to `if '"' in timestamp_str or '\'' in timestamp_str` instead

google/cloud/forseti/common/util/parser.py
```diff
@@ -59,17 +59,17 @@ def format_timestamp(timestamp_str, datetime_formatter):
...
             # Make sure the timestamp is not surrounded by any quotes
-            timestamp_str = timestamp_str.replace("'")
-            timestamp_str = timestamp_str.replace('"')
+            timestamp_str = timestamp_str.replace('\'')
```
joecheuk: Same here, one of the statements should be be replacing double quote.

joecheuk: I think this is actually a bug, it should be `timestamp_str.replace('"', '')`

google/cloud/forseti/common/util/parser.py
```diff
@@ -59,17 +59,17 @@ def format_timestamp(timestamp_str, datetime_formatter):
...
+            timestamp_str = timestamp_str.replace('\'')
```
joecheuk: This should be `timestamp_str.replace('\'', '')`

mirons-google: done

google/cloud/forseti/common/util/parser.py
mirons-google: done

google/cloud/forseti/common/util/parser.py
mirons-google: done

PR #1157, opened 2018-02-26
Remove gflag usage in favor of argparse

### LINKED ISSUE #581: Use argparse instead of gflags

mirons-google:
This resolves #581.


google/cloud/forseti/enforcer/enforcer.py
```diff
@@ -37,58 +33,46 @@
...
-                     lower_bound=0, upper_bound=50)
-
-flags.DEFINE_integer('maximum_project_writer_threads', 1,
-                     'The maximum number of projects with active write '
-                     'operations on project firewalls.',
-                     lower_bound=0, upper_bound=50)
-
-# Setup flags
-FLAGS = flags.FLAGS
+ARG_PARSER = argparse.ArgumentParser()
```
joecheuk: Should this block of code goes under `if __name__ == '__main__'`?

mirons-google: Rather than re-architect `enforcer` I went ahead and did argument parsing at the same level (global) as it was done before.

google/cloud/forseti/enforcer/enforcer.py
```diff
@@ -99,17 +83,18 @@ def initialize_batch_enforcer(global_configs, concurrent_threads,
...
+        max_write_threads (str): The maximum number of enforcement threads that
+            can be actively updating project firewalls.
+        max_running_operations (str): The maximum number of write operations per
+            enforcement thread.
+        dry_run (boolean): If True, will simply log what action would have been
+            taken without actually applying any modifications.
 
     Returns:
-      A BatchFirewallEnforcer instance.
+        BatchFirewallEnforcer:w
```
joecheuk: I think there is a typo here `:w`

google/cloud/forseti/enforcer/enforcer.py
```diff
@@ -99,17 +83,18 @@ def initialize_batch_enforcer(global_configs, concurrent_threads,
...
+        max_write_threads (str): The maximum number of enforcement threads that
+            can be actively updating project firewalls.
+        max_running_operations (str): The maximum number of write operations per
+            enforcement thread.
+        dry_run (boolean): If True, will simply log what action would have been
+            taken without actually applying any modifications.
 
     Returns:
-      A BatchFirewallEnforcer instance.
+        class: BatchFirewallEnforcer
```
joecheuk: Just wanted to clarify, shouldn't this be something like `BatchFirewallEnforcer: A BatchFirewallEnforcer instance.`

google/cloud/forseti/enforcer/enforcer.py
joecheuk: Got it, thanks!

google/cloud/forseti/enforcer/enforcer.py
mirons-google: Yup, fixed.

google/cloud/forseti/enforcer/enforcer.py
```diff
@@ -161,7 +115,54 @@ def main(argv):
 
     del argv
```
joecheuk: Does this delete statement affect the values argparser is going to get from the user?

mirons-google: It's unused and was from the begining.

mirons-google: I switched away from apputils here and argv isn't a problem anymore or at least we ignore it because we don't use it now.

PR #1165, opened 2018-02-27
Attempt to standardize, cleanup, and create missing __init__.py's

mirons-google:
This will be needed on the 2.0 website when we start to produce pydoc generated via sphinx.


deployment-templates/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Forseti Deployment Templates."""
+
+try:
```
joecheuk: How do we decide when/where to put this try/catch?

mirons-google: Only on namespaces we've declared in setup.py:

https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/setup.py#L31

joecheuk: It seems like we have this try/catch in deployment-templates/__init__.py and tests/__init__.py but these 2 folders are not declared in the NAMESPACE_PACKAGES?

mirons-google: Apologies, yes we still need to declare namespace when we are in a different root directory: `/google/` versus `/test/`

joecheuk: Thanks for the clarifications! Everything else looks good.

PR #1127, opened 2018-02-20
Separate out into versioned directories

denzel-morris:
This PR is a WIP, few minor things have to be tweaked and taken care of. Will update this PR and description.

PR #1167, opened 2018-02-27
Check for encrypted vars before using them in the build

mirons-google:


mirons-google:
This should resolve the build issues we're seeing in https://github.com/GoogleCloudPlatform/forseti-security/pull/1127.

@denzel-morris This should resolve your issue, please pull it in and see if it does.

travis/jekyll_build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+if  [ -z ${JGT+x} ]; then
+    # We don't have access to the encrypted vars.
+    bundle exec jekyll build
+else:
+    # We do have access to the encrypted vars.
+    JEKYLL_GITHUB_TOKEN=$JGTbundle exec jekyll build
```
joecheuk: There should be a space between $JGT and bundle.

mirons-google: Fixed, thanks.

PR #1168, opened 2018-02-27
GCS bucket IAM scanner

arbeit:
This pull request adds IAM policy support for buckets in the IAM policy
scanner / rules engine.

I apologize for the size of the PR but adding this feature was **a lot** of
work. What does the feature look like? Well, you can specify a forseti
rule for buckets as follows

      - name: only IAM members in my domain have objectViewer role on buckets
        mode: whitelist
        resource:
          - type: bucket
            applies_to: self
            resource_ids:
              - confidential-data
              - super-secret-stuff
        inherit_from_parents: true
        bindings:
          - role: roles/storage.objectViewer
            members:
              - user:*@your.domain

and violations will be detected whether the problematic policy bindings
are attached to the bucket directly or to any of its folder / project /
organization ancestors.

The `IamPolicyScanner` is the place where we have the big picture and a
full view of all policy resources whereas the `IamRulesEngine` operates
on a single resource and the bigger picture is lost.

That's why the `_add_bucket_ancestor_bindings()` function was added to
the former. It looks at a bucket's ancestors, collects all their
*storage relevant* policy bindings and adds them to the bucket's list of
bindings.

IOW, when the `IamRulesEngine` eventually looks at a bucket it actually
sees a complete set of storage relevant bindings originating from the
bucket and/or its ancestors.

Please take a look and let me know what you think.

Fixes issue #1042

P.S.: before you ask :) the `_add_bucket_ancestor_bindings()` function
can be extended/generalized to support projects and folders as well.
The only reason I refrained from doing so is that I wanted to finish
this PR asap so it can be integrated and tested for the NYC event later
this month.


google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,58 @@
...
+    NOTA BENE: this function only handles buckets and bindings relevant to
+    these at present (but can and should be expanded to handle projects and
+    folders going forward).
+
+    Args:
+        policy_data (list): list of (parent resource, iam_policy resource,
+            policy bindings) tuples to find violations in.
+    """
+    storage_iam_roles = frozenset([
+        'roles/storage.admin',
```
blueandgold: These values should not be hard-coded.  We should allow users to define the roles in the rules file.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,58 @@
...
+
+    If we find one more than one binding with the same role name, we need to
+    merge the members.
+
+    NOTA BENE: this function only handles buckets and bindings relevant to
+    these at present (but can and should be expanded to handle projects and
+    folders going forward).
+
+    Args:
+        policy_data (list): list of (parent resource, iam_policy resource,
```
blueandgold: is `parent resource` really the first element of the tuple?

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,58 @@
...
+        'roles/storage.objectAdmin',
+    ])
+    bucket_data = [(r, bs) for (r, _, bs) in policy_data if r.type == 'bucket']
+    for bucket, bindings in bucket_data:
+        ancestor_bindings = [
+            bs for (r, _, bs) in policy_data
+            if r.full_name != bucket.full_name
+            and not bucket.full_name.find(r.full_name)]
+        for anbs in ancestor_bindings:
+            for anb in anbs:
```
blueandgold: These nested for loops is really confusing, the previous list should be named as `all_ancestor_bindings`.
```
for ancestor_bindings in all_ancestor_bindings:
    for ancestor_binding in ancestor_bindings:

File: google/cloud/forseti/scanner/scanners/iam_rules_scanner.py [hunk 4]
```diff
@@ -32,6 +35,58 @@
...
+                try:
+                    [same_role_binding] = [
+                        iamb for iamb in bindings
+                        if iamb.role_name == anb.role_name]
+                except ValueError:
+                    # no bindings with the same 'role_name'
+                    bindings.append(anb)
+                else:
+                    # found a binding with the same 'role_name', merge members
+                    same_role_binding.merge(anb)
```
blueandgold: How does this update the `policy_data` in the outer context of this method?  I don't see where that happens.

google/cloud/forseti/common/gcp_type/bucket.py
```diff
@@ -0,0 +1,62 @@
...
+"""A Bucket Resource.
+
+See: https://cloud.google.com/storage/docs/json_api/v1/
+"""
+
+from google.cloud.forseti.common.gcp_type import resource
+
+
+class BucketLifecycleState(resource.LifecycleState):
+    """Represents the Folder's LifecycleState."""
```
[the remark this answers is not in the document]
arbeit: copy / paste error -- will fix.

google/cloud/forseti/common/gcp_type/iam_policy.py
```diff
@@ -187,7 +191,27 @@ def create_from(cls, binding):
...
+        try:
+            return cls(binding.get('role'), binding.get('members'))
+        except errors.InvalidIamPolicyMemberError:
+            LOGGER.warn(
+                'Invalid IAM policy member: %s.', binding.get('members'))
+            return None
+
+    def merge(self, other):
+        """Add `other` members to mine if the role names are the same.
+
```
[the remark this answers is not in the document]
arbeit: Yes, merging members from the ancestors is the use case.

google/cloud/forseti/common/gcp_type/iam_policy.py
```diff
@@ -187,7 +191,27 @@ def create_from(cls, binding):
...
             return binding
-        return cls(binding.get('role'), binding.get('members'))
+        try:
+            return cls(binding.get('role'), binding.get('members'))
+        except errors.InvalidIamPolicyMemberError:
+            LOGGER.warn(
+                'Invalid IAM policy member: %s.', binding.get('members'))
+            return None
+
+    def merge(self, other):
```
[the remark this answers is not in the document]
arbeit: Hmm .. there's nothing else to merge but the members :)

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,58 @@
...
+        policy_data (list): list of (parent resource, iam_policy resource,
+            policy bindings) tuples to find violations in.
+    """
+    storage_iam_roles = frozenset([
+        'roles/storage.admin',
+        'roles/storage.objectViewer',
+        'roles/storage.objectCreator',
+        'roles/storage.objectAdmin',
+    ])
+    bucket_data = [(r, bs) for (r, _, bs) in policy_data if r.type == 'bucket']
```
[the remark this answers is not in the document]
arbeit: Will do.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,58 @@
...
+    for bucket, bindings in bucket_data:
+        ancestor_bindings = [
```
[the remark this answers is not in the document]
arbeit: Will rewrite as suggested.
Re. the `not` in the second `if`: the original code read

    if bucket.full_name.find(resource.full_naname) == 0:

but `pylint` complained ("avoid comparing with zero") and then I rewrote to the expression above :-P

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,58 @@
...
+            bs for (r, _, bs) in policy_data
+            if r.full_name != bucket.full_name
+            and not bucket.full_name.find(r.full_name)]
+        for anbs in ancestor_bindings:
```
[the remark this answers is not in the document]
arbeit: Will do.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,58 @@
...
+            for anb in anbs:
+                if anb.role_name not in storage_iam_roles or anb in bindings:
+                    continue
+                # do we have a binding with the same 'role_name' already?
+                try:
+                    [same_role_binding] = [
```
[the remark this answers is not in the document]
arbeit: OK.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
arbeit: I don't understand.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
arbeit: Yes, indeed.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
arbeit: Sorry for the confusion. Will do.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
arbeit: Remember the other review, where we discussed this once? See https://jeffknupp.com/blog/2012/11/13/is-python-callbyvalue-or-callbyreference-neither/

arbeit: In essence `same_role_binding` is a reference to an element in the `bindings` which is a reference to data in `policy_data`. These are all reference to mutable objects and the latter can thus be changed via any one of them.

arbeit:
I believe I have addressed all the review comments above. Thank you for looking at this PR!

google/cloud/forseti/common/gcp_type/iam_policy.py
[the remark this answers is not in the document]
blueandgold: Then why do we need this comment to clarify where this is used?  That's why I suggested you can make the code self-documenting and drop the comment.  Otherwise, we could be merging roles (again at the point where this is used).
```
bucket_binding.merge_members(ancestor_binding)
```
image

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
blueandgold: I am not seeing this is really a `parent resource`.  This ```policy_data``` comes directly from `_retrieve()` and it looks like _a `resource` with parent information inside.  But the ```data```attribute is that of the current resource, not parent.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
blueandgold: There could be other roles that can grant access to gcs buckets: dataflow, dataproc, iam, etc.  So, by putting this in the config file, at least the user can see what roles we are enforcing on, and they can tweak if necessary.

Or, by hard-coding here, are we okay to basically only support scanning on the storage role?  And how do we let users know?

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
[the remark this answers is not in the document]
blueandgold: I don't see the change.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,61 @@
...
+    """
+    storage_iam_roles = frozenset([
+        'roles/storage.admin',
+        'roles/storage.objectViewer',
+        'roles/storage.objectCreator',
+        'roles/storage.objectAdmin',
+    ])
+    bucket_data = [(r, bs) for (r, _, bs) in policy_data if r.type == 'bucket']
+    for bucket, bucket_bindings in bucket_data:
+        all_ancestor_bindings = [
```
blueandgold: Expand this list construction out, please.  the 0 check can be rewritten.  Can you try this?

```
all_ancestor_bindings = []
for (resource, _, bindings) in policy_data:
    if resource.full_name == bucket.full_name:
        continue
    if resource.full_name not in bucket.full_name:
        continue
    all_ancestor_bindings.append(bindings)
```

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -32,6 +35,61 @@
...
+            bindings for (resource, _, bindings) in policy_data
+            # is the resource's full name a proper prefix of the bucket's full
+            # name?
+            if resource.full_name != bucket.full_name
+            # pylint: disable=compare-to-zero
+            and bucket.full_name.find(resource.full_name) == 0]
+
+        for ancestor_bindings in all_ancestor_bindings:
+            for ancestor_binding in ancestor_bindings:
+                if (ancestor_binding.role_name not in storage_iam_roles
```
blueandgold: Please break this up as well into 2 different `if`, so that when we are debugging it's easier to see where this is continued, and easier to read.
```
if ancestor_binding.role_name not in storage_iam_roles:
    continue
if ancestor_binding in bucket_bindings:
    continue
```

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
arbeit: Re. `parent resource`: what is a better name? How else should I call these objects?
Re. hard-coding storage roles: I suggest we merge the PR with these and open a separate issue to take care of it
Rewrote the code as suggested above.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
arbeit: Done.

PR #1176, opened 2018-02-28
Add audit_invocation_time to the violations table.

mirons-google:
Today with `2.0` we have one violations table. Scanners operate on a data-model, if the data-model hasn't changed the same violations will be appended to the table. Since we're not enforcing uniqueness on the  `violations` table to support other use-cases a problem arises on how to "group" violations to an occurrence of running scanner.

`audit_invocation_time`: This field is a timestamp that will be consistent across all violations produced in a given invocation of `google.cloud.forseti.scanner.run()`.

Note: This also removed the (apparently) unused `snapshot_timestamp` as well.


mirons-google:
@blueandgold can you suggest a way to do this in `groups_scanner.py` as it doesn't have access to the `base_scanner.BaseScanner` attributes because we never called it's `__init__`.

google/cloud/forseti/scanner/scanners/instance_network_interface_scanner.py
```diff
@@ -102,9 +101,8 @@ def _get_project_policies(self):
         project_policies = (
             project_dao
             .ProjectDao(self.global_configs)
-            .get_project_policies('projects',
-                                  self.
-                                  snapshot_timestamp))
+            .get_project_policies('projects')
```
joecheuk: This scanner is still referencing `project_dao` located at `google/cloud/forseti/common/data_access` and I believe the file has been removed already.

mirons-google: I punt to @blueandgold to advise on this.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -63,6 +63,7 @@ class Violation(base):
         full_name = Column(String(1024))
         inventory_data = Column(Text(16777215))
         inventory_index_id = Column(String(256))
+        invocation_id = Column(DateTime())
```
blueandgold: This is not an `id`.  This should either be `invocation_time` or `invocated_at` to be consistent with what we are using elsewhere, and also to better rerflect the column's DateTime type.

Also, within the context of this table, there is no idea of what `invocation` is. I know that `scanner` may not be future-proof, but we need to find something more descriptive.  Perhaps one possibility is to call this `audit_time`.

google/cloud/forseti/scanner/scanners/instance_network_interface_scanner.py
blueandgold: Good catch guys.  This whole _get_project_policies() is unused.  :)

mirons-google: I've opened: https://github.com/GoogleCloudPlatform/forseti-security/issues/1184

google/cloud/forseti/services/scanner/dao.py
mirons-google: `audit_invocation_time` it is.

mirons-google:
This now also resolves #1184

PR #1177, opened 2018-02-28
Installer clean up

joecheuk:


setup/gcp/installer/forseti_installer.py
```diff
@@ -172,23 +184,19 @@ def wait_until_vm_initialized(self, vm_name):
...
-        installation_type = self.config.installation_type.capitalize()
-        utils.print_banner('{} VM Initialization'.format(installation_type))
+        installation_type = self.config.installation_type
+        utils.print_banner('{} VM Initialization'.format(
+            installation_type.capitalize()))
         _, zone, name = gcloud.get_vm_instance_info(vm_name)
 
-        # VT100 control codes, use to remove the last line
-        erase_line = '\x1b[2K'
+        status_tracker = lambda: gcloud.check_vm_init_status(name, zone)
```
blueandgold: Why do we need a lamda function to call `gcloud.check_vm_init_status(name, zone)` here?

setup/gcp/installer/forseti_installer.py
```diff
@@ -122,15 +121,28 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
...
             self.project_id,
             self.organization_id,
             deployment_tpl_path,
             self.config.installation_type,
-            self.config.datetimestamp,
+            self.config.timestamp,
             self.config.dry_run)
 
-        deployment_completed = not return_code
+        # Status tracker function, will return true if the deployment is done.
```
blueandgold: nit: drop this comment, as our code should be self-documenting enough that we don't need comments to explain what the code is doing (i.e. we only need to document `why` the code is doing something).

setup/gcp/installer/forseti_installer.py
```diff
@@ -122,15 +121,28 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
...
+        status_tracker = (lambda: gcloud.check_deployment_status(
+            deployment_name, constants.DeploymentStatus.DONE))
+        loading_message = ('This may take a few minutes. Waiting '
+                           'for deployment to be completed...')
+        deployment_completed = utils.show_loading(max_loading_time=900,
```
blueandgold: This statement would read better as:
`deployment_completed = utils.is_deployment_completed()`

`show_loading` does not reflect what the method is really doing (i.e. check status and return a bool result)

setup/gcp/installer/forseti_installer.py
```diff
@@ -122,15 +121,28 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
...
             self.organization_id,
             deployment_tpl_path,
             self.config.installation_type,
-            self.config.datetimestamp,
+            self.config.timestamp,
             self.config.dry_run)
 
-        deployment_completed = not return_code
+        # Status tracker function, will return true if the deployment is done.
+        status_tracker = (lambda: gcloud.check_deployment_status(
```
blueandgold: To align with the method that you are wrapping, this naming should be `status_checker`, not tracker.  Should be consistent.

setup/gcp/installer/util/utils.py
```diff
@@ -313,24 +314,34 @@ def sanitize_conf_values(conf_values):
...
-def show_loading(loading_time, message='', max_number_of_dots=15):
+def show_loading(max_loading_time, exit_condition=None,
+                 message='', max_number_of_dots=15):
     """Show loading message, append dots to the end of the message up to
     a certain number of dots and repeat.
 
     Args:
-        loading_time (int): Loading time in seconds.
+        max_loading_time (int): Loading time in seconds.
+        exit_condition (func): Exit condition, a function that returns
```
blueandgold: This would be more understand if the name is `exit_condition_checker`, since a checker is what we are passing in, not a condition.  And this usage would be easier to read:
`
 if exit_condition_checker and exit_condition_checker():
`

joecheuk: Updated.

setup/gcp/installer/forseti_installer.py
joecheuk: Updated.

setup/gcp/installer/forseti_installer.py
joecheuk: Updated.

setup/gcp/installer/forseti_installer.py
joecheuk: Renamed to start_loading instead of show_loading, the method is being used in few places and not only deployment specific.

Issue #1174, opened 2018-02-28
full_name broken for iam_policy resources


arbeit:
Example:

    organization/518379861585/folder/485180267500/folder/944307890704/project/be-p1-196611/bucket/be-1-ext/iam_policy/bucket/be-1-ext/

Please pay attention to the last 3 segments in the `full_name`; the second to last segment (`bucket`) should not be there.

Branch: `2.0-dev`

arbeit:
A first look at the code suggest the bug is likely in `InventoryImporter::_convert_iam_policy()` in conjunction with `to_full_resource_name()` ??

Issue #1106, opened 2018-02-16
FS 2.0 Data Integrity Errors


mirons-google:
In the VM instance logs of a fresh 2.0 installation I see stack traces regarding integrety checking errors on inserts.

``
"Feb 15 23:43:28 forseti-server[11906]: IntegrityError: (_mysql_exceptions.IntegrityError) (1062, "Duplicate entry 'iam_policy/testpolicy-test' for key 'PRIMARY'") [SQL: u'INSERT INTO `1122243423423422_resources` (full_name, type_name, name, type, policy_update_counter, display_name, email, data, parent_type_name) VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)'] [parameters: (('organization/12345/iam_policy/12345/', 'iam_policy/12345', '12345', 'iam_policy', 0, '', '', '{"bindings": [{"role": "organizations/12345/roles/PersonFolderRole", "members": ["user:user@domain.com"]}, {"role": "organizations/12345 ... (16878 characters truncated) ... ``

If you need the full error log let me know.

joecheuk:
Fixed in PR #1102

PR #1185, opened 2018-03-01
Fix the construction of IAM policy 'full_name'.

arbeit:
@jiyunyao: please make sure this does not affect playground or explain.  Thanks!

The construction of a 'full_name' for 'iam_policy' resources (and
possibly others) is broken since 3 segments are appended to the
'full_name' of the parent resulting e.g. in

    organization/111222333/iam_policy/organization/111222333/

This breaks other code (e.g. get_resources_from_full_name()) assuming
that a full_name consists of pairs of segments i.e.

    organization/111222333/iam_policy/111222333/

Fixes issue #1174.

### DIFF
--- a/google/cloud/forseti/services/model/importer/importer.py
+++ b/google/cloud/forseti/services/model/importer/importer.py
@@ -680,7 +680,7 @@ class InventoryImporter(object):
         parent, full_res_name = self._get_parent(iam_policy)
         iam_policy_type_name = to_type_name(
             iam_policy.get_type_class(),
-            '-'.join(parent.type_name.split('/')))
+            ':'.join(parent.type_name.split('/')))
         iam_policy_full_res_name = to_full_resource_name(
             full_res_name,
             iam_policy_type_name)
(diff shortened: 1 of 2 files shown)


google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -679,8 +679,7 @@ def _convert_iam_policy(self, iam_policy):
         """
         parent, full_res_name = self._get_parent(iam_policy)
         iam_policy_type_name = to_type_name(
-            iam_policy.get_type_class(),
-            parent.type_name)
+            iam_policy.get_type_class(), parent.name)
```
jiyunyao: This will cause duplicate error when building the data model:
https://github.com/GoogleCloudPlatform/forseti-security/issues/1106
This will just reverse the previous fix on the above issue:
https://github.com/GoogleCloudPlatform/forseti-security/pull/1102/files#diff-5129d89f01ca9c52224878ef9ffd09e5R675

jiyunyao:
In this fix, assume there is an organization
`organization/111222333`
the iam_policy type_name will look like:
`iam_policy/111222333/`
and a folder
`folder/111222333`
the iam_policy type_name will also look like:
`iam_policy/111222333/`

PR #1191, opened 2018-03-01
Improve created_at field in the violations table

mirons-google:
`created_at` is a uniqueness string for all given violations in a particular run of scanner. I've refactored this to be more explicit and store it as a datetime object: `created_at_datetime`.

As part of this I've laid the groundwork for common timestamp formats.

### DIFF
--- a/google/cloud/forseti/services/scanner/dao.py
+++ b/google/cloud/forseti/services/scanner/dao.py
@@ -20,7 +20,6 @@ import hashlib
 import json
 
 from sqlalchemy import Column
-from sqlalchemy import DateTime
 from sqlalchemy import String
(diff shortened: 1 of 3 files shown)


google/cloud/forseti/services/scanner/dao.py
```diff
@@ -59,7 +71,7 @@ class Violation(base):
         __tablename__ = violations_tablename
 
         id = Column(Integer, primary_key=True)
-        created_at = Column(DateTime())
+        created_at_timestamp = Column(String(256))
```
blueandgold: This name change is much better and consistent with the public gcp documentations!  :+1:

PR #1195, opened 2018-03-02
Findings Notification

blueandgold:
Second try to submit (originally #1151 , but rebase failed).


google/cloud/forseti/notifier/notifier.py
```diff
@@ -55,30 +55,28 @@ def find_pipelines(pipeline_name):
 # pylint: enable=inconsistent-return-statements
 
 
-def convert_created_at_to_timestamp(violations):
-    """Convert violation created_at datetime to timestamp string.
-
+def convert_to_timestamp(violations):
```
joecheuk: I think the naming of this function is a bit confusing

PR #1209, opened 2018-03-06
Added timeout and retry when executing command

joecheuk:


setup/gcp/installer/util/utils.py
```diff
@@ -325,21 +326,36 @@ def extract_timestamp_from_name(instance_name, include_date=False):
...
                             stderr=subprocess.PIPE)
+
+    timer = threading.Timer(timeout_in_second, proc.kill)
+
+    timer.start()
     out, err = proc.communicate()
+    timer.cancel()
+
+    if proc.returncode and number_of_retry > 0:
+        return run_command(cmd_args, number_of_retry - 1)
```
blueandgold: Is there a message to tell the users what retry number this is?  And a message that there is an error and retry is necessary?

And exponential retry?  What do you think?

joecheuk: What about printing a message like 'There is an error executing command CMD_ARGS, retrying...' ? Do you think the user needs to know/ cares about the current number retry?

joecheuk: Do you think exponential retry can be sometimes too long? I think if the command takes more than 10 seconds to run then there might be some issues on that already.

PR #1219, opened 2018-03-07
Added quiet flag to gcloud compute ssh command

joecheuk:


setup/gcp/installer/util/gcloud.py
```diff
@@ -790,7 +790,7 @@ def check_vm_init_status(vm_name, zone):
 
     _, out, _ = utils.run_command(
         ['gcloud', 'compute', 'ssh', vm_name,
-         '--zone', zone, '--command', check_script_executed])
+         '--zone', zone, '--command', check_script_executed, '--quiet'])
```
blueandgold: Please add a comment to explain why the `--quiet` is needed here (and why it's important).  :)

PR #1231, opened 2018-03-08
PEP8 Improvements

mirons-google:
This PR includes cleanup for:
- Shadowing of python built-ins
- Redundant parentheses
- Missing or empty docstring
- List creation could be rewritten by list literal
- Function call can be replaced with set literal
- Dictionary creation could be rewritten by dictionary literal
- Simplify chained comparisons
- Argument passed to function is equal to default parameter value


google/cloud/forseti/common/gcp_api/compute.py
```diff
@@ -304,10 +309,10 @@ def __init__(self, **kwargs):
         super(_ComputeBackendServicesRepository, self).__init__(
             component='backendServices', **kwargs)
 
-class _ComputeDisksRepository(
-        repository_mixins.AggregatedListQueryMixin,
-        repository_mixins.ListQueryMixin,
-        _base_repository.GCPRepository):
+
+class _ComputeDisksRepository(repository_mixins.AggregatedListQueryMixin,
```
blueandgold: Why is this indentation preferred?  From what I can see, the original style is allowed, and should be preferred to make room for long args.

google/cloud/forseti/common/gcp_api/compute.py
```diff
@@ -574,10 +573,9 @@ def list_instances(self, resource, instance_group, region, **kwargs):
             self, resource, verb='listInstances', **kwargs)
 
 
-class _ComputeSubnetworksRepository(
-        repository_mixins.AggregatedListQueryMixin,
-        repository_mixins.ListQueryMixin,
-        _base_repository.GCPRepository):
+class _ComputeSubnetworksRepository(repository_mixins.AggregatedListQueryMixin,
+                                    repository_mixins.ListQueryMixin,
```
blueandgold: This indentation is not aligned.

google/cloud/forseti/common/gcp_type/firewall_rule.py
```diff
@@ -350,8 +349,9 @@ def _validate_direction(self):
...
         """
         if self.direction == 'INGRESS':
-            if (not self._source_ranges and not self._source_tags and not
-                    self.source_service_accounts):
+            if (not self._source_ranges and
+                    not self._source_tags and
+                    not self.source_service_accounts):
                 raise InvalidFirewallRuleError(
                     'Ingress rule missing required field oneof "sourceRanges" '
                     'or "sourceTags" or "sourceServiceAccounts": "%s".' % self)
```
blueandgold: I am really not good with compound conditions.  But based on the message of the raised exception, should the operator inside the `if` before `or` instead of `and`?

google/cloud/forseti/enforcer/gce_firewall_enforcer.py
```diff
@@ -139,13 +139,11 @@ def build_network_url(project, network):
...
     """
     return (u'%(root)s%(api_name)s/%(version)s/projects/%(project)s/global/'
-            'networks/%(network)s') % {
-                'api_name': API_NAME,
-                'network': network,
-                'project': project,
-                'root': API_ROOT,
-                'version': API_VERSION
-            }
+            'networks/%(network)s') % {'api_name': API_NAME,
```
blueandgold: Why this indentation change?  Seems the original indentation would be fine, and leaves more room.

google/cloud/forseti/scanner/audit/forwarding_rule_rules_engine.py
```diff
@@ -209,40 +209,38 @@ def find_match(self, forwarding_rule):
...
         # default cause is no match thus a return False
         if forwarding_rule.ip_protocol == 'ESP':
-            matched = (
-                (ip_matched) and
-                (scheme_matched) and
-                (protocol_matched))
+            matched = (ip_matched and
+                       scheme_matched and
+                       protocol_matched)
         elif forwarding_rule.ip_protocol == 'TCP' \
```
blueandgold: How about getting rid of the backslash continuation here?

google/cloud/forseti/scanner/scanners/bigquery_scanner.py
```diff
@@ -64,15 +64,13 @@ def _flatten_violations(violations):
...
-            violation_data = {}
-            violation_data['dataset_id'] = violation.dataset_id
-            violation_data['full_name'] = violation.full_name
-            violation_data['access_domain'] = violation.domain
-            violation_data['access_user_by_email'] = violation.user_email
-            violation_data['access_special_group'] = violation.special_group
-            violation_data['access_group_by_email'] = violation.group_email
-            violation_data['role'] = violation.role
-            violation_data['view'] = violation.view
+            violation_data = {'dataset_id': violation.dataset_id,
```
blueandgold: Yup, this actually is an improvement!

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
```diff
@@ -170,10 +170,9 @@ def _output_results(self, all_violations, resource_counts):
         if self.scanner_configs.get('output_path'):
             LOGGER.info('Writing violations to csv...')
             output_csv_name = None
-            with csv_writer.write_csv(
-                resource_name=resource_name,
-                data=all_violations,
-                write_header=True) as csv_file:
+            with csv_writer.write_csv(resource_name=resource_name,
```
blueandgold: Ditto as above, why this change?

pylintrc
```diff
@@ -17,7 +17,7 @@
 [MASTER]
 
 # The number of jobs to spawn for linting.
-jobs=8
+jobs=2
```
blueandgold: Why this change?  How is 2 better?  Seems we want more jobs to be done faster?

mirons-google: The timing was no different. I reverted this as it shouldn't be in this PR.

google/cloud/forseti/scanner/scanners/iam_rules_scanner.py
mirons-google: The visual line indent was the issue I believe here: https://www.python.org/dev/peps/pep-0008/#indentation

google/cloud/forseti/scanner/scanners/bigquery_scanner.py
mirons-google: Ack.

google/cloud/forseti/scanner/audit/forwarding_rule_rules_engine.py
mirons-google: Nice catch, fixed based on pep8 guidelines on operators.

google/cloud/forseti/enforcer/gce_firewall_enforcer.py
mirons-google: The visual line indent was the issue I believe here: https://www.python.org/dev/peps/pep-0008/#indentation

google/cloud/forseti/common/gcp_type/firewall_rule.py
mirons-google: The logic here hasn't changed if that's what you're asking just following pep8 guideline:
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

google/cloud/forseti/common/gcp_api/compute.py
mirons-google: Sorry, but I don't see how it's not aligned.

google/cloud/forseti/common/gcp_api/compute.py
mirons-google: https://www.python.org/dev/peps/pep-0008/#indentation

If we do need to accommodate longer arg lists then you can break re-do it but still adhering to the guidelines above.

mirons-google:
@blueandgold PTAL

google/cloud/forseti/common/gcp_api/compute.py
blueandgold: Yes, I did reference that.  But the guideline does allow the original indentation that was here.  So, why is this being changed?  And is it enforceable?  My issue is if we are expecting people to follow this, and make sure it's justified.

```
Yes:

# Aligned with opening delimiter.
foo = long_function_name(var_one, var_two,
                                           var_three, var_four)

# More indentation included to distinguish this from the rest.
def long_function_name(
        var_one, var_two, var_three,
        var_four):
    print(var_one)

# Hanging indents should add a level.
foo = long_function_name(
    var_one, var_two,
    var_three, var_four)
```

google/cloud/forseti/common/gcp_api/compute.py
blueandgold: Okay, no-ops.

google/cloud/forseti/common/gcp_type/firewall_rule.py
blueandgold: Okay, I filed an issue here for mark to take a look.

google/cloud/forseti/common/gcp_api/compute.py
mirons-google: Yes, I hope to integrate pep8 checking (pip install pycodestyle). But we have a couple other issues it's complaining about that aren't fixable and it doesn't support disabling on a line level.

mirons-google:
Sounds good. I opened an [issue](https://github.com/GoogleCloudPlatform/forseti-security/issues/1240) to either integrate pycodestyle in travis or in a suggest git script.

Issue #1206, opened 2018-03-05
Role missing permissions: roles/containeranalysis.notes.attacher


blueandgold:
Seeing some of these outputs when I create a model on newly crawled inventory.
![image](https://user-images.githubusercontent.com/6977544/37003877-5f66aee4-2083-11e8-85d5-4bbe539c529b.png)

Discussion with Joe & IAM indicates this is expected for alpha or experimental APIs.  So, perhaps for better user experience, move this to debug level, and/or an clear explanation why this is happening.  Otherwise, it's kind of scary and will cause confusion.

joecheuk:
No, this is not closed. Will look into it this week.

joecheuk:
Resolved #1238

PR #1238, opened 2018-03-09
Updated model warnings to log to debug level

joecheuk:
Resolves: #1206

### DIFF
--- a/google/cloud/forseti/services/model/importer/importer.py
+++ b/google/cloud/forseti/services/model/importer/importer.py
@@ -92,7 +92,7 @@ class InventoryImporter(object):
 
         Args:
             session (object): Database session.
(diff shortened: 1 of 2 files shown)


google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -92,7 +92,7 @@ def __init__(self,
 
         Args:
             session (object): Database session.
-            model (str): Model name to create.
+            model (Model): Model name to create.
```
blueandgold: Do you still need the `name` in the arg description, if this is not `str` type anymore.?

joecheuk: Updated the docstring.

Issue #1236, opened 2018-03-09
Failed to create model mysql_exceptions.IntegrityError


joecheuk:
IntegrityError: (_mysql_exceptions.IntegrityError) (1062, "Duplicate entry 'cloudsqlinstance/forseti-server-1205' for key 'PRIMARY'") [SQL: u'INSERT INTO `718eadd3b7514023924a0f416883d734_resources` (full_name, type_name, name, type, policy_update_counter, display_name, email, data, parent_type_name) VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)']

joecheuk:
Closed: #1241

PR #1241, opened 2018-03-09
Resource table integrity errors (fixed only for cloud sql instance)

joecheuk:


google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -902,6 +903,11 @@ def _convert_cloudsqlinstance(self, cloudsqlinstance):
         data = cloudsqlinstance.get_data()
         parent, full_res_name, type_name = self._full_resource_name(
             cloudsqlinstance)
+        parent_key = get_key_from_type_name(parent.type_name)
+        resource_identifier = '{}:{}'.format(parent_key,
```
blueandgold: Will we generate this `resource_identifier` elsewhere?  Would it be better to be in the `util` method, instead of the `get_key_from_type_name()`

joecheuk: resource_identifier will not be used elsewhere, this is a one off fix for cloudsql instance.

blueandgold: Got it, thanks!  LGTM

joecheuk:
Resolves: #1236

PR #1251, opened 2018-03-13
Port findings notification to 1.0

blueandgold:
The violation_data will contain inventory data which would make the hash unique.

The original PR is here:
https://github.com/GoogleCloudPlatform/forseti-security/pull/1195

### DIFF
--- a/google/cloud/security/common/data_access/violation_dao.py
+++ b/google/cloud/security/common/data_access/violation_dao.py
@@ -14,6 +14,8 @@
 
 """Provides the data access object (DAO) for Organizations."""
 
+from datetime import datetime
+import hashlib
 import json
 
 from collections import defaultdict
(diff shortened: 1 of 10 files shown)


google/cloud/security/common/data_access/violation_dao.py
```diff
@@ -82,16 +84,21 @@ def insert_violations(self, violations,
...
             raise db_errors.MySQLError(resource_name, e)
 
+        created_at = datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S')
+
         inserted_rows = 0
         violation_errors = []
         for violation in violations:
             violation = self.Violation(
+                violation_hash=_create_violation_hash(
+                    violation['resource_id'], violation['violation_data']),
```
joecheuk: Can resource_id and violation_data uniquely identify the violation? In 2.0 we are also using inventory_data.

google/cloud/security/notifier/notifier.py
```diff
@@ -221,6 +223,11 @@ def main(_):
     for pipeline in pipelines:
         pipeline.run()
 
+    if notifier_configs.get('violation').get('cscc').get('enabled'):
```
joecheuk: Do you think it's possible that the user only update the application and didn't update the configuration? This might throw null pointer exception.

google/cloud/security/common/data_access/violation_dao.py
blueandgold: Yes, because there is some inventory data embedded in the violation_data.  For example:
https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/security/scanner/scanners/cloudsql_rules_scanner.py#L63-L67

google/cloud/security/notifier/notifier.py
blueandgold: Good catch!  Done.

blueandgold:
fixes #1233

PR #1252, opened 2018-03-13
Added enable_console_log flag on the server

### LINKED ISSUE #1181: Add a flag to the server/CLI to turn on/off logs to console

joecheuk:
Resolves #1181


google/cloud/forseti/services/server.py
```diff
@@ -403,6 +405,9 @@ def serve(endpoint,
     # Configuring log level for the application
     logger.set_logger_level_from_config(log_level)
 
+    if enable_console_log:
```
blueandgold: Can you just move this higher up in the method, so that we can capture all the logging to the console?

google/cloud/forseti/common/util/logger.py
```diff
@@ -78,6 +84,18 @@ def set_logger_level(level):
     LOGLEVEL = level
     _map_logger(lambda logger: logger.setLevel(level))
 
+
+def enable_console_log():
+    """Enable console logging."""
```
blueandgold: Can you please clarify this comment to describe this is enabling console logging for existing handlers?

joecheuk: Updated.

google/cloud/forseti/services/server.py
joecheuk: Updated.

PR #1255, opened 2018-03-14
 Setup the client progress streaming for notifier and scanner.

### LINKED ISSUE #1180: In the CLI notifier and scanner should output to console as well as any notification pipelines
forseti scanner run
and
forseti notifier run

Basically show no output. When running from command line it would be nice to have a 'output to screen' option.

joecheuk:
Resolves #1180


google/cloud/forseti/notifier/notifier.py
```diff
@@ -194,5 +199,8 @@ def run(inventory_index_id, service_config=None):
             violations_as_dict,
             notifier_configs.get('violation').get('findings').get('gcs_path'))
 
-    LOGGER.info('Notification complete!')
+    log_message = 'Notification completed!'
+    progress_queue.put(log_message)
+    progress_queue.put(None)
```
blueandgold: why do we put a `None` here?

joecheuk: The None is used as an identifier to identify the end of the queue, once the loop reached None, it will exit the loop and end the generator.

blueandgold: Got it.  Thanks!

Issue #1256, opened 2018-03-14
Scanners shall only write violations to the database


arbeit:
Streamline the following scanners so their `_output_results()` method only stores the results / violations to the database (and does not poke the `notifier` any more):
 * `firewall_rules_scanner.py`
 * `iam_rules_scanner.py`
 * `iap_scanner.py`

Subtask of issue #1060

PR #1260, opened 2018-03-14
Remove scanner notification behavior

arbeit:
Fixes issue #1256 -- the following scanners do not call the notifier any more but merely store results (violations) to the database:
 * `firewall_rules_scanner.py`
 * `iam_rules_scanner.py`
 * `iap_scanner.py`


arbeit:
The sending of iap violations works

image

Issue #1060, opened 2018-02-09
Unify the Notification Between Scanner and Notifier


blueandgold:
Currently, IAM, IAP, and Firewall scanners send out notifications and file exports of violations, along with Notifier.  This needs to be unified, so that customers will have the same experience.

arbeit:
@blueandgold this is my understanding of what needs to be done:

# Summary
 * Various scanner implementations were copy / pasted / adjusted without refactoring the common code that e.g. emits the results
 * the `notifier` configuration section is gratuitously complex and repetitive (e.g. pipelines are defined for each particular scanner as opposed to the `notifier` in general)
 * some scanners have their own logic that emits `CVS` files or emails (e.g. `firewall_rules_scanner`) while others merely write the violations to the database (e.g. `cloudsql_rules_scanner` etc.)

# Proposal

 1. Pipelines should be defined only once (at the top level) in the `notifier` configuration.
 1. Violations (if notifications are enabled for the scanner in question) will be sent via all the pipelines configured.
 1. Resource-level pipeline specifications are invalid. Users may only define whether notifications are to be emitted for the resource / scanner in question.
 1. A single email / `CSV` file is to be sent / uploaded containing all violations found by the scanners for which notifications are enabled
 1. The `notifier::run()` method shall read these violations from the database and emit notifications as configured (how will the `notifier` run be triggered?)

blueandgold:
Fixed by #1260

PR #1269, opened 2018-03-19
[wip] notifier.run() emits a single notification per violation pipeline

arbeit:
The idea is to make `notifier.run()` emit a single notification per violation pipeline with the least amount of code change
 * the `notifier` section in `forseti_conf.yaml` does not change
 * the violations for all enabled resources are added to a single, shared pipeline instance (see [first commit](https://github.com/GoogleCloudPlatform/forseti-security/pull/1269/commits/a516401ca450f9bdeb57c202002b502341563ceb))
 * all (violation) pipelines need to be adjusted so they can handle violations for 1+ resources (see [second commit](https://github.com/GoogleCloudPlatform/forseti-security/pull/1269/commits/f923c1dc8fd0f4dd07a9eb456d4068688db771e6))

This fixes (part of) issue #1060 

TODO: switch all violation pipelines from `json` to `CVS`.

### DIFF
--- a/google/cloud/forseti/notifier/notifier.py
+++ b/google/cloud/forseti/notifier/notifier.py
@@ -166,7 +166,7 @@ def run(inventory_index_id, progress_queue, service_config=None):
         progress_queue.put(log_message)
 
     # build notification pipelines
-    pipelines = []
+    pipelines = dict()
     for resource in notifier_configs['resources']:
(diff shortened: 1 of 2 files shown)


arbeit:
This will not work. As Joe pointed out, the pipeline configuration may be very different in which case we cannot emit a single notification. Please ignore.

PR #1276, opened 2018-03-20
Fixes for Release 1.1.11

blueandgold:
Here are a few changes I needed to make to get 1.1.11 working correctly.  This is only the first pass.  I will test again, and more fixes might be possible.

### DIFF
--- a/google/cloud/security/notifier/notifier.py
+++ b/google/cloud/security/notifier/notifier.py
@@ -152,7 +152,28 @@ def process(message):
             payload.get('email_description'))
         return
 
-# pylint: disable=too-many-branches
+def run_cscc_notification(notifier_configs, violations_as_dict):
+    """Get the first class in the given sub module
+
+    Args:
+        notifier_configs (dict): Notification configuration.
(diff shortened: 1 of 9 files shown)


google/cloud/security/notifier/notifier.py
```diff
@@ -152,7 +152,28 @@ def process(message):
...
+    if not notifier_configs.get('violation').get('cscc').get('enabled'):
+        LOGGER.debug('CSCC notification is not enabled.')
+        return
+    if '{CSCC_BUCKET}' in (
+            notifier_configs.get('violation').get('cscc').get('gcs_path')):
+        LOGGER.debug('CSCC bucket is not configured.')
+        return
+    CsccPipeline().run(
+        violations_as_dict,
+        notifier_configs.get('violation').get('cscc').get('gcs_path'))
```
joecheuk: I think we have the same problem here, if the user didn't update their config file, this will throw NoneType doesn't have attribute exception.

joecheuk: Never-mind, you checked it at the beginning.

PR #1289, opened 2018-03-22
Updated run_forseti.sh to handle long model creation time

joecheuk:


setup/gcp/scripts/run_forseti.sh
```diff
@@ -36,12 +36,33 @@ sudo systemctl restart forseti.service
...
 MODEL_NAME=$(/bin/date -u +%Y%m%dT%H%M%S)
 echo "Running Forseti inventory."
 forseti inventory create --import_as ${MODEL_NAME}
 echo "Finished running Forseti inventory."
-sleep 10s
+sleep 5s
+
+LOOP_COUNT=0
+MAX_LOOP_COUNT=50
+# Wait until the model is initialized
```
blueandgold: Instead of `is initialized`, it would be nice to be more clear that we are waiting for models to be finished building.

setup/gcp/scripts/run_forseti.sh
```diff
@@ -36,12 +36,33 @@ sudo systemctl restart forseti.service
...
+GET_MODEL_STATUS="forseti model get ${MODEL_NAME} | python -c \"import sys, json; print json.load(sys.stdin)['status']\""
+MODEL_STATUS=`eval $GET_MODEL_STATUS`
+while [ "$MODEL_STATUS" == "CREATED" ] && ((LOOP_COUNT <= MAX_LOOP_COUNT));
```
blueandgold: It seems like this continuously polling for the model status can be better.  An alternative would be to make the inventory create to return a status, that includes the id, which can be used to call model create, which will be blocking until it's really done.  The check model status is good, to see if the model is usable or not.

ahoying:
So what we really want is the inventory creation to not return until after the model is created.

Following the code, the inventory starts the import here: 
https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/google/cloud/forseti/services/inventory/inventory.py#L247

Which calls:
https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/google/cloud/forseti/services/inventory/inventory.py#L193

Which calls:
https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/google/cloud/forseti/services/client.py#L201

Which defaults to background=True.

So the change to make is change the call to new_model in inventory.py to include background=False.

ahoying:
Or explicitly set --background=false in the arguments passed in to the cli.

joecheuk:
Hi @ahoying , the background value that gets passed in to the new_model() method is False, but the console output actually stopped after the inventory process. We will need to spend some time investigating why is background=false doesn't block the console output after the inventory creation process.

joecheuk:
Created issue #1296 to investigate why --import_as doesn't block when background flag is off.

setup/gcp/scripts/run_forseti.sh
```diff
@@ -36,12 +36,33 @@ sudo systemctl restart forseti.service
 # Wait until the service is started
 sleep 10s
 
+# Set the output format to json
+forseti config format json
+
 # Run inventory command
 MODEL_NAME=$(/bin/date -u +%Y%m%dT%H%M%S)
 echo "Running Forseti inventory."
 forseti inventory create --import_as ${MODEL_NAME}
```
blueandgold: Can you please add a TODO here, linking it to issue #1296?  Ultimately, we don't want to keep polling.

joecheuk: Updated.

Issue #470, opened 2017-07-14
Violation CSV writer produces Google Sheets-incompatible CSV


matthewg:
The scanner sent me email with violations containing a violations_data dict, as used by several scanners including IAM and IAP. The style of CSV quoting doesn't seem to play nicely with Google Sheets, see https://docs.google.com/spreadsheets/d/1sXZQLw_oVBk3wrHROquDYteyQKXce-pIBntZQWAW23U/edit#gid=1100928223 for an example: violation_data overflows into multiple cells.

PR #1292, opened 2018-03-22
Improved import of CSV files into Google sheets

arbeit:
Changing the CSV delimiter from comma to TAB does the trick (the problematic `violation_data` field has embedded commas).
Tested the import, works nicely, the `violation_data` field is not scattered across multiple columns any more.

Fixes issue #470.


arbeit:
Apparently, TAB delimited CSV files are a perfectly [valid dialect](https://docs.python.org/2/library/csv.html#csv.Dialect).

And, no. when you import a CSV (e.g. into Google sheets) you don't have to specify what the delimiter is. It just works.

arbeit:
According to [wikipedia](https://en.wikipedia.org/wiki/Comma-separated_values): "In addition, the term 'CSV' also denotes some closely related delimiter-separated formats that use different field delimiters. These include tab-separated values and space-separated values."

Issue #1198, opened 2018-03-02
Handle exception when model is not set


blueandgold:
If we try to use the scanner without setting the model, we get an stack trace exception.  Handle this more gracefully
```
ubuntu@forseti-client-2207-vm:~$ forseti scanner run
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.p
y", line 1025, in <module>
    main(sys.argv[1:], ENV_CONFIG)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.p
y", line 1007, in main
    services[config.service](client, config, output, config_env)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.p
y", line 607, in run_scanner
    actions[config.action]()
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.p
y", line 601, in do_run
    result = client.run()
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/clien
t.py", line 62, in wrapper
    raise Exception('API requires model to be set.')
Exception: API requires model to be set.
```

Issue #1280, opened 2018-03-20
Notifier and Scanner should handle errors properly while running in background.


joecheuk:
Client will hang when error happen during the scanning/notifying process.

PR #1295, opened 2018-03-23
Added error handling for scanner and notifier cli

joecheuk:
To ensure that if there is an exception occurred while scanning/notifying, the cli terminates with some error message.
Using scanner without setting the model should now be handled properly, too.

### DIFF
--- a/google/cloud/forseti/services/scanner/service.py
+++ b/google/cloud/forseti/services/scanner/service.py
@@ -82,9 +82,15 @@ class GrpcScanner(scanner_pb2_grpc.ScannerServicer):
         progress_queue = Queue()
 
         model_name = self._get_handle(context)
-        LOGGER.info('Run scanner service with model: %s', model_name)
-        self.service_config.run_in_background(
-            lambda: self._run_scanner(model_name, progress_queue))
+        if not model_name:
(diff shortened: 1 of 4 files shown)


google/cloud/forseti/services/scanner/service.py
```diff
@@ -82,15 +82,36 @@ def Run(self, _, context):
...
             yield scanner_pb2.Progress(message=progress_message)
 
+    def _run_scanner(self, model_name, progress_queue):
+        """Run scanner.
+
+        Args:
+            model_name (str): Model name.
+            progress_queue (Queue): Progress queue.
+        """
+        # pylint: disable=broad-except
```
blueandgold: Move this pylint disable to the same line as the except, so that this is not propagated down through this whole file.  Fix above too.

```
except Exception as e:  # pylint: disable=broad-except
```

joecheuk: Updated.

joecheuk:
Yes, broad-except is the only way to guaranteed the cli terminates. That's how inventory is handling this process, too.

joecheuk:
Resolves #1198 #1280

PR #1342, opened 2018-04-02
Start forseti service and cloudsql service with ubuntu instead of root

### LINKED ISSUE #1328: FS 2.0: Forseti Server should not run as root

joecheuk:
Resolves #1328


joecheuk:
This has been tested with fresh deployment on GCP.

PR #1351, opened 2018-04-03
Add flag to enable debug mode and suppress crawler's errors & warnings with it.

### LINKED ISSUE #1349: ERROR: "The project to be billed has not configured billing."
I did a fresh install of forseti-2-rc1 and in the initial scan that it is doing (which has been running for like 45 minutes), I see a lot of errors recurring (including the previously-mentioned issue with requestor pays buckets).

Currently, the `/tmp/deployment.log` has over 225,000 lines and growing. But when I `sort | uniq -c | sort -rn` on the file I see some errors recurring quite often.

For example, I have this same error on over 30,000 rows of the file currently:

```
{
  "error": {
    "errors": [
      {
        "domain": "global",
        "reason": "accountDisabled",
        "message": "The project to be billed has not configured billing.",
        "locationType": "header",
        "location": "Authorization"
      }
    ],
    "code": 403,
    "message": "The project to be billed has not configured billing."
  }
```

I agree that it's a useful error to know about, but It seems like it's spending too much effort retrying rather than proceeding to the next resource.

blueandgold:
Fixes #1349 

I have tested this functionally.  Inventory is the same.  But unable to verify if errors are saved correctly as there are no errors in our environment.

### DIFF
--- a/google/cloud/forseti/services/server.py
+++ b/google/cloud/forseti/services/server.py
@@ -264,7 +264,7 @@ class ServiceConfig(AbstractServiceConfig):
             global_config (dict): Global configurations
             forseti_db_connect_string (str): Forseti database string
             endpoint (str): server endpoint
-            enable_debug_mode (bool): Enable console logging.
+            enable_debug_mode (bool): Enable debug mode.
         """
 
         super(ServiceConfig, self).__init__()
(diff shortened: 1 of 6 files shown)


google/cloud/forseti/services/server.py
```diff
@@ -382,6 +386,7 @@ def serve(endpoint,
         forseti_config_file_path (str): Path to Forseti configuration file.
         log_level (str): Sets the threshold for Forseti's logger.
         enable_console_log (bool): Enable console logging.
+        enable_debug_mode (bool): Enable console logging.
```
joecheuk: The description is the same as the variable above, can it be more specific?

joecheuk:
Since this is a very CLI specific feature, can we put the flag into the protobuf request object for inventory creation instead? That way we don't need to touch the server to change the output.

blueandgold:
Great comment on the client-side flag....done!

Issue #1324, opened 2018-04-02
Notifier sends potentially obsolete(?) violations


arbeit:
When the `notifier` runs it sends *all* violations of a certain type even though they may be potentially obsolete (added by previous `scanner` runs). 
Should we not somehow mark the violations from the most recent `scanner` run and just notify of these?

Issue #1257, opened 2018-03-14
Scanners shall write summaries of their runs to the database


arbeit:
There is a need for scanner run summary data (start / end timestamps as well as resource counts) to be added to the database. The resource counts are needed for sending out the violations summary email.
Also all the scanners need to be extended to add this data on the completion of a run.

Subtask of issue #1060 
Blocks issue #1256

blueandgold:
Instead of writing this to a database, do you think that the summary data can just be passed to a notifier library for sending out?  Is there a need to store scanner summary for any other purposes, like reporting?

PR #1352, opened 2018-04-04
Only notify of violations from the most recent scanner run

arbeit:
Multiple, subsequent scanner runs may all add rows to the `violations` table. We only want to be notified of the violations found by the **most recent** scanner run, however.

This PR has been fully tested and
 * adds a `scanner_start_time` column to the `violations` table
 * extends the scanner `DAO` logic so that the
   * `scanner_start_time` value is *the same* for all `violations` rows inserted in the course of a scanner run (across all scanners)
   * `list()` method only returns the violations from the most recent scanner run to the notifier `run()` function.
 * extends the scanner `run()` function to initialize the `scanner_start_time` timestamp before actually running the scanners
 
Fixes issue #1324, #1257 (at least in part?), #1397.


google/cloud/forseti/services/scanner/dao.py
```diff
@@ -123,19 +126,20 @@ def create(self, violations, inventory_index_id):
...
                         inventory_index_id=inventory_index_id,
+                        resource_data=violation.get('resource_data'),
                         resource_id=violation.get('resource_id'),
                         resource_type=violation.get('resource_type'),
-                        full_name=violation.get('full_name'),
-                        rule_name=violation.get('rule_name'),
                         rule_index=violation.get('rule_index'),
-                        violation_type=violation.get('violation_type'),
+                        rule_name=violation.get('rule_name'),
+                        scanner_start_time=scanner_start_time,
```
blueandgold: Scanner time should not be used here. 
What should be here is `scanner_index_id`, which references a new `scanner_index` table. 
This is very analogous to the `inventory_index` table and how it's tied to the `gcp_inventory` table, via the `inventory_index_id`.

arbeit: Gotcha. Will rework the branch accordingly.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -150,14 +254,21 @@ def list(self, inventory_index_id=None):
                 list: List of Violation row entry objects.
             """
             with self.violationmaker() as session:
-                if inventory_index_id:
+                scanner_index = (
+                    session.query(ScannerIndex)
+                    .order_by(ScannerIndex.created_at_datetime.desc()).first())
```
blueandgold: A few comments here:
* This `list()` should also support `scanner_index_id` as a parameter.
* If only `scanner_index_id` is given, then find the violations for it.
* If only `inventory_index_id` is given, then find the `scanner_index_id` with the last `SUCCESS` from the `ScannerIndex` table.  Then use the found `scanner_index_id` to find the violations.
* If neither are given, then find the last `scanner_index_id`.
* If both are given... how do you think this should be handled?

tests/services/scanner/scanner_dao_test.py
```diff
@@ -88,7 +108,8 @@ def tearDown(self):
 
     def test_save_violations(self):
         """Test violations can be saved."""
-        saved_violations = populate_db().list()
+        access, _ = populate_db()
```
blueandgold: `access` should be `violation_access` to match what is in the application code.  Please change this everywhere.

tests/services/util/db.py
```diff
@@ -42,6 +46,16 @@ def create_test_engine_with_file(enforce_fks=True):
...
 
 
+def create_test_engine_with_existing_tempfile(tmpfile, enforce_fks=True):
+    """Create a test engine with the supplied `tmpfile`."""
+
+    logging.info('Creating database at %s', tmpfile)
+    engine = create_engine('sqlite:///{}'.format(tmpfile),
+                           sqlite_enforce_fks=enforce_fks,
+                           connect_args={'check_same_thread': False})
+    return engine, tmpfile
```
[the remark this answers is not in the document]
arbeit: Thanks for pointing to this! I was able to undo the changes to util/db.py altogether and improve the tests in the process.

tests/services/scanner/scanner_dao_test.py
arbeit: Done.

google/cloud/forseti/scanner/scanner.py
```diff
@@ -39,6 +76,7 @@ def run(model_name=None, progress_queue=None, service_config=None):
     violation_access = scanner_dao.define_violation(service_config.engine)
     service_config.violation_access = violation_access
 
+    scanner_configs['scanner_index_id'] = init_scanner_index(service_config)
```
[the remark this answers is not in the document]
arbeit: Sorry. I have adjusted the code to stop using the `scanner_configs`.

google/cloud/forseti/services/scanner/dao.py
arbeit: Here's what I did:
 * if both indices are supplied we fillter on `inventory_index_id` *and* `scanner_index_id`
 * if neither index is supplied we return all violations (i.e. no filtering whatsoever)
 * if either index is supplied we filter on that index

That seemed the most logical to me.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -60,15 +161,16 @@ class Violation(base):
         id = Column(Integer, primary_key=True)
         created_at_datetime = Column(DateTime())
         full_name = Column(String(1024))
-        resource_data = Column(Text(16777215))
         inventory_index_id = Column(String(256))
```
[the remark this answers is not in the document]
arbeit: Opened issue #1397 to capture your suggestion. Will prepare a separate PR for this.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -18,25 +18,126 @@
...
+    CREATED = 'CREATED'
+
+
+class ScannerIndex(BASE):
+    """Represents a scanner run."""
+
+    __tablename__ = 'scanner_index'
+
+    id = Column(String(256), primary_key=True)
+    created_at_datetime = Column(DateTime())
```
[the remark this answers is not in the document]
arbeit: Will be addressed in a separate PR, see issue #1397.

google/cloud/forseti/notifier/notifier.py
```diff
@@ -136,17 +136,23 @@ def run(inventory_index_id, progress_queue, service_config=None):
     global_configs = service_config.get_global_config()
     notifier_configs = service_config.get_notifier_config()
 
+    scanner_index_id = None
```
blueandgold: We should also support the use case if someone wants to run the notifier directly on the scanner_index_id.
```
forseti notifier run --scanner_index_id <scanner index id>
```
This okay to do in a follow-up.  Can you please open an issue to track this?

google/cloud/forseti/scanner/scanner.py
```diff
@@ -14,12 +14,70 @@
...
+def _error_message(failed):
+    """Construct error message for failed scanners.
+
+    Args:
+        failed (list): names of scanners that failed
+
+    Returns:
+        str: error message detailing the scanners that failed
+    """
+    return 'Scanner(s) with errors: %s' % ', '.join(failed)
```
blueandgold: We should add bit more message to explain what the error might mean: `Scanner(s) had errors and were unable to finish scaning for violations.`

google/cloud/forseti/scanner/scanner.py
```diff
@@ -39,11 +97,14 @@ def run(model_name=None, progress_queue=None, service_config=None):
     violation_access = scanner_dao.define_violation(service_config.engine)
     service_config.violation_access = violation_access
 
+    scanner_index_id = init_scanner_index(service_config)
     runnable_scanners = scanner_builder.ScannerBuilder(
         global_configs, scanner_configs, service_config, model_name,
         None).build()
 
     # pylint: disable=bare-except
+    succeeded = []
```
blueandgold: Instead of having 2 lists, you can try to have 1 list `run_states`.

Then you can just append `True` on success, or `False` on failure to `run_states`.  Then, you can go ```if all(run_states)```, or ```if any(run_states)``` to evaluate for the actual states.

This logic table should help.
https://stackoverflow.com/a/19389957

See my next comment below, on putting this into a helper method, and simplify `mark_scanner_index_complete()`

google/cloud/forseti/scanner/scanner.py
```diff
@@ -54,8 +115,13 @@ def run(model_name=None, progress_queue=None, service_config=None):
                 scanner.__class__.__name__)
             progress_queue.put(log_message)
             LOGGER.error(log_message, exc_info=True)
+            failed.append(scanner.__class__.__name__)
+        else:
+            succeeded.append(scanner.__class__.__name__)
     # pylint: enable=bare-except
     log_message = 'Scan completed!'
+    mark_scanner_index_complete(
+        service_config, scanner_index_id, succeeded, failed)
```
blueandgold: The logic should determining the run state would be better off in a separate helper method.
```
run_state = determine_run_state(successful_scanners, failed_scanners)
```
This would be much easier to test.  And can simplify the `mark_scanner_index_complete()` method's signature and functionality to a single operation.
```
mark_scanner_index_complete(
    service_config,
    scanner_index_id,
    run_state)
```

google/cloud/forseti/services/scanner/dao.py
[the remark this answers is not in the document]
blueandgold: Why does this have to be in a separate PR?  Having `inventory_index_id` column here is fundamental to the structure of this table that I think should be done in this PR.

google/cloud/forseti/services/scanner/dao.py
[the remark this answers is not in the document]
blueandgold: Please see my other comment about this.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -123,41 +238,55 @@ def create(self, violations, inventory_index_id):
...
-                            inventory_index_id).all()
-                    )
-                return (
-                    session.query(self.TBL_VIOLATIONS).all())
+                query = None
+                if inventory_index_id and scanner_index_id:
+                    query = (
+                        session.query(Violation)
+                        .filter(and_(
+                            Violation.inventory_index_id == inventory_index_id,
```
blueandgold: If both `inventory_index_id` and `scanner_index_id` are given, why does the query need to filter for both?  Wouldn't just the `scanner_index_id` be enough?  I am worried if someone gives inconsistent `inventory_index_id` and `scanner_index_id`.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -123,41 +238,55 @@ def create(self, violations, inventory_index_id):
...
+                            Violation.scanner_index_id == scanner_index_id)))
+                elif not (inventory_index_id or scanner_index_id):
+                    query = session.query(Violation)
```
blueandgold: Should we really support this use case of returning all the violations?

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -123,41 +238,55 @@ def create(self, violations, inventory_index_id):
...
-                if inventory_index_id:
-                    return (
-                        session.query(self.TBL_VIOLATIONS).filter(
-                            self.TBL_VIOLATIONS.inventory_index_id ==
-                            inventory_index_id).all()
-                    )
-                return (
-                    session.query(self.TBL_VIOLATIONS).all())
+                query = None
+                if inventory_index_id and scanner_index_id:
```
blueandgold: All the if/elif/else code here can be simplified if you pull this into a helper method.  `build_list_query`
```
if scanner_index_id:
    return foo_query

if not (inventory_index_id or scanner_index_id):
    return bar_query

return final_possible_query
```

google/cloud/forseti/notifier/notifier.py
arbeit: Done. Issue #1405 .

arbeit:
This is weird -- the build fail was caused by pylint violations in a file that is not touched by this PR. WTH!?

image

arbeit:
Also:

    $ PYTHONPATH=./ pylint --rcfile=./pylintrc ./setup/gcp/run_forseti_installer.py
    Using config file /home/altair/src/forseti-security-via-fork/pylintrc

    ------------------------------------
    Your code has been rated at 10.00/10

google/cloud/forseti/notifier/notifier.py
```diff
@@ -136,70 +138,69 @@ def run(inventory_index_id, progress_queue, service_config=None):
...
-    notifiers = []
-    for resource in notifier_configs['resources']:
-        if violations.get(resource['resource']) is None:
-            log_message = 'Resource \'{}\' has no violations'.format(
-                resource['resource'])
-            progress_queue.put(log_message)
+        scanner_index_id = scanner_dao.get_latest_scanner_id(
+            session, inventory_index_id)
+        if not scanner_index_id:
+            LOGGER.warn('No scanner index found.')
```
blueandgold: The notifier should halt here if there are no valid `scanner_index_id` found.  Otherwise, this would end up sending other violations back that's not relevant to the current forseti run.

blueandgold: The log line needs to be more specific.  Such as,
`No scanner index id found with success or partial success state for inventory_index_id xyz`, where xyz is the actual inventory_index_id that was used.

google/cloud/forseti/notifier/notifiers/findings.py
```diff
@@ -52,7 +52,7 @@ def _transform_to_findings(violations):
                     'resource_type': violation.get('resource_type'),
                     'resource_id': violation.get('resource_id'),
                     'rule_index': violation.get('rule_index'),
-                    'inventory_index_id': violation.get('inventory_index_id'),
+                    'scanner_index_id': violation.get('scanner_index_id'),
```
blueandgold: It's okay to add the `scanner_index_id` here.  But we still should keep the `inventory_index_id` reference because it will help the user to know right away, which inventory the violation is coming from, without having to do another lookup.

google/cloud/forseti/scanner/scanner.py
```diff
@@ -36,27 +93,39 @@ def run(model_name=None, progress_queue=None, service_config=None):
...
+        inventory_index_id = (
+            model_description.get('source_info').get('inventory_index_id'))
+        scanner_index_id = init_scanner_index(session, inventory_index_id)
+        runnable_scanners = scanner_builder.ScannerBuilder(
+            global_configs, scanner_configs, service_config, model_name,
+            None).build()
+
+        # pylint: disable=bare-except
+        succeeded = []
+        failed = []
```
blueandgold: I mentioned this before, instead of having 2 lists here, you can use 1, `run_statuses`.  And then use a simple any(), all() to find the overall status.
Here is a helpful logic table.
https://stackoverflow.com/a/19389957

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -18,150 +18,259 @@
...
+        if inv_index_id:
+            results = (
+                self.session.query(Violation, ScannerIndex)
+                .filter(and_(or_(
+                    ScannerIndex.scanner_status == 'SUCCESS',
+                    ScannerIndex.scanner_status == 'PARTIAL_SUCCESS'),
+                    ScannerIndex.inventory_index_id == inv_index_id))
+                .filter(Violation.scanner_index_id == ScannerIndex.id)
+                .all())
+            return [v for (v, _) in results]
```
blueandgold: We have decided last week to avoid using list comprehensions because our IDE debuggers can not step through them, which becomes a support issue.  So, can you please write this out?
```
violations = []
for v, _ in results:
  violations.append(v)
return violations
```

google/cloud/forseti/scanner/scanners/base_scanner.py
```diff
@@ -102,24 +104,14 @@ def _output_results_to_db(self, violations):
...
-        violation_access.create(violations, inventory_index_id)
-        # TODO: figure out what to do with the errors. For now, just log it.
-        LOGGER.debug('Inserted %s rows with %s errors',
-                     inserted_row_count, len(violation_errors))
-
-        return violation_errors
+        violation_access = self.service_config.violation_access
+        scanner_index_id = scanner_dao.get_latest_scanner_id(
+            violation_access.session, inventory_index_id,
+            index_state=IndexState.RUNNING)
```
blueandgold: This needs to be more specific here.  Could there be other `RUNNING` states due to previous crashes?  i.e. is `RUNNING` always guaranteed to be 1?

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -18,150 +18,259 @@
...
-                    session.query(self.TBL_VIOLATIONS).all())
-
-    base.metadata.create_all(dbengine)
-
-    return ViolationAccess
+    scanner_index = None
+    if not index_state:
+        scanner_index = (
+            session.query(ScannerIndex)
+            .filter(and_(or_(
```
blueandgold: Instead of `or_()`, you can also use the `in_()`, which would make the filter less repetitive and easier to extend in the future, something like below.
```
retrievable_states = [`SUCCESS`, `PARTIAL_SUCCESS`]
ScannerIndex.scanner_status.in_(retrievable_states)
```

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -18,150 +18,259 @@
...
+                violation_data=json.dumps(
+                    violation.get('violation_data')),
+                violation_hash=violation_hash,
+                violation_type=violation.get('violation_type')
+            )
+            self.session.add(violation)
+
+    def list(self, inv_index_id=None, scanner_index_id=None):
+        """List all violations from the db table.
+
```
blueandgold: The documentation for this method deserves to be expanded, since there are some assumptions being made on how it's supposed to be used.  So, in the method doc, we should describe what is being returned, based on what the parameters are.  e.g. if `inventory_index_id` is given, then all the violations from all the scanner runs will be returned, even if `scanner_index_id` is given.  If you need to look for violations from a specific scanner index, then specify the parameter without inventory_index_id.  And if no parameter is given, than every violation will be returned. etc.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -18,150 +18,259 @@
...
+        Args:
+            session (object): session object to work on.
+            message (str): Error message to set.
+        """
+        self.scanner_index_errors = message
+        session.add(self)
+        session.flush()
+
+
+def get_latest_scanner_id(session, inv_index_id, index_state=None):
```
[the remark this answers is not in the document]
arbeit: Done.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -18,150 +18,259 @@
...
+            results = (
+                self.session.query(Violation, ScannerIndex)
+                .filter(and_(or_(
+                    ScannerIndex.scanner_status == 'SUCCESS',
+                    ScannerIndex.scanner_status == 'PARTIAL_SUCCESS'),
+                    ScannerIndex.inventory_index_id == inv_index_id))
+                .filter(Violation.scanner_index_id == ScannerIndex.id)
+                .all())
+            return [v for (v, _) in results]
+        elif scanner_index_id:
```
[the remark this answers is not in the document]
arbeit: Done.

google/cloud/forseti/notifier/notifier.py
arbeit: Done.

google/cloud/forseti/notifier/notifiers/findings.py
arbeit: Done.

google/cloud/forseti/scanner/scanners/base_scanner.py
arbeit: There could be other `RUNNING` scanner index instances but we are getting the latest instance here. IOW, we are fine.

google/cloud/forseti/services/scanner/dao.py
arbeit: Done.

google/cloud/forseti/services/scanner/dao.py
arbeit: Done.

google/cloud/forseti/scanner/scanner.py
arbeit: I am sorry but I don't see how that would work in this case.

blueandgold: So, with just one list, can we do the following?
run_statuses = all([True, True, True])  # SUCCESS
run_statuses = any([True, False, True])  # PARTIAL_SUCCESS

tests/scanner/scanner_test.py
```diff
@@ -13,49 +13,159 @@
...
+        """Tear down method."""
+        super(ScannerRunnerTest, self).tearDown()
+
+    @mock.patch(
+        'google.cloud.forseti.scanner.scanners.iam_rules_scanner.iam_rules_engine', autospec=True)
+    @mock.patch(
+        'google.cloud.forseti.scanner.scanners.bucket_rules_scanner.buckets_rules_engine', autospec=True)
+    @mock.patch(
+        'google.cloud.forseti.services.server.ServiceConfig', autospec=True)
+    def test_with_runnable_scanners(
```
blueandgold: This test name does not quite reflect what is being tested.  The name might be more clear as `test_scanner_index_id_can_be_correctly_initialized()`

google/cloud/forseti/notifier/notifiers/findings.py
```diff
@@ -29,11 +29,22 @@
...
+        """`Findingsnotifier` initializer.
+
+        Args:
+            inv_index_id (str): inventory index ID
+        """
+        self.inv_index_id = inv_index_id
+
     @staticmethod
-    def _transform_to_findings(violations):
+    def _transform_to_findings(inv_index_id, violations):
```
blueandgold: This doesn't need to be a static method anymore, you can use the inv_index_id from self.

google/cloud/forseti/scanner/scanner.py
arbeit: Sure, but I also need the names of the scanners that failed in order to log the error message.

google/cloud/forseti/notifier/notifier.py
```diff
@@ -136,70 +138,73 @@ def run(inventory_index_id, progress_queue, service_config=None):
...
-
-    for retrieved_v in violations:
-        log_message = ('Retrieved {} violations for resource \'{}\''.format(
-            len(violations[retrieved_v]), retrieved_v))
-        LOGGER.info(log_message)
+        scanner_index_id = scanner_dao.get_latest_scanner_index_id(
+            session, inventory_index_id)
+        if not scanner_index_id:
+            LOGGER.error(
+                'No ((partial) success state) scanner index found for '
```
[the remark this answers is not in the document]
arbeit: Done.

tests/scanner/scanner_test.py
arbeit: Done.

google/cloud/forseti/notifier/notifiers/findings.py
arbeit: Done.

tests/notifier/notifier_test.py
```diff
@@ -119,6 +119,44 @@ def test_notifications_for_nonempty_violations(
...
+        ('google.cloud.forseti.notifier.notifiers.email_violations'
+         '.EmailViolations'), autospec=True)
+    @mock.patch(
+        'google.cloud.forseti.notifier.notifiers.gcs_violations.GcsViolations',
+        autospec=True)
+    @mock.patch(
+        'google.cloud.forseti.notifier.notifier.find_notifiers', autospec=True)
+    @mock.patch(
+        'google.cloud.forseti.notifier.notifier.scanner_dao', autospec=True)
+    def test_notifications_nonempty_violations_but_no_scanner_index(
```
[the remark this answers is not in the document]
arbeit: Done.

arbeit:
Re. the `expunge()`: our sessions have `autocommit` enabled i.e. when they are destructed they try to flush all  the "dirty" objects to the database and this results in "operational error" exceptions.

google/cloud/forseti/notifier/notifier.py
```diff
@@ -57,16 +57,18 @@ def find_notifiers(notifier_name):
...
     Args:
+        session (object): session object to work on.
         violations (sqlalchemy_object): List of violations as sqlalchemy
             row/record object with created_at_datetime.
     Returns:
         list: List of violations as sqlalchemy row/record object with
             created_at_datetime converted to timestamp string.
     """
     for violation in violations:
+        session.expunge(violation)
```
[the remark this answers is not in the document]
blueandgold: Thanks for the explanation.  It is just strange that the session needs to be touched here at all.  Perhaps we should make the list() return a non-sqlalchemy objects or turn-off auto-commit?

```Re. the expunge(): our sessions have autocommit enabled i.e. when they are destructed they try to flush all the "dirty" objects to the database and this results in "operational error" exceptions for violation objects whose created_at_datetime properties have been set to a string in the convert_to_timestamp() function.
Why? Because their column type is DateTime and not Text.```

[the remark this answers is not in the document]
Comment (arbeit) [reply google/cloud/forseti/notifier/notifier.py]: Turning off auto-commit would be quite a sweeping change. Make `list()` return non-sqlalchemy objects is probably better.

What's wrong with `expunge()` BTW?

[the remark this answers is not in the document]
Comment (blueandgold) [reply google/cloud/forseti/notifier/notifier.py]: I mentioned that this method's purpose should be to do a simple conversion to timestamp string.  If the session needs to be managed here, it is an unrelated operation, and perhaps likely that sqlalchemy data type should not be used here.

File: tests/services/scanner/scanner_dao_test.py [hunk 29]
```diff
@@ -60,21 +63,55 @@
...
         """Setup method."""
         ForsetiTestCase.setUp(self)
+        self.engine, self.dbfile = create_test_engine_with_file()
+        _session_maker = sessionmaker()
+        self.session = _session_maker(bind=self.engine)
+        storage.initialize(self.engine)
+        scanner_dao.initialize(self.engine)
+        self.session.flush()
+        self.violation_access = scanner_dao.ViolationAccess(self.session)
+        self.iidx_id1, self.iidx_id2, self.iidx_id3 = _setup_inv_indices(
```
blueandgold: Change the `iidx` to `inv_index_id` to be consistent with the application code.  Please fix everywhere.

google/cloud/forseti/notifier/notifier.py
```diff
@@ -136,70 +138,72 @@ def run(inventory_index_id, progress_queue, service_config=None):
...
+                    notifiers.append(chosen_pipeline(
+                        resource['resource'], inventory_index_id,
+                        violation_map[resource['resource']], global_configs,
+                        notifier_configs, notifier['configuration']))
+
+            # Run the notifiers.
+            for notifier in notifiers:
+                notifier.run()
+
+        if (notifier_configs.get('violation') and
```
blueandgold: This should be moved under the `else` block.  i.e. we don't want this to run if there is no valid scanner index id.

google/cloud/forseti/notifier/notifier.py
```diff
@@ -136,70 +138,72 @@ def run(inventory_index_id, progress_queue, service_config=None):
...
+                notifier_configs.get('violation').get('cscc').get('enabled')):
+            cscc_notifier.CsccNotifier('inventory_index_id').run(
```
blueandgold: ```'inventory_index_id'``` should not be a literal, i.e. quoted.

tests/notifier/notifiers/cscc_notifier_test.py
```diff
@@ -21,25 +21,25 @@
 from google.cloud.forseti.notifier.notifiers import cscc_notifier
 from google.cloud.forseti.services.scanner import dao as scanner_dao
 from tests.services.scanner import scanner_dao_test
-from tests.unittest_utils import ForsetiTestCase
 
-class CsccNotifierTest(ForsetiTestCase):
+class CsccNotifierTest(scanner_dao_test.DatabaseTest):
```
blueandgold: The naming here can be improved, as a test should not subclass from another test.  So, you should call this `BaseDbTestCase`.  Also, if this is something common, it should be moved to the `tests.unittest_util` package.

tests/services/scanner/scanner_dao_test.py
```diff
@@ -235,16 +278,248 @@ def test_create_violation_hash_with_inventory_data_not_string(self):
...
+        index2.scanner_status = IndexState.SUCCESS
+        self.session.add(index2)
+        self.session.add(scanner_dao.ScannerIndex.create('aaa'))
+        self.session.flush()
+        self.assertEquals(
+            expected_id, scanner_dao.get_latest_scanner_index_id(self.session, 'aaa'))
+
+    @mock.patch(
+        'google.cloud.forseti.services.scanner.dao.date_time', autospec=True)
+    def test_get_latest_scanner_index_id_with_specified_state(self, mock_date_time):
```
blueandgold: Change `specified` to `failure` as that is very specific to what is being tested here.

tests/notifier/notifier_test.py
```diff
@@ -119,6 +119,44 @@ def test_notifications_for_nonempty_violations(
...
+         '.EmailViolations'), autospec=True)
+    @mock.patch(
+        'google.cloud.forseti.notifier.notifiers.gcs_violations.GcsViolations',
+        autospec=True)
+    @mock.patch(
+        'google.cloud.forseti.notifier.notifier.find_notifiers', autospec=True)
+    @mock.patch(
+        'google.cloud.forseti.notifier.notifier.scanner_dao', autospec=True)
+    def test_notifications_are_not_sent_without_valid_scanner_index_id(
+        self, mock_dao, mock_find_notifiers, mock_gcs_violations_cls, mock_email_violations_cls):
```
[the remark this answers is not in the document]
arbeit: This is actually not quite true; this method utilizes e.g. both the `mock_email_violations` and the `mock_email_violations_cls` variables.

tests/services/scanner/scanner_dao_test.py
arbeit: Done.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -18,150 +18,276 @@
...
 
     Args:
-        dbengine (engine): sqlalchemy database engine
+        session (object): session object to work on.
+        inv_index_id (str): Id of the inventory index.
+        index_state (str): we want the latest `ScannerIndex` with this state
 
     Returns:
-        ViolationAcccess: facade for accessing violations.
+        object: the latest `ScannerIndex` row or `None`
```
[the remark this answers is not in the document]
arbeit: Done.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -18,150 +18,276 @@
...
+
+        Raises:
+            ValueError: if called with both the inventory and the scanner index
+        """
+        if not (inv_index_id or scanner_index_id):
+            return self.session.query(Violation).all()
+
+        if (inv_index_id and scanner_index_id):
+            raise ValueError(
+                'please call list() with the inventory index XOR the scanner '
```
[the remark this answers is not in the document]
arbeit: Done.

google/cloud/forseti/notifier/notifier.py
arbeit: Done.

tests/services/scanner/scanner_dao_test.py
arbeit: Done.

tests/notifier/notifiers/cscc_notifier_test.py
arbeit: That base class is not general purpose but tailored toward scanner testing -- I have thus renamed it but left it in the same place (`tests/services/scanner/scanner_dao_test.py`)

tests/notifier/notifier_test.py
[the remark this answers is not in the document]
blueandgold: Ah, okay.  This is fine.

google/cloud/forseti/notifier/notifier.py
blueandgold: I didn't see this addressed.

google/cloud/forseti/notifier/notifier.py
```diff
@@ -67,8 +67,8 @@ def convert_to_timestamp(violations):
             created_at_datetime converted to timestamp string.
     """
     for violation in violations:
-        violation.created_at_datetime = (
-            violation.created_at_datetime.strftime(
+        violation['created_at_datetime'] = (
```
blueandgold: The Args/Returns docs need to be updated since this not a sqlalchemy_object anymore.

blueandgold: Done.

google/cloud/forseti/notifier/notifier.py
blueandgold: Done.

PR #1365, opened 2018-04-04
Commented out iam custom rules

### LINKED ISSUE #1361: Conflicting Rules for IAM Policies
The 2 custom iam rules conflict with each other.  We should configure this better to avoid the conflict and confusing violations.

https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/rules/iam_rules.yaml#L31-L56

joecheuk:
Resolves #1361

### DIFF
--- a/rules/iam_rules.yaml
+++ b/rules/iam_rules.yaml
@@ -29,28 +29,28 @@ rules:
           - group:*@{DOMAIN}
 
   # custom rules
-  - name: Allow only service accounts to have access
-    mode: whitelist
-    resource:
-      - type: organization
-        applies_to: self_and_children
-        resource_ids:
-          - '*'
-    inherit_from_parents: true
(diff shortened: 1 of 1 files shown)


rules/iam_rules.yaml
```diff
@@ -29,28 +29,28 @@ rules:
...
-      - type: organization
-        applies_to: self_and_children
-        resource_ids:
-          - '*'
-    inherit_from_parents: true
-    bindings:
-      - role: roles/*
-        members:
-          - serviceAccount:*@*.gserviceaccount.com
+  #- name: Allow only service accounts to have access
```
blueandgold: I would like a little more verbiage here to tell the user that these are examples that they can reference.

joecheuk: Updated.

PR #1388, opened 2018-04-09
Fix the IAM Policy Scanner's required mode.

blueandgold:
This makes sure that required mode works correctly for the IAM policy scanner.

All rule members need to be in the IAM policy, in order not to have any violations.

Any additional members beyond the required members are allowed to be in the policy.


blueandgold:
The original code is working as intended.  So, no need for this PR.

PR #1389, opened 2018-04-10
Updated data model to handle duplicate gcp_inventory record

joecheuk:


joecheuk:
Creating another branch off 2.0-dev instead, the commit is not clean in this one.

PR #1390, opened 2018-04-10
Updated importer to handle duplicate gcp_inventory records

### LINKED ISSUE #1381: gcp_inventory table sometimes contains multiple records for the same resource
The same resource is being inserted multiple times to the gcp_inventory table and it is blocking the data model from building.

joecheuk:
Resolves #1381


google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -166,7 +170,7 @@ def run(self):
 
         autoflush = self.session.autoflush
         try:
-            self.session.autoflush = False
+            self.session.autoflush = True
```
blueandgold: question: Why is the change here from manual flush to autoflush?  Is there any side-effect?  Should there be a comment here to document why this is `True` so that people don't change this in the future?

joecheuk: Updated with comment

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -1091,6 +1145,50 @@ def _convert_organization(self, organization):
...
+        Args:
+            type_name (str): The type name (Primary key of the resource table).
+
+        Returns:
+            bool: Whether or not session contains Resource with
+                primary key = type_name.
+        """
+
+        # one_or_none returns None if the query selects no rows.
+        exists = self.session.query(
```
jiyunyao: This needs optimization. Doing a query every time might be costly. Can we use the resource_cache to check duplication instead?

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -1091,6 +1145,50 @@ def _convert_organization(self, organization):
...
+            self.dao.TBL_RESOURCE).filter(
+                self.dao.TBL_RESOURCE.type_name == type_name).one_or_none()
+
+        if exists:
+            LOGGER.warn('Duplicate type_name: %s', type_name)
+            return False
+        return True
+
+    def _is_role_unique(self, role_name):
```
jiyunyao: Same consideration here.

google/cloud/forseti/services/model/importer/importer.py
joecheuk: Updated, do you mind talking another pass at this?

joecheuk:
Thanks @jiyunyao !

PR #1407, opened 2018-04-12
Added user "forseti-db-admin" to the database and switch to using "forseti-db-admin" to connect to the db in the application

### LINKED ISSUE #921: Create MySQL user "forseti_user" via setup wizard for consistency in docs
Setup wizard defaults to using the "root" mysql user. Likely changes would be:

* Create the new user via deployment manager
* Update the generated Forseti config to use forseti_user

requested by @Tabrarian

joecheuk:
Resolves #921

### DIFF
--- a/deployment-templates/cloudsql/cloudsql-user.py
+++ b/deployment-templates/cloudsql/cloudsql-user.py
@@ -1,0 +1,34 @@
+# Copyright 2017 The Forseti Security Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


joecheuk:
GCP documentation on mysql users: https://cloud.google.com/sql/docs/mysql/users
Under section "User privileges" 
>> Users created by using the Google Cloud Platform Console have the same privileges as the root user. When you create a user by using the MySQL Client, you must grant that user privileges with the GRANT statement.

In our case we are creating it through GCP instead of MySQL Client so it will contain same privileges as root user.

PR #1419, opened 2018-04-16
Specified pip version to be 9.0.3

joecheuk:
Without the specification, pip will upgrade to the newest version (10.0.0) and it will break our initialization step due to some ImportError, details about the ImportError can be found here:
> https://github.com/pypa/pip/issues/5240

### DIFF
--- a/deployment-templates/compute-engine/forseti-instance.py
+++ b/deployment-templates/compute-engine/forseti-instance.py
@@ -159,7 +159,7 @@ $USER_HOME/cloud_sql_proxy -instances={}=tcp:{} &
 # Install Forseti Security.
 cd $USER_HOME
 rm -rf *forseti*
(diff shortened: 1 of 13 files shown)


deployment-templates/compute-engine/forseti-instance.py
```diff
@@ -159,7 +159,7 @@ def GenerateConfig(context):
 # Install Forseti Security.
 cd $USER_HOME
 rm -rf *forseti*
-pip install --upgrade pip
```
jackton1: ### Temporary Fix upgrade pip using.
```
curl https://bootstrap.pypa.io/get-pip.py | python3
```
Instead of  `pip install -U pip`

For pip2  `pip2 install --upgrade pip`

PR #1421, opened 2018-04-16
Allow Inventory crawler to run without Org access.

dufton:
Suggested modification for to the inventory crawler to attempt to continue, with a warning saved, if it encounters an ApiAccessError while reading the root Organization.

This allows the scanners to run on all projects that the Forseti service account has access to even if the service account doesn't have access to the Organization. The scanners will not be able to check Org-level policies or GSuite groups if the service account cannot access the Organization.


google/cloud/forseti/services/inventory/base/resources.py
```diff
@@ -429,15 +430,23 @@ class Organization(Resource):
...
 
         Returns:
             Organization: created Organization
         """
-        data = client.fetch_organization(resource_key)
-        return FACTORIES['organization'].create_new(data, root=True)
+        try:
+            data = client.fetch_organization(resource_key)
+            return FACTORIES['organization'].create_new(data, root=True)
+        except api_errors.ApiExecutionError as e:
```
blueandgold: Can we please add a warning log line here, so that this will show up in stackdriver instead of just the database?

ahoying:
This works, but it will not show all resources the user has access to, in some situations. For example, if the service account has access to view a project, but not a parent folder of the project, then the project will never be discovered.

A longer term solution would be to switch from using (projects/folders).list to using folders.search and projects.list with no parent specified, to find all projects and folders the account has access to, if the root organization is not available.

dufton:
Thank you for your review!

Issue #1166, opened 2018-02-27
The entry point for forseti_server doesn't print usage before trying to start


mirons-google:
It should look for required args and if not found print the help usage.

```
Traceback (most recent call last):
  File "/home/user/.virtualenvs/gflags/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/server.py", line 450, in <module>
    main()
  File "/home/user/.virtualenvs/gflags/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/server.py", line 446, in main
    args['forseti_config_file_path'], args['log_level'])
  File "/home/user/.virtualenvs/gflags/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/server.py", line 367, in serve
    raise Exception('No services to start.')
Exception: No services to start.
```

PR #1439, opened 2018-04-19
Make server command line arg processing more robust

arbeit:
The server now makes sure that the user supplies
 * at least one service
 * the config file path (also checks that the specified path points to a file and that the latter is readable)
 * the database connection string

on the command line. Fixes issue #1166.

### DIFF
--- a/google/cloud/forseti/services/server.py
+++ b/google/cloud/forseti/services/server.py
@@ -450,7 +450,6 @@ def serve(endpoint,
             return
 
 
-
 def check_args(args):
     """Make sure the required args are present and valid.
 
--- a/google/cloud/forseti/services/server.py
+++ b/google/cloud/forseti/services/server.py
@@ -450,6 +450,37 @@ def serve(endpoint,
             return
 
 
+
+def check_args(args):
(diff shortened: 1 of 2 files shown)


google/cloud/forseti/services/server.py
```diff
@@ -479,12 +485,39 @@ def main():
         '--enable_console_log',
         action='store_true',
         help='Print log to console.')
+
     args = vars(parser.parse_args())
 
+    if not args['services']:
```
mirons-google: nit: I think moving these to their own _argument_validator() function can keep main() cleaner.

google/cloud/forseti/services/server.py
```diff
@@ -479,12 +485,39 @@ def main():
...
+        sys.stderr.write('ERROR: please specify at least one service.\n\n')
+        parser.print_usage()
+        sys.exit(1)
```
mirons-google: These exit codes seem to be a "magic" number and probably warrant a comment regarding where they're coming from and how to increment and/or not change them in the future.

PR #1440, opened 2018-04-19
Data model handle duplicate data (member & membership)

joecheuk:
Handle same member with different casing (e.g. "abc@gmail.com" vs "ABC@gmail.com").
Added membership cache to make sure the membership we are inserting is unique.


google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -376,13 +385,21 @@ def _store_iam_policy(self, policy):
...
+            db_members = []
             for member in members:
-                member = member.replace(':', '/', 1)
+                member = member.replace(':', '/', 1).lower()
 
                 # We still might hit external users or groups
                 # that we haven't seen in gsuite.
-                if member not in self.member_cache and \
-                        member not in self.member_cache_policies:
+                if member in self.member_cache and member not in db_members:
```
jiyunyao: the if elif here is a little bit confusing. Can you clear the logic here a little bit?

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -341,8 +343,15 @@ def group_name(parent):
                 type=m_type,
                 member_name=name)
 
-        self._membership_cache.append(
-            (group_name(parent), member))
+        parent_group = group_name(parent)
+
+        if parent_group not in self.membership_cache:
+            self.membership_cache[parent_group] = set()
```
jiyunyao: what is the purpose of membership_cache? Also to avoid duplication?

joecheuk: Yes, we can have duplicate calls due to members with different casing. 
e.g. (group/QWE, user/ABC) and (group/qwe, user/abc) will both get appended to the membership list if we don't check for duplicate.

google/cloud/forseti/services/model/importer/importer.py
joecheuk: I agreed it's confusing to have this elif statement. I have updated to have the second if statement to check for both                 `member not in self.member_cache and member not in self.member_cache_policies` instead

google/cloud/forseti/services/model/importer/importer.py
jiyunyao: In this case, I would say rename it. For example, membership_cache -> membership_graph, membership -> membership_items. The name can be more descriptive.

PR #1444, opened 2018-04-21
Updated importer to use 2 sessions

joecheuk:
One session is not enough because within the same session you can only flush/commit when you finished accessing the database. We using yield per as a paging mechanism to query the database since there can be a lot of items so we won't be able to flush/commit until we get all the items back which will result in a giant block of sql statement.

We will use 2 sessions instead, one to read the data from the gcp_inventory table and one for flush/commit in between reading.


google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -239,8 +276,54 @@ def run(self):
...
+            inventory_iterable (Iterable): Inventory iterable.
+            pre_action (func): Action taken before iterating the
+                inventory list.
+            action (func): Action taken during the iteration of
+                the inventory list.
+            post_action (func): Action taken after iterating the
+                inventory list.
+            flush_count (int): Flush every flush_count times.
+            is_resource (bool): If we are dealing with resources.
+        Returns:
```
blueandgold: Add a blank line.

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -239,8 +276,54 @@ def run(self):
...
+            int: Number of item iterated.
+        """
+
+        if pre_action:
+            pre_action()
+        item_counter = 0
+        ret_val = None
```
blueandgold: Instead of using the abbreviated `ret_val`, why not use `result`?

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -239,8 +276,54 @@ def run(self):
...
+        for inventory_data in inventory_iterable:
+            item_counter = item_counter + 1
+            if isinstance(inventory_data, tuple):
+                ret_val = action(*inventory_data)
+            elif not is_resource:
+                ret_val = action(inventory_data)
+            else:
+                ret_val = action(inventory_data, ret_val)
```
blueandgold: When we fall into this `else` block, isn't `ret_val` = None?  Is this the same as the above `elif` block?  `action(inventory_data)`

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -239,8 +276,54 @@ def run(self):
...
+            if not item_counter % flush_count:
+                session.flush()
+        if is_resource:
+            action(None, ret_val)
```
blueandgold: qq: why is action() being called again for?  Why not include this in the `else` block above, which is for processing when `is_resource` is True?

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -239,8 +276,54 @@ def run(self):
...
+    @staticmethod
+    def model_action_wrapper(session,
+                             inventory_iterable,
+                             pre_action,
+                             action,
+                             post_action,
+                             flush_count,
+                             is_resource=False):
+        """Model action wrapper.
+
```
blueandgold: Please add a comment here to describe why this `model_action_wrapper` is needed, i.e. reduce repetitive code usage and checks.

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -239,8 +269,50 @@ def run(self):
             self.model.set_done(item_counter)
         finally:
             self.session.commit()
+            self.session.autocommit = autocommit
             self.session.autoflush = autoflush
 
+    @staticmethod
+    def model_action_wrapper(session,
```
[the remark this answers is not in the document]
joecheuk: Without this wrapper, we will have to copy and paste this if statement across all the for loops

if not item_count % flush_count:
   self.session.flush()

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -193,40 +199,64 @@ def run(self):
...
-                    self._convert_iam_policy(policy)
-                self._store_iam_policy_post()
-
+                self.model_action_wrapper(
+                    self.session,
+                    inventory.iter(gcp_type_list,
+                                   fetch_iam_policy=True),
+                    self._store_iam_policy_pre,
+                    self._store_iam_policy,
+                    None,
```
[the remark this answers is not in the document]
joecheuk: Updated.

PR #1447, opened 2018-04-24
Fix docstring formatting issues

denzel-morris:
To be merged before #1409. This takes care of all docstring formatting issues, bringing them into compliance with Google's Pydoc style guide.

It should be noted for future reference that the `Returns:` block does not support multi-line indentation like the `Params:` block does:
```python
# WRONG
"""
Returns:
    bool: True if successful,
        False otherwise
"""
# RIGHT
"""
Returns:
    bool: True if successful,
    False otherwise
"""
```

The first example will render improperly and cause errors to be thrown.

Furthermore, we must break example return values out into their own blocks. They cannot be included in the `Returns:` block without rendering improperly:
```python
"""
Returns:
    list: A list of project_ids enabled for bigquery.
 
    If there are no project_ids enabled for bigquery an empty list will
    be returned.

An example return value::

    ['project-id',
     'project-id',
     '...']
"""
```

### DIFF
--- a/google/cloud/forseti/common/gcp_api/api_helpers.py
+++ b/google/cloud/forseti/common/gcp_api/api_helpers.py
@@ -31,7 +31,7 @@ def get_delegated_credential(delegated_account, scopes):
 
     Returns:
         service_account.Credentials: Credentials as built by
-            google.oauth2.service_account.
+        google.oauth2.service_account.
     """
     request = requests.Request()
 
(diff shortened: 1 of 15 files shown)


mirons-google:
Thanks @denzel-morris . I updated your PR with a few improvements. Please comment back on this PR when you can confirm that sphinx still builds correctly with my changes.

denzel-morris:
@mirons-google everything checks out, I initially thought that double colons were necessary for a blockquote to be output. Looks like I was incorrect, thanks for the correction.

PR #1449, opened 2018-04-24
Apply rouge highlighting to inline code

denzel-morris:
I've updated the stylesheet to apply highlighting to inline code. Like block-level codeblocks, you must specify the syntax for inline code as well.

This is done in kramdown with an [inline attribute list](https://kramdown.gettalong.org/syntax.html#code-spans) ("IAL"):
```
`example code here`{:.language-python}
```

With this IAL in place, inline code will now pick up any applicable syntax highlighting from the updated stylesheet.

### DIFF
--- a/_sass/rouge/_github_codeblock.scss
+++ b/_sass/rouge/_github_codeblock.scss
@@ -1,209 +1,209 @@
-.highlight table td { padding: 5px; }
-.highlight table pre { margin: 0; }
-.highlight .cm {
+.highlight table td, .highlighter-rouge table td { padding: 5px; }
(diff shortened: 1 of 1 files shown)


mirons-google:
Thanks @denzel-morris. Does this mean the previous formatting of ``````yaml...``` no longer work? If so can you update the existing codeblocks to this new styling format?

denzel-morris:
@mirons, all multiline codeblocks with the triple-quote language format work as expected and intended.

This only relates to single-quote inline code tags.

Just to confirm, it's the single-quote inline tags you would like for me to update to specify the language?

mirons-google:
Please, but let's do it in a follow-up PR. This works for now, thanks.

PR #1460, opened 2018-04-26
Added encoding to make the MySQL String column case sensitive

joecheuk:


google/cloud/forseti/services/dao.py
```diff
@@ -258,51 +258,174 @@ def define_model(model_name, dbengine, model_seed):
...
+
+    else:
+        class Resource(base):
+            """Row entry for a GCP resource."""
+            __tablename__ = resources_tablename
+
+            full_name = Column(String(2048), nullable=False)
+            # The collation type is used here to allow the String column
+            # to be case sensitive.
+            type_name = Column(String(512, collation='utf8mb4_bin'),
```
blueandgold: Would something like this work?
```
type_name = Column(get_string(), primary_key=True)
```
And it is inside `get_string()` that you have the logic to determine whether a `String()` is returned with the collation attribute or not?

Might also want to be able to passing in a parameter to determine the size, eg. 512

google/cloud/forseti/services/dao.py
```diff
@@ -259,22 +259,40 @@ def define_model(model_name, dbengine, model_seed):
...
+
+        Returns:
+            String: Sqlalchemy String.
+        """
+        if db_dialect.lower() == 'sqlite':
+            return String(column_size)
+        else:
+            return String(column_size, collation='utf8mb4_bin')
+
+    # Sqlite doesn't support collation type, need to define different
```
blueandgold: This comment should be inside the get_string_by_dialect().  Please also add a comment that this is to support case sensitivity.

PR #1463, opened 2018-04-26
Port retention_days to 2.0

blueandgold:
This is a port of PR #1129.  This works on the CLI for now.  Will follow-up to make it usable by the cron job.

### DIFF
--- a/google/cloud/forseti/services/inventory/inventory.py
+++ b/google/cloud/forseti/services/inventory/inventory.py
@@ -330,7 +330,7 @@ class Inventory(object):
         cutoff_datetime = (
             utc_now - datetime.timedelta(days=retention_days))
         LOGGER.info('Cut-off datetime to start purging is: %s',
-                     cutoff_datetime)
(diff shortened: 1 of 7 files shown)


google/cloud/forseti/services/inventory/inventory.py
```diff
@@ -303,3 +309,49 @@ def delete(self, inventory_id):
...
+    def purge(self, retention_days):
+        """Purge the gcp_inventory data that's older than the retention days.
+
+        Args:
+            retention_days (int): Days of inventory tables to retain.
+
+        Returns:
+            str: Purge result.
+        """
+        LOGGER.debug('retention_days is: %s', retention_days)
```
joecheuk: Is there any reason why these logs are DEBUG instead of INFO? Usually clients don't turn no the  debug flag and might be hard to debug if they found some of the data missing.

blueandgold:
Fixes #1347

Issue #1326, opened 2018-04-02
explainer list_permissions should print an error as opposed to a traceback


arbeit:
When invoked without a permission `explainer list_permissions` prints a Traceback

    (fsec) rules $ forseti --endpoint [::]:50051 explainer list_permissions 
    Traceback (most recent call last):
      File "/home/altair/virtenvs/fsec/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.py", line 1025, in <module>
        main(sys.argv[1:], ENV_CONFIG)
      File "/home/altair/virtenvs/fsec/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.py", line 1007, in main
        services[config.service](client, config, output, config_env)
      File "/home/altair/virtenvs/fsec/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.py", line 829, in run_explainer
        actions[config.action]()
      File "/home/altair/virtenvs/fsec/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.py", line 762, in do_list_permissions
        config.role_prefixes)
      File "/home/altair/virtenvs/fsec/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/client.py", line 61, in wrapper
        return f(*args, **kwargs)
      File "/home/altair/virtenvs/fsec/lib/python2.7/site-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/client.py", line 607, in query_permissions_by_roles
        request, metadata=self.metadata())
      File "/home/altair/virtenvs/fsec/local/lib/python2.7/site-packages/grpc/_channel.py", line 507, in __call__
        return _end_unary_response_blocking(state, call, False, deadline)
      File "/home/altair/virtenvs/fsec/local/lib/python2.7/site-packages/grpc/_channel.py", line 455, in _end_unary_response_blocking
        raise _Rendezvous(state, None, None, deadline)
    grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with (StatusCode.UNKNOWN, Exception calling application: No roles or role prefixes specified)>

The absence of the parameter should be checked and result in a proper error message instead.

PR #1469, opened 2018-04-27
Fix explainer CLI tracebacks

arbeit:
Fixes issue #1326.

### DIFF
--- a/google/cloud/forseti/services/cli.py
+++ b/google/cloud/forseti/services/cli.py
@@ -32,12 +32,13 @@ class DefaultParser(ArgumentParser):
     error message, it will print the help message (-h).
     """
 
-    def error(self, _):
+    def error(self, message):
         """This method will be triggered when error occurred.
 
         Args:
             _ (str): Error message.
         """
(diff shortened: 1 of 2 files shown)


google/cloud/forseti/services/cli.py
```diff
@@ -1006,7 +1020,10 @@ def main(args,
     if not services:
         services = SERVICES
     output = outputs[config.out_format]()
-    services[config.service](client, config, output, config_env)
+    try:
+        services[config.service](client, config, output, config_env)
+    except ValueError as e:
+        parser.error(e.message)
```
joecheuk: Do you think it's better to centralize the error checking logic inside the parser? So instead of having the check in each of the explain methods, we only need to check them to the parser.

arbeit: This is the case already apart from the 2 notable exceptions I am fixing in this PR.
`argparse` does not support the following semantics: "I want at least one of the following args/options"
That's why such a case needs to be checked manually.

PR #1479, opened 2018-04-30
Make the inventory purge runnable from the cron job.

blueandgold:
The cron job will default to use the `retention_days` from the configuration yaml file.

This also fixes a bug where the `AbstractInventoryConfig(dict)`, where the inherited dict makes it so that the config values can not be set.


google/cloud/forseti/services/cli.py
```diff
@@ -87,7 +87,11 @@ def define_inventory_parser(parent):
         help='Purge all inventory data older than the retention days.')
     purge_inventory_parser.add_argument(
         'retention_days',
-        help='Number of days to retain the data.')
+        default=None,
```
joecheuk: Is the default value None necessary? I think if you put a default value = None here it might get converted to a string "None" which will cause problem in the if statement since (not "None") is false.

google/cloud/forseti/services/server.py
```diff
@@ -112,7 +112,7 @@ def get_storage_class(self):
         raise NotImplementedError()
 
 
-class AbstractInventoryConfig(dict):
+class AbstractInventoryConfig(object):
```
joecheuk: Will this break other part of the code? if it's not a dictionary you can no longer access using the .get(key) method.

google/cloud/forseti/services/cli.py
blueandgold: This will stays None (i.e. null) even when stuffed into proto string, so this will be handled correctly.

google/cloud/forseti/services/server.py
blueandgold: Reverted.  Will fix this in another PR.

google/cloud/forseti/services/cli.py
blueandgold: It's not really needed... but I like that's explicitly set to None here.

blueandgold:
Fixes #1347

PR #1480, opened 2018-04-30
Installer updates (Remove requirement for exporting service account for DWD)

### LINKED ISSUE #1399: Remove requirement for exporting service account for DWD
The google.auth library can support signing a service account jwt with a delegated subject using the GCP IAM signBlob API (https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts/signBlob), without requiring an exported key.

Here is a detailed setup example: https://gist.github.com/jonparrott/e465797e1071a1402d35bbb4e2370cbc.

The admin API calls should switch to using this model to set up the credentials, so Forseti will only need a single service account for the Inventory, which can handle crawling the GCP and admin APIs.

DWD will still need to be configured for the service account, but the key will no longer need to be exported and stored on disk.

joecheuk:
Resolves #1399


setup/gcp/installer/util/constants.py
```diff
@@ -104,7 +104,8 @@ class DeploymentStatus(Enum):
     'roles/storage.objectViewer',
     'roles/storage.objectCreator',
     'roles/cloudsql.client',
-    'roles/logging.logWriter'
+    'roles/logging.logWriter',
+    'roles/iam.serviceAccountTokenCreator'
```
ahoying: I'd like to see if this can just be set on the service account instead of the project. Using the SVC_ACCT_ROLES (which should otherwise be deleted as I don't think anything else is using it.)

PR #1484, opened 2018-05-01
bug fix - wrong indentation so the member is not populating properly

joecheuk:


tests/services/inventory/test_data/mock_gcp_results.py
```diff
@@ -428,6 +428,26 @@
 }}
 """
 
+CRM_PROJECT_IAM_POLICY_MEMBER_MULTI_ROLES = """
+{
+ "version": 1,
+ "bindings": [
+  {
+   "role": "roles/appengine.codeViewer",
```
blueandgold: Actually, I like the previous test data better.  Where there is another role, that does not contain the abc_user.  Can you please add that, just to make sure this test is robust?

tests/services/model/importer/importer_test.py
```diff
@@ -102,6 +102,11 @@ def test_inventory_importer_basic(self):
                 self.assertFalse(
                         len(filter(None, policy.full_name.split('/'))) % 2)
 
+        # Make sure binding_members table is populated properly when there are users with multiple
+        # roles in different projects.
+        access_list = data_access.query_access_by_member(session, 'user/abc_user@forseti.test', [])
+        self.assertEqual(2, len(access_list))
```
blueandgold: nit: purely for style and clarity, can you please change `access_list` to `abc_user_access`.  Then the code will read much better.

tests/services/model/importer/importer_test.py
```diff
@@ -102,6 +102,11 @@ def test_inventory_importer_basic(self):
...
+
```
blueandgold: In addition to just checking on the number of elements, can you please also assert on the specific roles?  In addition to my comment above, this will really make sure that the member & roles are correct.

PR #1485, opened 2018-05-02
Updated gce host and docker to use ubuntu image version 18.04, added google cloud sdk installation in the startup script

### LINKED ISSUE #1474: Research and update the ubuntu image to the newest major release
We are currently using

    image-project: ubuntu-os-cloud
    image-family: ubuntu-1604-lts 

And it applies to both server and client VMs

joecheuk:
This is the newest Ubuntu image supported by GCP.
Details about all the supported images: https://cloud.google.com/compute/docs/images

The version of OpenSSH installed 
```
joe@forseti-server-vm-xxxx:~$ ssh -v localhost
OpenSSH_7.6p1 Ubuntu-4, OpenSSL 1.0.2n  7 Dec 2017
```

Reason why we need to install google cloud sdk while running the startup script:
>When the startup script of the VM runs, it doesn't have google cloud sdk library installed (e.g. gsutil is not available) that's why we need to add this line here to install the google cloud sdk library before the rest of the installation beings.

Resolves #1474

### DIFF
--- a/deployment-templates/compute-engine/server/forseti-instance-server.py
+++ b/deployment-templates/compute-engine/server/forseti-instance-server.py
@@ -119,6 +119,7 @@ exec 2>&1
 # Ubuntu update.
 sudo apt-get update -y
 sudo apt-get upgrade -y
+sudo apt-get update && sudo apt-get --assume-yes install google-cloud-sdk
 
 USER_HOME=/home/ubuntu
 
--- a/deployment-templates/deploy-forseti-client.yaml.in
+++ b/deployment-templates/deploy-forseti-client.yaml.in
@@ -36,7 +36,7 @@ resources:
   properties:
     # GCE instance properties
     image-project: ubuntu-os-cloud
-    image-family: ubuntu-1604-lts
+    image-family: ubuntu-1804-lts
     instance-type: n1-standard-2
     zone: {FORSETI_SERVER_ZONE}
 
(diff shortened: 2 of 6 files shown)


blueandgold:
qq: Has this been tested on GCP, including some basic functionalities like inventory, data model, scanner, and notification will run without any problem?

joecheuk:
Yes, inventory/explain/model/scanner/notifier all ran without any problem.

deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -119,6 +119,7 @@ def GenerateConfig(context):
 # Ubuntu update.
 sudo apt-get update -y
 sudo apt-get upgrade -y
+sudo apt-get update && sudo apt-get --assume-yes install google-cloud-sdk
```
blueandgold: Can we separate this change into another PR, since it's not directly related to switching to the image version?  Generally speaking, it's good practice to do this separation so that any problems are easier to troubleshoot and find.

blueandgold:
If you don't want to separate into separate PR, at least also clarify the title to include the updating change.

deployment-templates/compute-engine/server/forseti-instance-server.py
joecheuk: This is a problem only occurs in image version 18.04. When the startup script of the VM runs, it doesn't have google cloud sdk library installed (e.g. complaining gsutil command not found) that's why we need to add this line here to install the google cloud sdk library before the rest of the installation beings.

blueandgold: Ah, okay, that makes sense then.  LGTM!

PR #1486, opened 2018-05-02
Added createdAt field in the model response protos to expose the created at datetime

### LINKED ISSUE #1468: forseti model to show creation time
Could you add creation timestamp information for created models?

It would be easier to understand when I created them, especially when I have a few, currently I'm forced to name them with timestamps to remember.

I'd like to get the information when I do `forseti model list`

joecheuk:
Resolves #1468

### DIFF
--- a/google/cloud/forseti/services/model/model.proto
+++ b/google/cloud/forseti/services/model/model.proto
@@ -59,8 +59,9 @@ message ModelSimplified {
   string handle = 1;
   string name = 2;
   string status = 3;
-  string description = 4;
-  string message = 5;
+  string createdAt = 4;
+  string description = 5;
+  string message = 6;
 }
 
 message GetModelRequest {
@@ -71,9 +72,10 @@ message ModelDetails {
   string handle = 1;
   string name = 2;
   string status = 3;
-  string description = 4;
-  string message = 5;
-  string warnings = 6;
+  string createdAt = 4;
+  string description = 5;
+  string message = 6;
+  string warnings = 7;
 }
 
 message PingRequest {
--- a/google/cloud/forseti/services/model/model.proto
+++ b/google/cloud/forseti/services/model/model.proto
@@ -59,9 +59,9 @@ message ModelSimplified {
   string handle = 1;
   string name = 2;
   string status = 3;
-  string createdAt = 4;
-  string description = 5;
-  string message = 6;
+  string description = 4;
+  string message = 5;
+  string createdAt = 6;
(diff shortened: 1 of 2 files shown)


google/cloud/forseti/services/model/model.proto
```diff
@@ -59,8 +59,9 @@ message ModelSimplified {
   string handle = 1;
   string name = 2;
   string status = 3;
-  string description = 4;
-  string message = 5;
+  string createdAt = 4;
```
blueandgold: Is it okay the change the id here?  IIUC, the ids should not be changed, but I'm not sure why that is.  Do you?

https://developers.google.com/protocol-buffers/docs/proto3

image

joecheuk: Updated.

google/cloud/forseti/services/model/service.py
```diff
@@ -133,16 +137,30 @@ def GetModel(self, request, _):
...
+        """Get model created_at datetime in string format.
+        Example return value: '2011-11-03 18:21:26'
+
+        Args:
+            model (Model): Model dao object.
+
+        Return:
+            str: created_at datetime in string format.
+        """
+        return model.created_at_datetime.now().strftime('%Y-%m-%d %H:%M:%S')
```
blueandgold: Can we use one of the centralized timestamp string from here?  `DEFAULT_FORSETI_HUMAN_TIMESTAMP`

https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/google/cloud/forseti/common/util/string_formats.py#L49

joecheuk: Updated.

google/cloud/forseti/services/model/service.py
```diff
@@ -133,16 +138,31 @@ def GetModel(self, request, _):
...
+                                          createdAt=created_at_str,
                                           description=model.description,
                                           message=model.message,
                                           warnings=model.warnings)
         return model_pb2.ModelDetails()
 
+    @staticmethod
+    def _get_model_created_at_str(model):
+        """Get model created_at datetime in string format.
+        Example return value: '2011-11-03 18:21:26'
```
blueandgold: I think this might not match what the `DEFAULT_FORSETI_HUMAN_TIMESTAMP`, so might want to drop it.
Instead in the doc string, can just say:
```
Get model created_at datetime in human readable string format.
```

PR #1488, opened 2018-05-03
Prevent cron job to run if the previous job is not yet finished.

### LINKED ISSUE #1436: run_forseti.sh should detect previously launched jobs and act accordingly
`run_forseti.sh` is the command executed via cron. This command should detect that existing invocations of it are _still_ running and should log and then die. Configuration around the number of times to do this could also be implemented to work around hung jobs.

joecheuk:
Resolves #1436

### DIFF
--- a/deployment-templates/compute-engine/server/forseti-instance-server.py
+++ b/deployment-templates/compute-engine/server/forseti-instance-server.py
@@ -212,7 +212,7 @@ USER=ubuntu
 # queue up the jobs.
 # If the cron job failed the acquire lock on the process, it will log a warning message to syslog.
 
-(echo "{run_frequency} /usr/bin/flock -n /tmp/forseti_cron_runner.lock $FORSETI_HOME/setup/gcp/scripts/run_forseti.sh || echo \"Warning: previous cron job 'run_forseti.sh' is still running and the new cron job is attempting to run, exiting...\" 2>&1 | logger") | crontab -u $USER -
+(echo "{run_frequency} /usr/bin/flock -n /tmp/forseti_cron_runner.lock $FORSETI_HOME/setup/gcp/scripts/run_forseti.sh || echo \"[forseti-security] Warning: New Forseti cron job will not be started, because previous Forseti job is still running.\" 2>&1 | logger") | crontab -u $USER -
 echo "Added the run_forseti.sh to crontab under user $USER"
 
 echo "Execution of startup script finished"
--- a/deployment-templates/compute-engine/server/forseti-instance-server.py
+++ b/deployment-templates/compute-engine/server/forseti-instance-server.py
@@ -212,7 +212,7 @@ USER=ubuntu
 # queue up the jobs.
 # If the cron job failed the acquire lock on the process, it will log a warning message to syslog.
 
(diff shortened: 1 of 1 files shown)


deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -203,7 +203,15 @@ def GenerateConfig(context):
...
-(echo "{run_frequency} $FORSETI_HOME/setup/gcp/scripts/run_forseti.sh") | crontab -u $USER -
+
+# Use flock to prevent rerun of the same cron job when the previous job is still running.
+# If the lock file does not exist under the tmp directory, it will create the file and put a lock on top of the file.
+# When the previous cron job is not finished and the new one is trying to run, it will attempt to acquire the lock
+# to the lock file and fail because the file is already locked by the previous process.
+# The -n flag in flock will fail the process right away when the process is not able to acquire the lock so we won't
+# queue up the jobs.
+
+(echo "{run_frequency} /usr/bin/flock -n /tmp/forseti_cron_runner.lock $FORSETI_HOME/setup/gcp/scripts/run_forseti.sh") | crontab -u $USER -
```
blueandgold: Does this resolve the scenario when the user-triggered forseti process is running, and it would be killed by the cron job restarting the server?

joecheuk: No, it does not - the issue you mentioned would be a separate fix to have the server not restart when we pull in the newer configurations from the gcs bucket.

blueandgold: I think there could / should be one solution that would apply to both scenarios, i.e. detect there is a running forseti process, whether cron-triggered or user-triggered, and should not be killed.  So, I like to hear more about how you would approach this.

blueandgold: Okay, let's go with this as discussed offline for now.  But it would be ideal for the second cron job to run in parallel alongside the first one, rather than just abort the second one, so that any environment change can be picked up.  We really also need to address the server not restarting so that user-triggered process does not get killed.

blueandgold: We also need to add a log somewhere that the second cron job is not running, due to the first one not finished.

deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -203,7 +203,18 @@ def GenerateConfig(context):
...
+# Use flock to prevent rerun of the same cron job when the previous job is still running.
+# If the lock file does not exist under the tmp directory, it will create the file and put a lock on top of the file.
+# When the previous cron job is not finished and the new one is trying to run, it will attempt to acquire the lock
+# to the lock file and fail because the file is already locked by the previous process.
+# The -n flag in flock will fail the process right away when the process is not able to acquire the lock so we won't
+# queue up the jobs.
+# If the cron job failed the acquire lock on the process, it will log a warning message to syslog.
+
+(echo "{run_frequency} /usr/bin/flock -n /tmp/forseti_cron_runner.lock $FORSETI_HOME/setup/gcp/scripts/run_forseti.sh || 
+    echo "Warning: previous cron job 'run_forseti.sh' is still running and the new cron job is attempting to run,
```
blueandgold: Should we add the `[forseti-security]` prefix, so that this can be easy to find, as our other application logs?

Also, the log message might be better as:
```
Warning: New Forseti cron job will not be started, because previous Forseti job is still running.
```

Issue #1406, opened 2018-04-12
Investigate and improve fluentd logging configuration


mirons-google:
In several issues we have poor logging because only portions of the logging statement are being logged.

Related issues:
#1315 #127 #1014 #1179 #1202 

This issue requests that we investigate [configuration](https://docs.fluentd.org/v1.0/articles/config-file#2-ldquomatchrdquo-tell-fluentd-what-to-do) options and/or changes that can be made to fix this.

blueandgold:
@ahoying @mirons-google 
I spent quite some time to look into and understand this issue.  As it turns out, we are currently already tied into stackdriver, with the installation of `google-fluentd` on the server VM.  

https://github.com/GoogleCloudPlatform/forseti-security/blob/master/deployment-templates/compute-engine/forseti-instance.py#L141

What this means, is that forseti logs are not even saved into the local `/var/log/syslog`.  Instead, they are intercepted and routed directly to stackdriver.  What this further means, is that it is not straightforward to use the `multiline` parser as that depends on the `in_tail` plugin which works on an actual file.

https://docs.fluentd.org/v0.12/articles/parser_multiline

Ultimately, we need to decide to stay with using `google-fluentd`, or use the stackdriver client library, or focus our effort to switch and configure a more platform-neutral version of fluentd.

PR #1489, opened 2018-05-04
Workaround for Stackdriver dropping newlines.

blueandgold:
This is workaround for issue #1406 , notably to fix the problem where log records with newlines `\n` are cut-off after the first line.  For a more long-term fix, we need to discuss & decide if we will continue to use google-fluentd (which means considering the stackdriver client library), or use the more platform neutral fluentd, where configuration efforts can be better justified.

More details in #1406


joecheuk:
Can we have a logger wrapper that replace all the new line characters with tabs when we log the message instead of updating all these individual logs?

PR #1494, opened 2018-05-05
Setup dedicated Forseti log to fix newlines being cut off on Stackdriver

### LINKED ISSUE #1406: Investigate and improve fluentd logging configuration

blueandgold:
Setup Dedicated Forseti log so that we can properly solve the log line being cut-off in stackdriver due to the usage of `\n`. 

image

I will have tested it on GCP, but will test the deployment portion when this PR looks good overall.

Fixes #1406


configs/logging/logrotate/forseti
```diff
@@ -0,0 +1,7 @@
+/var/log/forseti.log {
+  daily
+  rotate 30
+  missingok
+  notifempty
+  create 640 root root
```
ahoying: If forseti server isn't running as root it won't be able to write to the log. This should be set to be owned by the user the server runs as.

deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -159,6 +159,13 @@ def GenerateConfig(context):
 pip install -q --upgrade setuptools wheel
 pip install -q --upgrade -r requirements.txt
 
+# Setup Forseti logging
+touch /var/log/forseti.log
```
ahoying: Like above, you'll need to chown this to the user forseti server runs as.

google/cloud/forseti/common/util/logger.py
```diff
@@ -44,12 +50,16 @@ def get_logger(module_name):
...
-    # TODO: Move this into a configuration file.
-    syslog_handler = logging.handlers.SysLogHandler()
-    syslog_handler.setFormatter(logging.Formatter(SYSLOG_LOG_FMT))
+
+    if os.path.exists('/var/log/forseti.log'):  # ubuntu on GCE
+        default_log_handler = logging.FileHandler('/var/log/forseti.log')
+    else:
+        default_log_handler = logging.handlers.SysLogHandler()
+
+    default_log_handler.setFormatter(logging.Formatter(SYSLOG_LOG_FMT))
```
ahoying: We probably want to include a the timestamp as '%(asctime)s' before the levelname for the messages in forseti.log, so you need two formats, one for syslog, and one for forseti.log.

configs/logging/fluentd/forseti.conf
```diff
@@ -0,0 +1,24 @@
+<source>
+    @type tail
+
+    <parse>
+        @type multiline
+        format_firstline /^\[forseti\-security\]/
+        format1 /^(?<message>.*)/
```
ahoying: Once you add the time in, you need to extract it in this parser so that the log has the right timestamp in stackdriver:

https://docs.fluentd.org/v0.12/articles/parser_multiline#example

google/cloud/forseti/common/util/logger.py
blueandgold: image

configs/logging/logrotate/forseti
blueandgold: Done

deployment-templates/compute-engine/server/forseti-instance-server.py
blueandgold: Done

google/cloud/forseti/common/util/logger.py
ahoying: Just a nit, this will cause the timestamp to be output twise for the SysLogHandler since syslog also adds a timestamp. I'm not sure that's really a problem.

blueandgold: Done, we came this far, might as well take care of this.  :)

blueandgold:
Even making this filterable by level now.
image

PR #1497, opened 2018-05-07
Terminate the installation if the deployment process failed.

### LINKED ISSUE #1495: Installer should check if the deployment manager is finished properly
Installer should check if the deployment manager is finished properly, output warning/error message and exit the installation if the deployment contains errors.

joecheuk:
Resolves #1495

### DIFF
--- a/setup/gcp/installer/util/gcloud.py
+++ b/setup/gcp/installer/util/gcloud.py
@@ -830,11 +830,9 @@ def check_deployment_status(deployment_name, status):
         ['gcloud', 'deployment-manager', 'deployments', 'describe',
          deployment_name, '--format=json'])
 
-    deployment_err_msg = 'Error occurred during the deployment, exiting...'
-
     if return_code:
         print(err)
-        print(deployment_err_msg)
+        print(constants.MESSAGE_DEPLOYMENT_ERROR)
         sys.exit(1)
 
     deployment_info = json.loads(out)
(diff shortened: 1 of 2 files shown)


setup/gcp/installer/util/gcloud.py
```diff
@@ -827,13 +830,24 @@ def check_deployment_status(deployment_name, status):
...
+    deployment_operation = deployment_info['deployment']['operation']
+
+    deployment_status = deployment_operation['status']
+
+    deployment_error = deployment_operation.get('error', {})
+
+    if deployment_error:
+        print(deployment_error)
+        print(deployment_err_msg)
+        sys.exit(1)
```
blueandgold: Is termination the right thing to do?  Or is it something recoverable that the user can try again, after the rest of the installation is finished.

joecheuk: After some research on this topic, this is actually very hard to generalize since not all the errors are recoverable (for example, if you have an instance with the same name, the deployment will always fail no matter how many times you try it). I think by printing the error and exit the installation is a safe way of handling the exceptions for now since the user will have the knowledge of what happened and will be able to either file an issue with the error message or re install again if it's the error is not something serious. We should add more documentation around this area, let me update the error message to something more clear to the user knows when the db related issue occurred, they can try re running the installation process.

PR #1508, opened 2018-05-08
Added edge case check when there is no rule defined in forwarding rule.

### LINKED ISSUE #1507: Forwarding rule scanner is producing violations even if there is no rule
The scanner is producing results even when all the rules in the forwarding rule file is commented out.

joecheuk:
Resolves #1507

### DIFF
--- a/google/cloud/forseti/scanner/audit/forwarding_rule_rules_engine.py
+++ b/google/cloud/forseti/scanner/audit/forwarding_rule_rules_engine.py
@@ -75,6 +75,8 @@ class ForwardingRuleRulesEngine(bre.BaseRulesEngine):
 
         # If there is no forwarding rules defined in the rule file then no
         # forwarding rule is violated.
+        # TODO: Maybe we can move this up a level so we don't have to go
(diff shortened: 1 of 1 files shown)


google/cloud/forseti/scanner/audit/forwarding_rule_rules_engine.py
```diff
@@ -73,6 +73,11 @@ def find_policy_violations(self, forwarding_rule,
             self.build_rule_book()
         resource_rules = self.rule_book.get_resource_rules()
 
+        # If there is no forwarding rules defined in the rule file then no
+        # forwarding rule is violated.
+        if not resource_rules:
```
blueandgold: I am just wondering if it would be more efficient to have this check upstream, where the rule book is built, and raise an exception so that we can save some cycle like this.  Okay as is, but please put a TODO for the future.

joecheuk: Updated.

Issue #1511, opened 2018-05-09
Consider An Easier Way to Run the Server in Debug Mode


blueandgold:
Currently, when doing troubleshooting, we need to stop forseti server via `systemctl` command, and then manually find the server command and then start it with the ```--log_level debug```.  It would be awesome if there's an easier workflow for this.

joecheuk:
Now this can be done with

`forseti server log_level set <LOG_LEVEL>`

PR #1514, opened 2018-05-09
Added server_config service and cli

### LINKED ISSUE #1396: Mitigate Restarting the Server for Reading Updated Config File
Currently, the server needs to be restarted in order to read any changes in an updated `forseti_conf.yaml` file.  This would effectively kill another running processes.  Such as if the user is doing something and the cron job runs.

So, it would be nice if the server does not have to be restarted, i.e. the server will read in the conf file when it does an operation.

joecheuk:
Resolves #1396  #1511 

You are now able to reload the configuration, get/set log level with server_config service.

e.g.

```
forseti server log_level get

forseti server log_level set

forseti server configuration get 

forseti server configuration reload **optional_config_file_path**
```


blueandgold:
One small change to be consistent with current usage from `forseti config show`.
```
forseti server_config log_level show
```

google/cloud/forseti/services/client.py
```diff
@@ -129,6 +131,66 @@ def run(self):
                              metadata=self.metadata())
 
 
+class ServerConfigClient(ForsetiClient):
+    """Server configuration service allows the client to update the
```
blueandgold: This should be one line.  How about `Allows the client to update the server configuration.`

google/cloud/forseti/services/cli.py
```diff
@@ -164,61 +164,66 @@ def define_config_parser(parent):
...
+        'get',
+        help='Get the server configuration.'
     )
 
-    reload_config_parser = action_subparser.add_parser(
-        'reload_configuration',
-        help='Reload server configuration.')
+    reload_config = config_subparser.add_parser(
+        'reload',
+        help='Load the server configuration.'
```
blueandgold: `reload` instead of `load`

google/cloud/forseti/services/cli.py
```diff
@@ -684,7 +689,7 @@ def do_run():
     actions[config.action]()
 
 
-def run_server_config(client, config, output, _):
+def run_server(client, config, output, _):
     """Run scanner commands.
```
blueandgold: Not scanner

google/cloud/forseti/services/server.py
```diff
@@ -280,12 +277,104 @@ def __init__(self,
...
+            # Lock before performing the update to avoid multiple updates
+            # at the same time.
+
+            self.forseti_config = forseti_config
+
+            # Setting up individual configurations
+            forseti_inventory_config = forseti_config.get('inventory', {})
+            inventory_config = InventoryConfig(
+                forseti_inventory_config.get('root_resource_id', ''),
+                forseti_inventory_config.get('gsuite_service_account_key_file',
```
blueandgold: Can we get rid of this now?

google/cloud/forseti/services/server/service.py
```diff
@@ -0,0 +1,155 @@
...
+        Args:
+            request (ReloadServerConfigurationRequest): The grpc request object.
+            _ (object): Context of the request.
+
+        Returns:
+            ReloadServerConfigurationReply: The ReloadConfigurationReply grpc object.
+        """
+
+        LOGGER.info('Reloading server configurations')
+        success, err_msg = self.service_config.update_configuration(
```
blueandgold: Instead of `success`, this should be more clear as `is_success`, so that we can avoid something where `success = False`

PR #1526, opened 2018-05-11
2.0 inventory db cleanup

ahoying:
Resolves: #1537, #1529 

Update sql schema to improve performance.

* Update inventory dao:
  * Use a bigint for the id instead of a string, the id is based a microsecond timestamp.
  * Use an enum for the category instead of a string.
  * Add a new parent_id field to replace parent_resource_id and
  parent_resource_type for joining a resource to its parent.
  * Store the id from the database in the resource as the inventory_key.
  * Add duplicate resource detection on write, and change to update.
  * Clean up indexes to use the new schema format efficiently.

* Update model importer:
  * Use resource id and parent_id for joining resource to its parent.
  * Clean up flush logic.
  * Remove duplicate resource detection, since that was moved to the
  inventory dao.

* Update scanner and violations dao:
  * Use a bigint for the id instead of a string, like inventory dao.

* Clean up tests.


google/cloud/forseti/common/util/date_time.py
```diff
@@ -75,21 +76,53 @@ def get_utc_now_datetime():
...
     Returns:
           str: A timestamp in the classes default timestamp format.
     """
-    utc_now = get_utc_now_datetime()
+    utc_now = date or get_utc_now_datetime()
     return utc_now.strftime(string_formats.DEFAULT_FORSETI_HUMAN_TIMESTAMP)
 
 
-def get_utc_now_timestamp():
+def get_utc_now_timestamp(date=None):
```
blueandgold: qq: is the optional arg here for testing purpose?  If yes, I can find some existing examples of how to patch this.

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -1343,16 +1278,16 @@ def _is_role_unique(self, role_name):
...
 
         Args:
             resource (object): Resource to put in the cache.
+            resource_id (int): The database key for the resource.
         """
 
-        type_name = resource.type_name
         full_res_name = resource.full_name
-        self.resource_cache[type_name] = (resource, full_res_name)
+        self.resource_cache[resource_id] = (resource, full_res_name)
```
blueandgold: qq: can you please explain a bit why this change from `type_name` to `resource_id` is made?

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -827,15 +820,13 @@ def _convert_iam_policy(self, iam_policy):
             iam_policy (object): IAM policy to store.
         """
         _, full_res_name = self._get_parent(iam_policy)
-        parent_type_name = self._parent_type_name(iam_policy)
+        parent_type_name = self._type_name(iam_policy)
```
blueandgold: qq: another question here about the switch from `_parent_type_name()` to `_type_name()`

google/cloud/forseti/services/model/importer/importer.py
ahoying: This change is made because we now track the relationship between parent and child resources using resource.parent_id = resource.id. The is no longer a parent_resource_type and parent_resource_id to join on.

google/cloud/forseti/common/util/date_time.py
ahoying: It's not just for testing. We also use it anywhere we already have a datetime object and just need it formatted.

google/cloud/forseti/services/model/importer/importer.py
ahoying: An iam_policy has the same type_name as it's parent object, and we no longer track the parent_resource_type and parent_resource_id in the object.

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -730,13 +713,26 @@ def write(self, resource):
...
+            return
+
         rows = Inventory.from_resource(self.inventory_index, resource)
+
+        # Insert the resource row to get the inventory id, buffer other rows.
         for row in rows:
-            self.buffer.add(row)
+            if row.category == Categories.resource:
+                self.session.add(row)
+                self.session.flush()
```
joecheuk: I think flushing everytime is very expensive, using the buffer should be able to increase the performance, I tried flushing every row vs the db buffer flush every 1024 rows, the difference is roughly 1 minute in our test domain (3mins 2x seconds vs 4mins 2x seconds)

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -730,13 +713,26 @@ def write(self, resource):
...
+                resource.set_inventory_key(row.id)
+            else:
+                row.parent_id = resource.inventory_key()
+                self.buffer.add(row)
```
joecheuk: self.session.flush will not flush the buffer, you will need to call the commit() method in the Storage class to do that.

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -205,16 +204,22 @@ def run(self):
...
-                    if not item_counter % 3000:
+                    if not item_counter % 1000:
+                        # Flush database every 1000 resources
                         LOGGER.debug('Flushing model write session: %s',
                                      item_counter)
                         self.session.flush()
-                LOGGER.debug('Start storing resources into models.')
+
                 self._store_resource(None, last_res_type)
+                if item_counter % 1000:
```
joecheuk: Need to increment the item_counter by 1 after the last `self._store_resource(None, last_res_type)` line otherwise item_counter will remain the same as before.

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -321,20 +326,26 @@ def model_action_wrapper(session,
...
+            if not idx % flush_count:
+                # Flush database every flush_count resources
+                LOGGER.debug('Flushing write session: %s.', idx)
                 session.flush()
 
         if post_action:
             post_action()
-        return item_counter
+
+        if idx % flush_count:
```
joecheuk: This will happen only if the post_action modify the idx variable.

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -638,17 +598,40 @@ def _get_resource_rows(self, key):
...
+        Returns:
+            int: The resource id of the existing resource, else 0.
+        """
+        row = self.session.query(Inventory.id).filter(
+            and_(
+                Inventory.inventory_index_id == self.inventory_index.id,
+                Inventory.category == Categories.resource,
+                Inventory.resource_type == resource.type(),
+                Inventory.resource_id == resource.key(),
+            )).one_or_none()
```
joecheuk: one_or_none() will raise `sqlalchemy.orm.exc.MultipleResultsFound` if multiple object identities are returned. We are expecting multiple rows with same inventory_index_id, category, resource_type and resource_id, what should we do in that case?

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -652,7 +661,7 @@ def _convert_appengine_resource(self, gae_resource):
             parent=parent)
 
         self.session.add(resource)
```
joecheuk: If there is no is unique check on the resource, I think the duplicate key will occur again because the resource object has no reference to the unique identifier of the inventory resource (gae_resource.id), this is same for other resources. So what might happen is when 2 gae_resource with different resource id and same contents, it will generate the same resource object and insert it to the resource table.

google/cloud/forseti/services/inventory/storage.py
ahoying: This code is written to ensure that there is only ever one or none. In write, if one already exists the write is transformed to an update to ensure there are no duplicates.

If we ever get a MultipleResultsFound that should be an exception to blocks the creation of the inventory because a core assertion has been violated.

google/cloud/forseti/services/model/importer/importer.py
ahoying: _store_resource is passed in a resource of None, that call is to only ensure that any post handlers for a resource are run if they exist. So the item counter doesn't need to increment.

google/cloud/forseti/services/model/importer/importer.py
ahoying: No, it happens if the number of items in inventory_iterable is not an even divisor for the flush_count.

For example:

len(inventory_iterable) = 1300
flush_count = 1000

There will be one flush when idx = 1000, and then after the for loop returns, idx will = 1300, which will trigger the second flush for the remain 300 rows.

google/cloud/forseti/services/model/importer/importer.py
ahoying: The unique constraint was added to the inventory, so this is explicitly no longer an issue when building the model.

google/cloud/forseti/services/inventory/storage.py
ahoying: commit is still done like before when the inventory is complete, nothing changed there.

google/cloud/forseti/services/inventory/storage.py
ahoying: I made this more explicit in the code. If you re-run the test with the new schema, you'll see this code is significantly faster than the previous version due to the more efficient schema and index.

google/cloud/forseti/services/inventory/storage.py
joecheuk: Instead of query the database for every single insert, what do you think about having a look up table to increase the performance? if the size of the table is a concern maybe we can use a customized bloom filter?

Issue #1527, opened 2018-05-12
IOError: [Errno 13] Permission denied: '/var/log/forseti.log'


blueandgold:
This happened when I use the the forseti client on the server VM as myself.

But when I use the forseti client on the client VM, there is no error and I get a good reseult

```
henry@forseti-server-vm-0753:~$ ls -la /var/log/forseti.log
-rw-r--r-- 1 ubuntu root 29546 May 12 02:17 /var/log/forseti.log
```

```
Last login: Sat May 12 00:31:56 2018 from 173.194.94.96
henry@forseti-server-vm-0753:~$ forseti model list

Traceback (most recent call last):
  File "/usr/local/bin/forseti", line 11, in <module>
    load_entry_point('forseti-security==2.0.0', 'console_scripts', 'forseti')()
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/stubs.py", line 43, in RunForsetiCli
    import google.cloud.forseti.services.cli as forseti_cli
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/services/cli.py", line 24, in <module>
    from google.cloud.forseti.common.util import file_loader
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/common/util/file_loader.py", line 21, in <module>
    from google.cloud.forseti.common.gcp_api import storage
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/common/gcp_api/storage.py", line 23, in <module>
    from google.cloud.forseti.common.gcp_api import _base_repository
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/common/gcp_api/_base_repository.py", line 30, in <module>
    from google.cloud.forseti.common.gcp_api import errors as api_errors
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/common/gcp_api/errors.py", line 19, in <module>
    LOGGER = logger.get_logger(__name__)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.0.0-py2.7.egg/google/cloud/forseti/common/util/logger.py", line 60, in get_logger
[...]

PR #1528, opened 2018-05-12
Make the forseti.log writeable when user tries to use the client in the server VM.

blueandgold:
More details in #1527 .  @ahoying Is this a good thing to do, from a security point of view?  The assumption here is that anyone who can get into the VM, then he is an admin, and should be able to write to the ```/var/log/forseti.log```.

But please let me know if there is something better.


blueandgold:
@ahoying has suggested to make the `/var/log/forseti.log` not writeable to the cli user.  And instead wrap it in a try to fall back to syslog.

PR #1530, opened 2018-05-14
Writetosyslog

blueandgold:
Users of the CLI on the server VM can not write to `/var/log/forseti.log`, so falling back to syslog.

Creating this PR in this branch here for easier testing, since there is a GCP VM that will pull from this branch.  Afterwards, will move to the the original PR #1528


blueandgold:
This was just for testing.

PR #1532, opened 2018-05-14
Merge rc2 into dev

joecheuk:


configs/server/forseti_conf_server.yaml.sample
```diff
@@ -57,6 +57,11 @@ inventory:
     #   0 : delete all previous inventory data before running
     retention_days: -1
 
+    # Number of days to retain inventory data:
```
blueandgold: This would duplicate lines 55 to 58

joecheuk: Updated.

PR #1542, opened 2018-05-16
Add lightbox to all responsive images

denzel-morris:
Users can now click/tap on any responsive images in the documentation and it will open a standard lightbox with the full-size image.


mirons-google:
Here's an example of what's broken with this PR

```
- ./_www/www/search/index.html
  *  internal script /third_party/lightbox/js/lightbox.min.js does not exist (line 193)
  *  internally linking to /third_party/lightbox/css/lightbox.min.css, which does not exist (line 37)
     <link rel="stylesheet" type="text/css" href="/third_party/lightbox/css/lightbox.min.css">
```

denzel-morris:
Thanks for catching that; it was due to the top-level `.gitignore` file ignoring all `dist/`. I've added an exception for `third_party/`.

PR #1545, opened 2018-05-16
Updated to use compute engine library's ping method to check if metadata server is available 

### LINKED ISSUE #1539: 2018-05-15 11:34:43,336 ERROR google.cloud.forseti.common.util.metadata_server(_issue_http_request): Error with request: [Errno -2] Name or service not known
Got this when creating inventory from 2.0-dev.

image

joecheuk:
Resolves #1539


google/cloud/forseti/common/util/metadata_server.py
```diff
@@ -64,7 +64,7 @@ def _issue_http_request(method, path, headers):
         http_client.request(method, path, headers=headers)
         return http_client.getresponse()
     except (socket.error, httplib.HTTPException) as e:
-        LOGGER.error('Error with request: %s', e)
+        LOGGER.debug('Error with request: %s', e)
```
ahoying: I think instead we should change METADATA_SERVER_HOSTNAME to the IP instead of the host name, which may not resolve. The IP is '169.254.169.254'. Take a look at https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/compute_engine/_metadata.py. We may even want to use the environment variable like they are.

blueandgold: qq: Is there any concern that the IP address could change in the future?

joecheuk: Updated to to use compute engine's metadata ping function to determine whether or not the meta server is reachable

google/cloud/forseti/common/util/metadata_server.py
```diff
@@ -20,6 +20,9 @@
 import httplib
 import socket
 
+from google.auth.compute_engine import _metadata
```
ahoying: I'm a little concerned about relying on a private module, as they could change the implementation at some point, I'd like to have some test to validate this is working as intended.

Issue #1325, opened 2018-04-02
Handle exception when Client can not connect to Server


blueandgold:
When the client fails to connect to the server, we get an stack trace as below.  This should be handled more gracefully.

![image](https://user-images.githubusercontent.com/6977544/38201038-96d807d2-364b-11e8-9a85-cbefb9a64e2d.png)

PR #1552, opened 2018-05-17
print error msg as opposed to  a stack trace when server is down

arbeit:
Fixes issue #1325. The cli now prints a nice error message and exits with a non-zero value in case a `grpc` exception is raised.


blueandgold:
arbeit was a googler and does not need CLA.

Issue #1493, opened 2018-05-04
sqladmin.v1beta4 returning 403 "operationInProgress" during the deployment


joecheuk:
![image](https://user-images.githubusercontent.com/5330393/39656728-ddf63a1a-4fb6-11e8-8a74-39f6c6473899.png)

This happens once in a while, I suspect it's because both database creation and SQL instance user creation are ran at the same time and that's why we are having this error.

blueandgold:
I just see this too.  And I am on the latest 2.0-dev.

PR #1554, opened 2018-05-17
changed db user back to root

joecheuk:
Resolves: #1493 

Related: #921

### DIFF
--- a/install/gcp/scripts/initialize_forseti_services.sh
+++ b/install/gcp/scripts/initialize_forseti_services.sh
@@ -28,7 +28,7 @@ if ! [[ -f $FORSETI_SERVER_CONF ]]; then
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


install/gcp/scripts/initialize_forseti_services.sh
```diff
@@ -28,7 +28,7 @@ if ! [[ -f $FORSETI_SERVER_CONF ]]; then
     exit 1
 fi
 
-SQL_SERVER_LOCAL_ADDRESS="mysql://forseti-db-admin@127.0.0.1:${SQL_PORT}"
+SQL_SERVER_LOCAL_ADDRESS="mysql://root@127.0.0.1:${SQL_PORT}"
```
blueandgold: Can you put a comment here that there is an issue to create a non-root user, and link to the the issue?

PR #1557, opened 2018-05-17
Added exception handling in the main method.

joecheuk:


google/cloud/forseti/services/cli.py
```diff
@@ -15,6 +15,7 @@
 """Forseti CLI."""
 
 from argparse import ArgumentParser
+import grpc
```
blueandgold: This is not a standard library, and should be move down below, and separated from the standard libraries by a newline.

PR #1591, opened 2018-05-23
Update resource name to use hash of timestamp & org id

joecheuk:


install/gcp/installer/configs/config.py
```diff
@@ -40,3 +41,21 @@ def __init__(self, **kwargs):
...
+        Hash the timestamp and organization id and take the first 7 characters.
+
+        Lowercase is needed because some resource name are not allowed to have
+        uppercase.
+
+        Args:
+            organization_id (str): Organization id.
+        """
+        if not self.identifier:
+            message = self.datetimestamp + organization_id
```
blueandgold: nit: small request, can you please make it "organization_id + self.datetimestamp".  This ordering just feels we are going form top to bottom.

install/gcp/installer/configs/config.py
```diff
@@ -40,3 +41,21 @@ def __init__(self, **kwargs):
         self.dry_run = bool(kwargs.get('dry_run'))
         self.bucket_location = kwargs.get('gcs_location')
         self.installation_type = None
+
+    def generate_identifier(self, organization_id):
+        """Generate resource unique identifier.
+
+        Hash the timestamp and organization id and take the first 7 characters.
```
blueandgold: Can you please add a quick comment here that this hashing combination is to ensure global uniqueness of the bucket names?

install/gcp/installer/configs/config.py
joecheuk: Updated.

install/gcp/installer/configs/config.py
joecheuk: Updated.

PR #1622, opened 2018-05-31
Added parameter to skip SendGrid configuration

angelzamir:
Hello,

This PR is for allow executing Forseti with an extra parameter (--skip-sendgrid-config) in order to avoid the SendGrid API Key input (raw_input) and the recipient email through console. 

#### Detail:
When executing Forseti without specify the sendgrid_api_key and notification_recipient_email parameters (or setting them as ""), later on the installation Forseti asks for them, so, there's no a programatic way to avoid the SendGrid configuration without interactive input.


angelzamir:
The pylint failed with:

```
R: 20, 0: Too many instance attributes (8/7) (too-many-instance-attributes)
```

It could be possible to increase that limit? or may be put the comment to avoid it?

```py
# pylint: disable=too-many-instance-attributes
```

Thanks!

joecheuk:
Hi Angel,

Can you add the pylint comment beside the method?

Thanks,
Joe

angelzamir:
Hi Joe,

Done

Thanks!

install/gcp/installer/configs/server_config.py
```diff
@@ -37,6 +39,7 @@ def __init__(self, **kwargs):
         self.notification_recipient_email = (
             kwargs.get('notification_recipient_email'))
         self.gsuite_superadmin_email = kwargs.get('gsuite_superadmin_email')
+        self.skip_sendgrid_config = bool(kwargs.get('skip_sendgrid_config'))
```
joecheuk: If I am not mistaken, action='store_true' will store the boolean value (True/False), is the cast here necessary?

angelzamir: I did it based on this: 
https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-rc3/install/gcp/installer/configs/config.py#L36

joecheuk: ok, let's stick with this style for now then

joecheuk:
One more comment - can you change the PR to point toward 2.0-dev instead of 2.0-rc3?

angelzamir:
Hi, done.

Thanks!

joecheuk:
Seems like there are a lot of conflicts, the RC3 branch is supposed to be frozen other than bug fixes so this change should really go into 2.0-dev instead. Can you resolve the conflicts or create a new PR with base branch=2.0-dev?

angelzamir:
Right, done at: #1624

PR #1625, opened 2018-06-01
Update architecture.md

mirons-google:


mirons-google:
Good catch! Fixed. PTAL.

Tabrarian:
I'm not sure if it likes that; the style md is printing on the page and the list is showing in numbers. I found this reference that might help, but I think it would require changes to the base CSS: https://gist.github.com/bertobox/3503850

mirons-google:
Ah, disregard the way GitHub renders it as it uses a different markdown converter. I verified it works with ours.

PR #1627, opened 2018-06-01
Added parameter to skip check authed user in domain

angelzamir:
Hello,

This PR is for allow executing Forseti with an extra parameter (--skip-check-authed-user) in order to avoid the check if authed user is in the domain. Very similar to #1624 

#### Detail:
When executing Forseti outside cloud shell, there's no a programatic way to avoid Forseti asking confirmation (y/n) through the console when the authed user is not in the same domain as the organization.


install/gcp/gcp_installer.py
```diff
@@ -77,6 +77,9 @@ def run():
                         choices=['client', 'server'],
                         help='Type of the installation, '
                              'either client or server')
+    parser.add_argument('--skip-check-authed-user',
+                        action='store_true',
+                        help='Skip check if auther user is in domain')
```
joecheuk: typo: authed

install/gcp/installer/forseti_installer.py
```diff
@@ -199,8 +199,9 @@ def preflight_checks(self):
                                          is_cloudshell)
         self.organization_id = gcloud.lookup_organization(self.project_id)
         self.config.generate_identifier(self.organization_id)
-        self.check_if_authed_user_in_domain(
-            self.organization_id, authed_user)
+        if not self.config.skip_check_authed_user:
```
joecheuk: If we are using a service account to auth gcloud, shouldn't the authed user become the service account in this case? Is there any reason we need to skip this?

joecheuk:
I think it makes more sense to create a "--quiet" flag and disable all the q&a when the flag is true instead of having one flag per question.

blueandgold:
Or even an "--auto" mode, for automated installation without the promptings.  Also, can we take a step back and consider this more holistically, such as what we are trying to accomplish here so that we have a more coherent approach that is maintainable and flexible between the user mode, and auto mode?

install/gcp/installer/forseti_installer.py
angelzamir: That makes sense, so instead of the param --skip-check-authed-user, the check depends on if there's a service account activated? I mean, delete this param and only check if user belongs to domain when there's no a service account identified? @joecheuk

angelzamir:
I've included the change on second commit at #1640

PR #1640, opened 2018-06-05
Added support for using service account instead of a user

angelzamir:
Hello,

In some cases outside of Cloud Shell we can't use a user but a service account. This PR helps detect if there's any credentials file in the environment variables and activate the service account for using it instead of a specific user.

### DIFF
--- a/install/gcp/installer/forseti_installer.py
+++ b/install/gcp/installer/forseti_installer.py
@@ -192,7 +192,8 @@ class ForsetiInstaller(object):
         utils.print_banner('Pre-installation checks')
         self.check_run_properties()
         self.branch = utils.infer_version(self.config.advanced_mode)
-        self.project_id, authed_user, is_cloudshell = gcloud.get_gcloud_info()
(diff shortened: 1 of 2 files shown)


morgante:
@joecheuk @blueandgold Could you take a look?

joecheuk:
Hi @angelzamir  @morgante , we are releasing Forseti v2.0 soon so in the coming week or so we will be focusing on doing that and we don't plan to merge in any new changes during this time period. We should be able to get the PR in after the release. Is that fine with you?

morgante:
That will likely significantly delay the Forseti Terraform module, as @angelzamir will be working on other projects following the end of this week.

@angelzamir Is it possible to build against a fork for now and have our module use that fork until this is merged?

angelzamir:
@morgante sure, I think the module could points to my fork, working on that.

morgante:
@blueandgold Can we keep this open? I believe it's still needed for Terraform support.

PR #1642, opened 2018-06-06
setup updates

joecheuk:


_docs/_latest/setup/install.md
```diff
@@ -42,31 +42,37 @@ steps below:
 
 ### Run setup
 
-  1. After you've started Cloud Shell, download Forseti. The installer is
-     included. Getting `2.0-dev-rc1` branch will install Forseti v2.0 Release
-     Candidate 1.
+  1. After you've started Cloud Shell, download Forseti. The installer is included.
```
mirons-google: included where? when?

_docs/_latest/setup/install.md
```diff
@@ -42,31 +42,37 @@ steps below:
...
 
      To install both client and server:
      
      ```bash 
-     python setup/installer.py
+     python install/gcp_installer.py
      ```
 
-      To see additional configurations for the setup:
+     To see additional configurations for the setup:
```
mirons-google: Please disregard use of a colon before we show something, it's not needed.

mirons-google: Please fix this everywhere in this doc, e.g. L90.

_docs/_latest/setup/install.md
mirons-google: Please put a `---` before the first section.

joecheuk: Updated

joecheuk: Updated

_docs/_latest/setup/install.md
joecheuk: Updated

PR #1650, opened 2018-06-06
Updated upgrade documentation

joecheuk:
The upgrade strategy has been changed due to various concerns, here is the updated version.

### DIFF
--- a/_docs/_latest/setup/upgrade.md
+++ b/_docs/_latest/setup/upgrade.md
@@ -5,71 +5,44 @@ order: 00
 
 # {{ page.title }}
 
-This guide explains how to use the Forseti upgrade tool.
+This guide explains how to upgrade your Forseti from v1.x to v2.
 
-## Before you begin
+## Important notes
 
-Before you upgrade Forseti, you'll need the following:
+ * We currently don't support data migration from v1 to v2, so you will need to archive the database manually 
+   for future reference if the data is important to you.
+ * [Forseti v2 configuration]({% link _docs/latest/configure/forseti/index.md %}) is different than v1 so 
+   you can not replace the v2 configuration file with the v1 configuration file.
+ * Resources in inventory are no longer configurable.
 
- * A G Suite super admin account to complete the [Domain-wide delegation steps]({% link _docs/latest/configure/gsuite.md %}). 
-Forseti 2.0 requires G Suite being enabled, so you'll need this before you start the installation.
+## Upgrading your v1 instance
 
-{% capture 1x_upgrade %}
+The suggested way of upgrading your Forseti instance is to do create a new project, [install 
+the latest v2 instance]({% link _docs/latest/setup/install.md %}), move all the rule files 
+from the v1 bucket to the v2 bucket.
 
-## Important caveats
-
- * The Forseti Security 2.0 installer only migrates configuration and rule files.
- * The Forseti Security 2.0 installer will attempt to re-use previous service accounts and projects.
- * The Forseti Security 2.0 installer will not destroy existing v1.X data.
- * Forseti Security 2.0 data is not compatible with v1.X's and therefore a new database is created.
  
-## Activate Google Cloud Shell
+### Moving the rule files from v1 to v2
 
-It's recommended to use [Cloud Shell](https://cloud.google.com/shell/docs/quickstart) to run the
-Forseti installer. This ensures you're using the latest version of Cloud SDK since it's included
-in Cloud Shell. To prepare to run the Forseti setup wizard, follow the steps below:
+Since the functionality of the existing scanners didn't change, you can re-use the rule 
+files defined for your v1 instance in v2.  
 
-  1. Access the [Cloud Platform Console](https://console.cloud.google.com/).
-  1. In the **Select a project** drop-down list at the top of the console, select the project where
-  you have Forseti v1.x deployed.
-  1. On the top right of the console, click **Activate Google Cloud Shell**. The Cloud
-  Shell panel opens at the bottom of the page.
+Run the following command to copy all the rule files from the v1 bucket to the v2 bucket.
+
+```bash
+# Replace <YOUR_V1_BUCKET> with your v1 forseti GCS bucket and
+# <YOUR_V2_BUCKET> with your v2 forseti GCS bucket.
+
+gsutil cp gs://<YOUR_V1_BUCKET>/rules/*.yaml gs://<YOUR_V2_BUCKET>/
+```
+
   
-## Run setup
+### Archiving your existing Cloud SQL Database
 
-  1. After you activate Cloud Shell, download Forseti. The installer is included.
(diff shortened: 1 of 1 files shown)


_docs/_latest/setup/upgrade.md
```diff
@@ -5,71 +5,44 @@ order: 00
...
 # {{ page.title }}
 
-This guide explains how to use the Forseti upgrade tool.
+This guide explains how to upgrade your Forseti from v1.x to v2.
 
-## Before you begin
+## Important notes
 
-Before you upgrade Forseti, you'll need the following:
+ * We currently don't support data migration from v1 to v2, so you will need to archive the database manually
```
mirons-google: remove the use of the word currently as we don't have any intention so this won't change.

_docs/_latest/setup/upgrade.md
```diff
@@ -5,71 +5,44 @@ order: 00
...
 
-## Before you begin
+## Important notes
 
-Before you upgrade Forseti, you'll need the following:
+ * We currently don't support data migration from v1 to v2, so you will need to archive the database manually 
+   for future reference if the data is important to you.
+ * [Forseti v2 configuration]({% link _docs/latest/configure/forseti/index.md %}) is different than v1 so 
+   you can not replace the v2 configuration file with the v1 configuration file.
+ * Resources in inventory are no longer configurable.
```
mirons-google: Configuration of the resources that are inventoried is not configureable

_docs/_latest/setup/upgrade.md
```diff
@@ -5,71 +5,44 @@ order: 00
 
 # {{ page.title }}
 
-This guide explains how to use the Forseti upgrade tool.
+This guide explains how to upgrade your Forseti from v1.x to v2.
```
mirons-google: a `---` before the first section.

joecheuk: Updated.

_docs/_latest/setup/upgrade.md
joecheuk: Updated.

_docs/_latest/setup/upgrade.md
joecheuk: Updated

_docs/_latest/setup/upgrade.md
```diff
@@ -5,72 +5,53 @@ order: 00
...
+files defined for your v1 instance in v2.  
+
+Run the following command to copy all the rule files from the v1 bucket to the v2 bucket.
 
+```bash
+# Replace <YOUR_V1_BUCKET> with your v1 forseti GCS bucket and
+# <YOUR_V2_BUCKET> with your v2 forseti GCS bucket.
+
+gsutil cp gs://<YOUR_V1_BUCKET>/rules/*.yaml gs://<YOUR_V2_BUCKET>/
+```
```
blueandgold: There should be something about how they can migrate their configurations.  How specific do you think we need to be?

joecheuk: I think there is a big enough different in v1 config and v2 config that the user should look at the config and configure it themselves in v2 instead of trying to migrate anything.

_docs/_latest/setup/upgrade.md
```diff
@@ -5,72 +5,105 @@ order: 00
...
+ * We don't support data migration from v1 to v2, so you will need to archive the database manually 
+   for future reference if the data is important to you.
+ * [Forseti v2 configuration]({% link _docs/latest/configure/forseti/index.md %}) is different than v1 so 
+   you can not replace the v2 configuration file with the v1 configuration file.
+ * Configuration of the resources that are inventoried is not configureable.
 
-{% capture 1x_upgrade %}
 
-## Important caveats
+## Upgrading your v1 instance
```
blueandgold: Hmmm, this is really not about upgrading v1 instance.  We should say this is about upgrading to v2.  How about something like this:
```
## Upgrade to v2
```

_docs/_latest/setup/upgrade.md
```diff
@@ -5,72 +5,105 @@ order: 00
...
+
+The suggested way of upgrading your Forseti instance is to do create a new project, [install 
+the latest v2 instance]({% link _docs/latest/setup/install.md %}), move all the rule files 
+from the v1 bucket to the v2 bucket.
```
blueandgold: We should add at the end, that they should let v1 run side-by-side and then shutdown when they are comfortable that v2 is working properly.

_docs/_latest/setup/upgrade.md
```diff
@@ -5,72 +5,105 @@ order: 00
...
-       [G Suite data collection]({% link _docs/latest/configure/gsuite.md %})
-       and is required.
-       Ask your G Suite admin if you don't know which super admin email to use.
-  1. Forseti is now upgraded to v2.x. To manually remove unused resources, follow the instructions
-  at the end of the installation process.
-  
-{% endcapture %} 
+### Moving the rule files from v1 to v2
+
+Since the functionality of the existing scanners didn't change, you can re-use the rule
```
blueandgold: I would remove the part "Since the functionality of the existing scanners didn't change, "

```
You can re-use the rule files defined for your v1 instance in v2 by copying them to the v2 buckets, and update the configuration as below.
```

_docs/_latest/setup/upgrade.md
joecheuk: Updated

_docs/_latest/setup/upgrade.md
joecheuk: Updated

_docs/_latest/setup/upgrade.md
joecheuk: Updated

_docs/_latest/setup/upgrade.md
```diff
@@ -5,72 +5,106 @@ order: 00
...
-Before you upgrade Forseti, you'll need the following:
+## Important notes
 
- * A G Suite super admin account to complete the [Domain-wide delegation steps]({% link _docs/latest/configure/gsuite.md %}). 
-Forseti 2.0 requires G Suite being enabled, so you'll need this before you start the installation.
+ * We don't support data migration from v1 to v2, so you will need to archive the database manually 
+   for future reference if the data is important to you.
+ * [Forseti v2 configuration]({% link _docs/latest/configure/forseti/index.md %}) is different than v1 so 
+   you can not replace the v2 configuration file with the v1 configuration file.
+ * Configuration of the resources that are inventoried is not configureable.
```
Tabrarian: I don't understand what this means. Do we mean that "you can't configure resources that are inventoried" or something else?

Tabrarian: Thanks for helping clarify this offline! How about changing this to:

In v2, all resources are inventoried. You won't be able to configure Inventory to include or exclude resources.

joecheuk: Thanks for the suggestion! Updated.

PR #1656, opened 2018-06-08
Update the server config to include the service account key scanner.

blueandgold:


configs/server/forseti_conf_server.yaml.sample
```diff
@@ -102,6 +102,8 @@ scanner:
           enabled: false
         - name: ke_version_scanner
           enabled: true
+        - name: service_account_key_scanner
```
joecheuk: This should be service_account_key instead of service_account_key_scanner

blueandgold: good catch, fixed!

PR #1665, opened 2018-06-11
Updated hardcoded resource types in violation to use the resource types defined in the ResourceType class

joecheuk:


google/cloud/forseti/common/gcp_type/resource.py
```diff
@@ -39,26 +39,28 @@ class ResourceType(object):
     SERVICE_ACCOUNT_KEY = 'serviceaccount_key'
 
     # Compute engine
-    BACKEND_SERVICE = 'backend_service'
-    FIREWALL_RULE = 'firewall_rule'
-    FORWARDING_RULE = 'forwarding_rule'
+    BACKEND_SERVICE = 'backendservice'
```
ahoying: To make sure this stays in sync, it would be nice if /google/cloud/forseti/services/inventory/base/resources.py either used these constants, or these constants were set from the .type() method on the resource classes. I prefer .../inventory/base/resources.py to be the source of truth, but that will require updating the type() method for all resource classes to be a staticmethod, so you can then reference resources.BackendService.type() for example. It makes more sense for that to be a staticmethod anyway, since it always returns a string constant.

joecheuk: Updated

PR #1680, opened 2018-06-14
Massive website cleanup

mirons-google:
This standardizes many things we do. It also improves the organization and titling.

- * instead of - for ordered lists
- One space after a period
- consistent use of codeblock descriptors
- : before a codeblock
- index.md ordering matches sidebar ordering
- No more wrapping title names
- consistent file naming replacing _ with -
- Some copy improvements


_docs/_latest/configure/scanner/descriptions.md
```diff
@@ -96,7 +96,7 @@ For examples of how to define scanner rules for your forwarding rules, see the
 [`forwarding_rules.yaml`](https://github.com/GoogleCloudPlatform/forseti-security/blob/stable/rules/forwarding_rules.yaml)
 rule file.
 
-## Groups scanner
+## G Suite scanner
```
joecheuk: I think in the yaml file we are using `group` for the scanner and `groups_violations` for the violation, do you think it might confuse user if we change the doc to g suite scanner?

mirons-google: Fair point, we are looking at more than groups though. We inventory: groups, group members, and domain users.

In the future we could have an auditor that looks at user violations and it could be user_violations? It seems renaming groups_violations to gsuite_violations isn't wise or trivial?

_faq/resource/scanners.md
```diff
@@ -21,7 +21,7 @@ For more information, see [scanner descriptions]({% link _docs/latest/configure/
 | Enabled APIs Scanner |
 | Firewall Rules Scanner |
 | Forwarding Rules Scanner |
-| Google Groups Scanner |
+| G Suite Scanner |
```
joecheuk: Same issue as above, maybe Groups scanner makes more sense for now.

_docs/_latest/configure/scanner/descriptions.md
joecheuk: Yes, I agreed it makes more sense to rename it to G Suite. Maybe it's better to do the renaming in the doc after we updated the Forseti code?

_faq/resource/scanners.md
mirons-google: Reverted.

_docs/_latest/configure/scanner/rules.md
```diff
@@ -230,7 +238,7 @@ To enable the blacklist notifier, add the followings to the notifier section in
                 gcs_path: gs://{__YOUR_SCANNER_BUCKET__}/scanner_violations
 ```
 
-## Google Groups rules
+## G Suite rules
```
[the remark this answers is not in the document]
mirons-google: I'll change this to avoid the issue for now, but it needs resolved sooner rather than later.

PR #1681, opened 2018-06-14
Add Log Sinks (Exports) to the Forseti Inventory

dufton:
Adds a client for the Stackdriver Logging API to read log sinks from projects, folders and organizations. The Forseti service account already has permissions to list sinks through the iam.securityReviewer role.

Save sinks as a new resource type.


google/cloud/forseti/services/inventory/base/resources.py
```diff
@@ -1904,6 +1927,46 @@ def iter(self):
...
+
+class FolderSinkIterator(ResourceIterator):
+    """The Resource iterator implementation for Folder Sink"""
+
+    def iter(self):
+        """Yields:
+            Resource: Sink created
+        """
+        gcp = self.client
+        folderid = self.resource['name'].split('/', 1)[-1]
```
joecheuk: Should we put this logic into the get_name method in the foldersinkrepos instead?

dufton: Looks like we don't even need this logic since LoggingFoldersSinksRepository.get_name just adds the prefix back anyway

dufton:
The server service account will not need any extra permissions. It should already have roles/iam.securityReviewer which provides the logging.sinks.list permission

joecheuk:
Thanks for the clarifications!

dufton:
Thanks for your review!

PR #1700, opened 2018-06-20
Added support for using service account instead of a user account

johnrevans6:
Currently, the Forseti Installer doesn't support running the install script using a Service Account. A similar PR [1] was submitted in the past and this PR builds off of it. The previous PR only accounts for the case where gcloud hasn't yet been authenticated. Meaning, if you authenticate with an SA, then it will still fail the domain check. This PR addresses that case by explicitly checking to see if the authenticated user is an SA. If it is, then it sets the is_service_account flag to true and the domain check is skipped.

[1] #1640


johnrevans6:
I'm actually going to close this PR out and start fresh. There's quite a bit of drift between the repo I was basing the fix off of, and the dev branch. I'll submit a new PR that pushes against dev. Thanks, Joe!

Issue #1703, opened 2018-06-21
Model database missing field separator in full_name field in 2.0-rc3


jceresini:
The model table has a field named `full_name`, which is apparently the `parent_type_name` and `type_name` fields concatenated. (Side note, the `type_name` field is the `type` and `name` fields concatenated). The `full_name` field is missing the field separator. Here is an example of some resources in the model database:

```
> select full_name, type_name, parent_type_name, name, type from `4c7bbdfe47888e85289671d594ed8253_resources` where type='network'\G
*************************** 1. row ***************************
       full_name: project/forseti-testxsubnetwork/6917533303373712976/
       type_name: network/6917533303373712976
parent_type_name: project/forseti-testxsub
            name: 6917533303373712976
            type: network
1 row in set (0.0393 sec)
```

```
select full_name, type_name, parent_type_name, name, type from `cb910b2f96d36ce76ef2a6447c96c4da_resources` where type='subnetwork' limit 1\G
*************************** 1. row ***************************
       full_name: project/forseti-testxsubnetwork/1482312637655436839/
       type_name: subnetwork/1482312637655436839
parent_type_name: project/forseti-testx
            name: 1482312637655436839
            type: subnetwork
1 row in set (0.0397 sec)
```

In this case I created 2 projects (forseti-testx, and forseti-testxsub) to demonstrate that this could, in extremely rare cases, create overlapping `full_name` identifiers. However, this looks like its just typo when generating the full_name. That field also has a trailing '/' so im guessing the field separator was just misplaced.

I'm also curious why the database has so much redundant data. Is there a technical reason to store the data like that?

jceresini:
Oh, one other thing I think is worth mentioning. I think it could be valuable to store the name of the api a resource was retrieved from. For example, `compute` for instances, disks, etc. The concern is that google products may end up have overlapping type names across APIs. Possible example could be a `table` type in BigQuery and in BigTable.

joecheuk:
Hi @jceresini , all full name ends with a '/' that's why in the code we have '{parent_full_name}{resource_type_name}/'. I ran the same query on my environment and it seems like I am able to get a proper full name. Is there any reason why the full_names you have don't contain an organization? Forseti intended to run on an Org level and inventory data within that Org.

`
mysql> select full_name, type_name, parent_type_name, name, type from c67f9bf44ca488a1ee452738cb0cca64_resources where type='network' limit 1\G;
*************************** 1. row ***************************
[...]

PR #1704, opened 2018-06-21
Fix typo in generating a resources full_name

jceresini:
Fixing simple typo reported in #1703

### DIFF
--- a/google/cloud/forseti/services/dao.py
+++ b/google/cloud/forseti/services/dao.py
@@ -94,7 +94,7 @@ class Model(MODEL_BASE):
     watchdog_timer_datetime = Column(DateTime())
     created_at_datetime = Column(DateTime())
     etag_seed = Column(String(32), nullable=False)
-    message = Column(Text())
+    message = Column(Text(16777215))
     warnings = Column(Text(16777215))
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


jceresini:
I signed it!

google/cloud/forseti/services/utils.py
```diff
@@ -169,7 +169,7 @@ def to_full_resource_name(full_parent_name, resource_type_name):
         str: full_resource_name of the child
     """
 
-    return '{}{}/'.format(full_parent_name, resource_type_name)
+    return '{}/{}'.format(full_parent_name, resource_type_name)
```
joecheuk: Hi @jceresini , all full name ends with a '/' that's why in the code we have '{parent_full_name}{resource_type_name}/', is there any reason why you want to switch the position?

jceresini: Please take a look at my comments above. I really dont think that trailing slash is intentional, theres nothing separating the parent from the resource. And as I mentioned, this could lead to ambiguous or even overlapping full_name values. 

In the schema currently:
parent_type_name = `[parent_type]`/`[parent_name]`
type_name = `[type]`/`[name]`
full_name = `[parent_type_name][type_name]`/

Note that the last one listed there has no field separator between the 2 values its comprised of.

jceresini: Oh, sorry; my comments are in the issue linked above, not in this thread. #1703

joecheuk: Hi @jceresini , you would be right if we are passing in parent_type_name into this method, but the method is taking in parent_full_name and parent_full_name is a full name so it should end with '/', can you look at my comment in issue #1703 ? We are curious about the environment you installed Forseti rc3 on.

PR #1705, opened 2018-06-21
Fix full_res_name value when the root_resource is a project or a folder

jceresini:
When running inventory with a project or a folder as the root_resource_id (in the server config), the full_res_name is generated incorrectly. This ensures that in both cases the full_res_name has a trailing `/` as is expected when to_full_resource_name is called on any child resources. Please see #1703 for additional information.


joecheuk:
Thanks for fixing this!

joecheuk:
Fixed in #1788

PR #1706, opened 2018-06-21
Added Service Account support for installing Forseti

johnrevans6:
Currently, the Forseti Installer doesn't support running the install script using a Service Account. A similar PR [1] was submitted in the past and this PR builds off of it. The previous PR only accounts for the case where gcloud hasn't yet been authenticated. Meaning, if you authenticate with an SA, then it will still fail the domain check. This PR addresses that case by explicitly checking to see if the authenticated user is an SA. If it is, then it sets the is_service_account flag to true and the domain check is skipped. Finally, the SA is activated so that it can be used in the setup process.

[1] #1640


install/gcp/installer/util/gcloud.py
```diff
@@ -45,11 +47,47 @@ def get_gcloud_info():
...
-    return project_id, authed_user, is_devshell
+    return project_id, authed_user, is_devshell, is_service_account
+
+def _get_service_account_json_path():
+    """Search in the environment variables for Google Credentials
+    Returns:
+        str: The value of the first non-empty environment variable
+    """
+
+    if 'GOOGLE_CREDENTIALS' in os.environ:
```
blueandgold: I am wondering if it would be better to use a FLAG to tell the installer where the key file is, instead of having this method with these logic here?  Such as `--service-account-key-path`.  It would simplify things on our side, with less logic.  Would a flag like this work on your side?

Then above, you can check if this flag is configured, then authed_user is service account.

```
if not service_account_key_path:
    self.check_if_authed_user_in_domain(
        self.organization_id, authed_user)
else:
    gcloud.activate_service_account(authed_user)
```

How do you think?

install/gcp/gcp_installer.py
```diff
@@ -62,6 +62,8 @@ def run():
     parser.add_argument('--no-cloudshell',
                         action='store_true',
                         help='Bypass Cloud Shell requirement')
+    parser.add_argument('--service-account-key-path',
+                        help='Absolute path for service account key file')
```
blueandgold: Can you please add into the help, that the filename is also needed?  Something like:
```
help='Absolute path and filename for service account key file'
```

install/gcp/installer/configs/config.py
```diff
@@ -34,6 +34,8 @@ def __init__(self, **kwargs):
                               datetime.datetime.now().strftime('%Y%m%d%H%M%S'))
         self.identifier = None
         self.force_no_cloudshell = bool(kwargs.get('no_cloudshell'))
+        self.service_account_key_path = kwargs.get(
+            'service_account_key_path') or None
```
blueandgold: Don't need the `or None` here as the flag will already default to None.  Also, `some_dict.get('foo')` will also default to `None`.

install/gcp/installer/util/gcloud.py
```diff
@@ -30,6 +30,7 @@ def get_gcloud_info():
         str: GCP project id
         str: GCP Authenticated user
         bool: Whether or not the installer is running in cloudshell
+        bool: Whether or not authenticated user is a service account
```
blueandgold: Remove as this is not needed anymore.

install/gcp/installer/util/gcloud.py
```diff
@@ -51,6 +52,23 @@ def get_gcloud_info():
             sys.exit(1)
     return project_id, authed_user, is_devshell
 
+def activate_service_account(authed_user, key_path):
+    """Activate the service account with gcloud
```
blueandgold: Add a period at the end of this sentence.  This will be in the generated pydoc that we put online.

install/gcp/installer/util/gcloud.py
```diff
@@ -51,6 +52,23 @@ def get_gcloud_info():
...
+    Args:
```
blueandgold: Add a new blank line here.

install/gcp/installer/util/gcloud.py
```diff
@@ -51,6 +52,23 @@ def get_gcloud_info():
...
+        authed_user (str): service account email
+        key_path (str): absolute path to service account key file
+    """
+
+    return_code, _, err = utils.run_command(
+        ['gcloud', 'auth', 'activate-service-account',
+         authed_user, '--key-file=' + key_path])
```
blueandgold: In the context of this method, you should just call the authed_user to be `service_account`.

install/gcp/installer/util/gcloud.py
```diff
@@ -51,6 +52,23 @@ def get_gcloud_info():
             sys.exit(1)
     return project_id, authed_user, is_devshell
 
+def activate_service_account(authed_user, key_path):
+    """Activate the service account with gcloud
+    Args:
+        authed_user (str): service account email
```
blueandgold: Capitalize and add period.
```
Service account email.
```

install/gcp/installer/util/gcloud.py
```diff
@@ -51,6 +52,23 @@ def get_gcloud_info():
...
+        key_path (str): absolute path to service account key file
```
blueandgold: Capitalize and add period.  And add filename.
```
Absolute path and filename to service account key file.
```

install/gcp/gcp_installer.py
johnrevans6: Done

install/gcp/installer/configs/config.py
johnrevans6: Done

install/gcp/installer/util/gcloud.py
johnrevans6: Done

install/gcp/installer/util/gcloud.py
johnrevans6: Done

johnrevans6:
Thanks for the feedback! I've made the changes you've requested, and I'll check back in the morning to see if there's any outstanding work that needs to be done (ie linter errors, etc).

install/gcp/installer/util/gcloud.py
```diff
@@ -51,6 +51,24 @@ def get_gcloud_info():
...
+    """Activate the service account with gcloud.
+
+    Args:
+        service_account (str): Service account email
+        key_path (str): Absolute path to service account key file
+    """
+
+    return_code, _, err = utils.run_command(
+        ['gcloud', 'auth', 'activate-service-account',
+         service_account, '--key-file=' + key_path])
```
blueandgold: Now that I'm looking at this `key_path` naming with the gcloud command's naming, should would align them?
i.e. change `key_path` to `key_file` everywhere, including the flag?  How do you think?
```
'--key-file=' + key_file
```

install/gcp/gcp_installer.py
```diff
@@ -62,6 +62,9 @@ def run():
     parser.add_argument('--no-cloudshell',
                         action='store_true',
                         help='Bypass Cloud Shell requirement')
+    parser.add_argument('--service-account-key-path',
```
blueandgold: Please see my comment below about changing this to `--service-account-key-file`.

install/gcp/installer/util/gcloud.py
johnrevans6: Done

install/gcp/gcp_installer.py
johnrevans6: Done

PR #1762, opened 2018-07-05
Run cloudsqlproxy service as root

### LINKED ISSUE #1761: cloudsqlproxy service failed to start
Due to insufficient permissions

Error message:

> google: error getting credentials using well-known file (/home/ubuntu/.config/gcloud/application_default_credentials.json): open /home/ubuntu/.config/gcloud/application_default_credentials.json: permission denied

joecheuk:
Removed User= from cloudsqlproxy service.

By default, Systemd starts the executable stated in ExecStart= as root so it will contain the permissions required to run the cloudsqlproxy script.

Resolves #1761

### DIFF
--- a/install/gcp/scripts/initialize_forseti_services.sh
+++ b/install/gcp/scripts/initialize_forseti_services.sh
@@ -66,7 +66,6 @@ SQL_PROXY_SERVICE="$(cat << EOF
 [Unit]
 Description=Cloud SQL Proxy
 [Service]
-User=ubuntu
 Restart=always
 RestartSec=3
 ExecStart=$SQL_PROXY_COMMAND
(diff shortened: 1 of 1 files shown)


install/gcp/scripts/initialize_forseti_services.sh
```diff
@@ -66,7 +66,6 @@ SQL_PROXY_SERVICE="$(cat << EOF
 [Unit]
 Description=Cloud SQL Proxy
 [Service]
-User=ubuntu
```
blueandgold: Can you please just add a comment here:
```
# See issue #1761 for why this needs to be run as root.
```

joecheuk: Updated.

PR #1769, opened 2018-07-09
Added 2.0 upgrade instruction

joecheuk:


_docs/_latest/setup/upgrade.md
```diff
@@ -104,6 +104,56 @@ To learn more about these fields, see [Configure]({% link _docs/latest/configure
 {% endcapture %}
 {% include site/zippy/item.html title="Upgrading 1.X installations" content=1x_upgrade uid=0 %}
 
+{% capture 2x_upgrade %}
+
+## Upgrade Forseti v2 to a newer version using deployment manager.
+
+Note: If you used the Forseti setup wizard to deploy, You can locate the deployment template in your
```
blueandgold: Instead of `setup wizard`, I think we are calling it `installer` now.

joecheuk: Updated.

joecheuk:
Thanks for the heads up, for this update, we want v2.0.1 and later versions to contain the changes.

PR #1770, opened 2018-07-10
Allow data model to be built without an organization root.

blueandgold:
This PR allows the data model to be built without an organization root, so that scanners can have the data they needed for scanning.  This is especially useful, when the root is a folder.

However, this means that Explain will not work fully & properly without the whole organization.  So after discuss with team, implementing a check to disable Explain functionalities, if we are not using the organization root.

image

@ahoying @joecheuk I will add test coverage if this looks good.  One thing I am not too thrilled about, is the code repetition on the `is_supported` checks.  We discussed not using decorators, so if you can think of anything cooler to do here, please let me know.


google/cloud/forseti/services/explain/service.py
```diff
@@ -79,10 +97,16 @@ def ListResources(self, request, context):
         Returns:
             object: proto message of list of resources
         """
+        reply = explain_pb2.ListResourcesReply()
+
+        if not self.is_supported:
```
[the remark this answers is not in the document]
ahoying: The method must return the exact Message type specified in the Service configuration:

https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/services/explain/explain.proto#L19

You can't create a new message type just for this error, it has to return an empty proto of the type specified.

google/cloud/forseti/services/explain/service.py
```diff
@@ -79,10 +97,16 @@ def ListResources(self, request, context):
...
+            context.set_code(StatusCode.FAILED_PRECONDITION)
```
ahoying: You could create a small private helper method and use that.

```
def _set_not_supported_status(context, reply):
  context.set_code(..)
  context.set_details(..)
  return reply
```

Then in you can just do:

```
if not self.is_supported:
  return _set_not_supported_status(context, reply)
```

google/cloud/forseti/services/explain/service.py
```diff
@@ -47,6 +52,18 @@ def _get_handle(self, context):
...
 
+    def _determine_is_supported(self):
+        """Determine whether Explainer is supported for clients to use.
+
+        Returns:
+            bool: True if Explainer module can be used.
+        """
+        root_resource_id = (
+            self.explainer.config.inventory_config.root_resource_id)
+        if 'organizations' in root_resource_id:
```
ahoying: Is this the only thing that needs to be checked? What about if dwd is not configured?

google/cloud/forseti/services/explain/service.py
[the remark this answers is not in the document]
joecheuk: Thanks for the clarifications!

tests/services/explain/service_test.py
```diff
@@ -0,0 +1,54 @@
...
+        self.grpc_explainer = service.GrpcExplainer(mock_explainer_api)
+
+    def tearDown(self):
+        """Tear down method."""
+        ForsetiTestCase.tearDown(self)
+
+    def test_determine_is_supported(self):
+        inventory_config = self.grpc_explainer.explainer.config.inventory_config
+        inventory_config.root_resource_id = 'organizations/11111'
+        self.assertTrue(self.grpc_explainer._determine_is_supported())
```
ahoying: Nit, make this two different tests so on assert failing doesn't prevent the second assert from getting run.

tests/services/explain/service_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+        inventory_config.root_resource_id = 'folders/22222'
+        self.assertFalse(self.grpc_explainer._determine_is_supported())
+
+    def test_failed_precondition(self):
```
ahoying: I would love to see a test like https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/tests/services/api_tests/model_test.py#L134 for a functional integration test of the service when is_supported is false.

blueandgold: Done, good tip.

tests/services/explain/explain_test.py
```diff
@@ -0,0 +1,113 @@
...
+            GrpcExplainerFactory,
+            GrpcInventoryFactory,
+            GrpcModellerFactory,
+        ])
+
+
+class ExplainerTest(ForsetiTestCase):
+    """Test based on declarative model."""
+
+    def setUp(self):
```
ahoying: Nit: No need to have setUp here if it isn't used.

tests/services/explain/explain_test.py
```diff
@@ -0,0 +1,113 @@
...
+        pass
+
+    def test_explain_is_supported(self):
+        """Test explain is supported."""
+
+        inventory_config_with_organization_root = (
+            InventoryConfig(gcp_api_mocks.ORGANIZATION_ID, '', {}, '')                                                   )
+        self.setup = create_tester(
```
ahoying: Nit: you don't need to assign this variable to self, you could just create a local variable instead (same below).

tests/services/explain/explain_test.py
```diff
@@ -0,0 +1,113 @@
...
+        """Test explain is not supported."""
+
+        inventory_config_with_folder_root = (
+            InventoryConfig(gcp_api_mocks.FOLDER_ID, '', {}, '')                                                   )
+        self.setup = create_tester(inventory_config_with_folder_root)
+
+        def test(client):
+            """Test implementation with API client."""
+            try:
+                client.explain.list_roles('')
```
ahoying: Nit: you could use self.assertRaisesRegex here instead. https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp.

tests/services/explain/explain_test.py
blueandgold: Done.

tests/services/explain/explain_test.py
blueandgold: Done

tests/services/explain/explain_test.py
blueandgold: Great tip, didn't know about this before.

PR #1784, opened 2018-07-11
Removed old documentation

joecheuk:


joecheuk:
This change is not needed.

PR #1811, opened 2018-07-23
Adding documentation for creation of security keys

EricBeach:


_includes/docs/latest/deployment-prerequisites.md
```diff
@@ -31,7 +31,7 @@ Scanner, and Enforcer, follow the steps below:
 
   1. Go to the [GCP Console](https://console.cloud.google.com/iam-admin/serviceaccounts)
   and create a new service account.
-  1. Create and download a JSON key for the service account.
+  1. Create and download a JSON key for the service account [see docs](https://cloud.google.com/iam/docs/creating-managing-service-account-keys).
```
blueandgold: At this point, we already directed the user to the Cloud Console on the UI.  So, it would flow better if we tell them specifically where on the UI to create the link.  i.e using the `Actions` button (the 3 vertical dots).

```
Click the More Actions button (3 vertical dots), to create and download a JSON key for the service account.
```

PR #1823, opened 2018-07-27
Documenting solution to virtualenv issues on clean Linux install

EricBeach:


_docs/_latest/develop/dev/setup.md
```diff
@@ -136,6 +136,14 @@ following:
 In the above example, `/SOME/PATH/TO` represents the path specific to your
 system. Make sure to use the values from your terminal.
 
+------------------
+
+If on Linux executing ``mkvirtualenv forseti-security`` results in
+``bash: mkvirtualenv: command not found`` then try the following:
+
+(1) ``sudo pip install virtualenv virtualenvwrapper``
+(2) ``sudo pip install --upgrade pip``
```
blueandgold: qq: Is there a reason why pip needs to be upgraded afterwards?  Is step 1 by itself sufficient to resolve mkvirtualenv not found?

EricBeach: Good question; I don't know the answer; after banging on this one for probably four hours and tracking down different possible answers, this is the one that worked for me. I did both together, but I can try just adding the first one.

PR #1824, opened 2018-07-27
Updated installation instructions to include steps to install server instance only

### LINKED ISSUE #1821: Add documentation on installing server only.
User input: only the direct security admin team is using forseti, and so they don't need to use, install, or pay for the client VM. The server install only option was not obvious to them, and we should highlight that we have this option in our documentation.

joecheuk:
Resolves #1821

### DIFF
--- a/_docs/v2.1/setup/install.md
+++ b/_docs/v2.1/setup/install.md
@@ -67,6 +67,12 @@ steps below:
      python install/gcp_installer.py
      ```
 
+     If you don't plan to share your Forseti instance with other people, 
+     you can choose to install the server instance only and access Forseti from there.
+     ```bash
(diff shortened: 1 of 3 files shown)


_docs/v2.1/setup/install.md
```diff
@@ -67,6 +67,12 @@ steps below:
      python install/gcp_installer.py
      ```
 
+     If you don't plan to share your Forseti instance with other people,
```
blueandgold: We should try to clarify "people".  Either something like `non-admin` or "If forseti users already have the same permission as the server".

joecheuk: Updated.

PR #1833, opened 2018-07-31
Added retry and fall back when google.auth.default() fails

### LINKED ISSUE #1832: Forseti inventory stopped running after a day or two due to auth problem
Error message:

> Error occurred on the server side, message: <_Rendezvous of RPC that terminated with:
>         status = StatusCode.UNKNOWN
>         details = "Exception iterating responses: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see https://developers.google.com/accounts/docs/application-default-credentials."
>         debug_error_string = "{"created":"@1533055140.251333959","description":"Error received from peer","file":"src/core/lib/surface/call.cc","file_line":1083,"grpc_message":"Exception iterating responses: Could not automatically determine credentials. Please set GOOGLE_APPLICATION_CREDENTIALS or explicitly create credentials and re-run the application. For more information, please see https://developers.google.com/accounts/docs/application-default-credentials.","grpc_status":2}"
> >
>

joecheuk:
Credentials will be None if google.auth.default() still failed after 5 trials and the inventory should proceed with other resources.

Resolves #1832

### DIFF
--- a/google/cloud/forseti/common/gcp_api/_base_repository.py
+++ b/google/cloud/forseti/common/gcp_api/_base_repository.py
@@ -28,6 +28,7 @@ import google.auth
 from google.auth.credentials import with_scopes_if_required
 
 from google.cloud import forseti as forseti_security
+from google.cloud.forseti.common.gcp_api import api_helpers
 from google.cloud.forseti.common.gcp_api import _supported_apis
 from google.cloud.forseti.common.gcp_api import errors as api_errors
 from google.cloud.forseti.common.util import logger
@@ -175,7 +176,7 @@ class BaseRepositoryClient(object):
         if not credentials:
             # Only share the http object when using the default credentials.
             self._use_cached_http = True
-            credentials, _ = google.auth.default()
+            credentials, _ = api_helpers.get_google_default_credentials()
         self._credentials = with_scopes_if_required(credentials,
                                                     list(CLOUD_SCOPES))
 
--- a/google/cloud/forseti/common/gcp_api/_base_repository.py
+++ b/google/cloud/forseti/common/gcp_api/_base_repository.py
@@ -170,10 +170,8 @@ class BaseRepositoryClient(object):
                 limiter for this service.
             **kwargs (dict): Additional args such as version.
         """
-        self._use_cached_http = False
         if not credentials:
             # Only share the http object when using the default credentials.
-            self._use_cached_http = True
             credentials, _ = api_helpers.get_google_default_credentials()
 
         # Lock may be acquired multiple times in the same thread.
@@ -252,8 +250,7 @@ class BaseRepositoryClient(object):
         with self._repository_lock:
             return repository_class(gcp_service=self.gcp_services[version],
                                     credentials=self._credentials,
-                                    rate_limiter=self._rate_limiter,
-                                    use_cached_http=self._use_cached_http)
+                                    rate_limiter=self._rate_limiter)
 # pylint: enable=too-many-instance-attributes
 
 # pylint: disable=too-many-instance-attributes, too-many-arguments
@@ -264,7 +261,7 @@ class GCPRepository(object):
                  num_retries=NUM_HTTP_RETRIES, key_field='project',
                  entity_field=None, list_key_field=None, get_key_field=None,
                  max_results_field='maxResults', search_query_field='query',
-                 rate_limiter=None, use_cached_http=True):
+                 rate_limiter=None):
         """Constructor.
 
         Args:
@@ -289,9 +286,6 @@ class GCPRepository(object):
             search_query_field (str): The field name used to filter search
                 results.
             rate_limiter (object): A RateLimiter object to manage API quota.
-            use_cached_http (bool): If set to true, calls to the API will use
-                a thread local shared http object. When false a new http object
-                is used for each request.
         """
         self.gcp_service = gcp_service
         self._credentials = credentials
@@ -314,8 +308,6 @@ class GCPRepository(object):
         self._max_results_field = max_results_field
         self._search_query_field = search_query_field
         self._rate_limiter = rate_limiter
-
-        self._use_cached_http = use_cached_http
         self._local = LOCAL_THREAD
 
     @property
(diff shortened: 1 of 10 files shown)


google/cloud/forseti/services/inventory/inventory.py
```diff
@@ -255,7 +255,7 @@ def do_inventory():
                     return result.get_summary()
 
                 except Exception as e:
-                    LOGGER.exception(e)
+                    LOGGER.error(e, exc_info=True)
```
blueandgold: According to the answer with the 77 votes, LOGGER.exception(e) is the same as LOGGER.error(e, exc_info=True).  So, please verify why this change would actually help to catch the error.

If that is true, then I would prefer LOGGER.exception because it is more readable, and it allows you to pass in a customer message and still show the stacktrace.

https://stackoverflow.com/questions/5191830/how-do-i-log-a-python-error-with-debug-information

joecheuk: Verified exception() is same as error() with exc_info=True, updated it back to exception.

https://github.com/python/cpython/blob/ea68d83933e6de6cabfb115ec1b8888301947369/Lib/logging/__init__.py#L1367

google/cloud/forseti/common/gcp_api/_base_repository.py
```diff
@@ -503,4 +522,6 @@ def _execute(self, request):
                                        num_retries=self._num_retries)
         return request.execute(http=self.http,
                                num_retries=self._num_retries)
```
blueandgold: nit: remove one of the blank lines

google/cloud/forseti/common/gcp_api/_base_repository.py
```diff
@@ -488,6 +488,25 @@ def execute_query(self, verb, verb_arguments):
...
+            dict: The response from the API.
+        """
+        try:
+            return self._execute_request(request)
+        except exceptions.RefreshError as e:
+            # If there is problem refreshing the token, we will recreate a new
+            # Credential object and use that to refresh the request.
+            LOGGER.exception(e)
+            self._credentials, _ = api_helpers.get_google_default_credentials()
+            self._credentials.refresh(request)
```
blueandgold: It seems like we need to set the `CLOUD_SCOPES` here.  Maybe centralize it in the `api_helpers.get_google_default_credentials()` everywhere?

blueandgold: The usage of `refresh()` is a bit inconsistent.  I see that in the `__init__()` of `BaseRepositoryClient` class that we don't use the `refresh()` is not being used.  Why is that?

And if we really need, how about centralizing it in the `get_google_default_credentials()` as well?

google/cloud/forseti/common/gcp_api/_base_repository.py
```diff
@@ -488,6 +488,25 @@ def execute_query(self, verb, verb_arguments):
...
+        Returns:
+            dict: The response from the API.
+        """
+        try:
+            return self._execute_request(request)
+        except exceptions.RefreshError as e:
+            # If there is problem refreshing the token, we will recreate a new
+            # Credential object and use that to refresh the request.
+            LOGGER.exception(e)
+            self._credentials, _ = api_helpers.get_google_default_credentials()
```
ahoying: I'm not sure this is enough. The credential is used to create an authenticated http object in https://github.com/GoogleCloudPlatform/forseti-security/pull/1833/files#diff-ad7b8ba466c514dddccaa5f11b5a3d5aR325.

The http object is already included in the request.

The flow is:

BaseRespositoryClient -> Create credential -> Initialize GCPRepository instances for the API -> Create a thread local cached http object shared by all GCPRepository instances across all APIs (except for gsuite) -> use authenticated http object to call API methods.

Each thread has its own cached http object, which has a self.credentials object that is a pointer to the original credentials, it won't see the replaced credentials.

You could use delattr(self._local, 'http'), although that will only fix the current thread and not any other running threads.

Speaking of threads, you should not replace self._credentials without getting a lock to ensure it's not actively being used in another thread.

google/cloud/forseti/common/gcp_api/_base_repository.py
joecheuk: Updated

joecheuk: refresh will take in a request and we don't have the request object in the __init__ method inside the BaseRepositoryClient class, the request is built with method _build_request() in the GCPRepository class

google/cloud/forseti/common/gcp_api/_base_repository.py
joecheuk: Updated

google/cloud/forseti/common/gcp_api/_base_repository.py
ahoying: Looking through https://github.com/GoogleCloudPlatform/google-auth-library-python-httplib2/blob/master/google_auth_httplib2.py#L178 and https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/compute_engine/credentials.py#L35, I don't think it's the credentials that needs to be replaced, I think it's the local http object that needs to be replaced.

In fact, with google auth, I wonder if it even makes sense to do the caching any more, that was created because of oauth2client. You could try to disable the caching in self.http so it always returns a new AuthorizedHttp object.

ahoying: Actually, it looks like https://github.com/GoogleCloudPlatform/google-auth-library-python/blob/master/google/auth/compute_engine/credentials.py#L35 is not thread safe. If multiple threads call refresh at the same time, it could corrupt the object, that may be the root cause of our issue.

ahoying: That would explain the 400 response from the metadata server.

joecheuk:
The issue is addressed in #2091

PR #1836, opened 2018-08-01
update stacktrace in broad excepts issue#1797

kssrini:
updated stacktrace in broad excepts issue#1797


blueandgold:
Hi Srini, thanks for doing this!

However, there are a few more places with the broad excepts.  The simplest way to find them everywhere is by doing a global search in your IDE, using the `except Exception` phrase.  Then you will see more results like below.

There are a few more items you need to do:

1: if you see `except Exception:`, then change them to `except Exception as e:`

2: if you do not see any logging in the except block of code, then add `Logger.exception(e)`.

image

google/cloud/forseti/services/dao.py
```diff
@@ -571,7 +571,8 @@ def denorm_group_in_group(cls, session):
 
                     rows_affected = bool(session.execute(qry).rowcount)
                     iterations += 1
-            except Exception:
+            except Exception as e:
+                LOGGER.exception(e)
                 LOGGER.error(Exception.message)
```
blueandgold: We can delete this line now:
```
LOGGER.error(Exception.message)
```

google/cloud/forseti/services/inventory/crawler.py
```diff
@@ -22,6 +22,10 @@
 from google.cloud.forseti.services.inventory.base import crawler
 from google.cloud.forseti.services.inventory.base import gcp
 from google.cloud.forseti.services.inventory.base import resources
+from google.cloud.forseti.common.util import logger
```
blueandgold: Please sort this import line to be alphabetical order.

google/cloud/forseti/services/inventory/crawler.py
```diff
@@ -21,9 +21,13 @@
 
 from google.cloud.forseti.services.inventory.base import crawler
 from google.cloud.forseti.services.inventory.base import gcp
+from google.cloud.forseti.common.util import logger
```
blueandgold: This import line should be above line 22.
i.e. above:
```
from google.cloud.forseti.services.inventory.base import crawler
```

PR #1839, opened 2018-08-01
Add Billing Account log sinks to Inventory

dufton:
Add Billing Accounts and their log sinks to the Forseti Inventory.

The required permissions on the billing API are already granted to the Forseti service account if it has the iam.securityReviewer role at the Org level.


joecheuk:
Hi @dufton , a few of the unit tests failed after the update, would you mind updating those tests before merging in the PR?

PR #1846, opened 2018-08-03
[Fixes 1816] Adding direct deep-links into Google Cloud Shell that also clones a git repo.

### LINKED ISSUE #1816: Update the website to take advantage of the git_repo param
Our docs could be simplified by offering a link to open a cloud shell and clone the repo right away with a URL like the one below.

https://console.cloud.google.com/cloudshell/open?git_repo=https%3A%2F%2Fgithub.com%2FGoogleCloudPlatform%2Fforseti-security&page=editor

EricBeach:
Fixes #1816

### DIFF
--- a/_docs/_latest/setup/install.md
+++ b/_docs/_latest/setup/install.md
@@ -33,31 +33,20 @@ template, and creates a Forseti deployment.
 It's best to use
 [Cloud Shell](https://cloud.google.com/shell/docs/quickstart) to run the Forseti
 installer. This ensures you're using the latest version of Cloud SDK since it's
-included in Cloud Shell. To prepare to run the Forseti setup wizard, follow the
-steps below:
-
-  1. Go to the [GCP Console](https://console.cloud.google.com/).
-  1. In the **Select a project** drop-down list at the top of the console,
-     select the project you want to deploy Forseti to.
-  1. On the top right, click the icon to **Activate Google Cloud Shell**. The
-     Cloud Shell panel opens at the bottom of the page.
+included in Cloud Shell.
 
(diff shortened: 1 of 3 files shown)


_docs/_latest/setup/install.md
```diff
@@ -33,31 +33,20 @@ template, and creates a Forseti deployment.
...
 
 ### Run setup
 
-  1. After you've started Cloud Shell, download Forseti. The installer is
-  included in the `install/` directory:
-
-      ```bash
-      git clone https://github.com/GoogleCloudPlatform/forseti-security.git
-      ```
+  1. [Open Google Cloud Shell and download Forseti code](https://console.cloud.google.com/cloudshell/open?git_repo=https%3A%2F%2Fgithub.com%2FGoogleCloudPlatform%2Fforseti-security&page=editor)
```
blueandgold: This will cause a complication, in that the project will not be set (see screenshots below).

IOW: there was a very specific reason why we had the step above to select the project first, and then open the cloudshell.  Because that will ensure cloudshell is activated on the right project that will host forseti.

Or is there a way to update the URL to set the project?

image

image

EricBeach: There would be no way to include the desired project ahead of time in the URL since I don't know what each user's project will be. Maybe you and @mirons-google should chat about desired functionality here. If you want the project to be preset when the user loads Cloud Shell, I don't think we can use this feature. I'm happy to drop this PR or you/Mike can reach agreement about what to do here.

EricBeach:
After talking with @blueandgold , we're going to close this and the underlying Issue out.

PR #1851, opened 2018-08-06
Adding solution for error repeatedly encountered in forseti setup

EricBeach:
"bash: workon: command not found" and its solution on Linux machine.

### DIFF
--- a/_docs/_latest/develop/dev/setup.md
+++ b/_docs/_latest/develop/dev/setup.md
@@ -153,6 +153,14 @@ try the following:
 
 ``sudo apt install default-libmysqlclient-dev``
 
+------------------
+
+If on Linux executing ``workon forseti-security``
+results in ``bash: workon: command not found`` then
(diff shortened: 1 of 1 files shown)


_docs/_latest/develop/dev/setup.md
```diff
@@ -153,6 +153,14 @@ try the following:
 
 ``sudo apt install default-libmysqlclient-dev``
 
+------------------
+
+If on Linux executing ``workon forseti-security``
+results in ``bash: workon: command not found`` then
+check whether workon is in the source path. Try:
```
blueandgold: The `source` command actually runs the `virtualenvwrapper.sh` in the current shell.  So, it would be better to reword `check`, to something like:

```
....then ensure workon is in the source path
```

EricBeach:
Made a tweak to wording. Let me know what you think.

EricBeach:
I did squash and merge, but 3 commits still came in. Odd.

PR #1855, opened 2018-08-07
Noting process for running unit tests in Docker after making changes

EricBeach:


_docs/_latest/develop/dev/testing.md
```diff
@@ -59,3 +59,18 @@ If you want to execute a subset of tests (e.g., all tests in a specific subdirec
     ```bash
     docker -l error exec -it build /bin/bash -c "python -m unittest discover -s tests/common -p '*_test.py'"
     ```
+
+### Continual Testing
```
blueandgold: This section heading can be more clear, with something along the line `Update Code and Re-Run Test`.

blueandgold: qq: I do not use docker for my testings.  So, once the tests are run in docker, is the docker container left running and not teared down?  If that's the case, we should document how to tear down.

_docs/_latest/develop/dev/testing.md
```diff
@@ -59,3 +59,18 @@ If you want to execute a subset of tests (e.g., all tests in a specific subdirec
...
+This copy is static and not dynamic, so if you make changes to your local code and
+want to run the tests in Docker with your local changes, you need to do the following:
+
+    ```bash
+    # Find the Container ID for the Image "forseti/build"
+    docker ps
+    
+    # Copy the files from the local disk to the image
+    # Replace "e4d76477e7d0" with the Container ID from above
+    docker cp ./ e4d76477e7d0:/forseti-security/
```
blueandgold: Instead of using `e4d76477e7d0`, we can do a self-documenting place holder, and drop the above comment.
```
docker cp ./ <forseti/build_container_id>:/forseti-security/
```

Issue #109, opened 2017-03-09
Handle SQLAlchemy Exceptions in the DAO When Schema Changes


blueandgold:
For instance in g/c/f/services/scanner/dao.py

https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/google/cloud/forseti/services/scanner/dao.py#L138

https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/google/cloud/forseti/services/scanner/dao.py#L154

carise:
#228 must be done before thinking about how to handle db migrations.

joecheuk:
Relates #1828

mirons-google:
@joecheuk can this be closed?

joecheuk:
@mirons-google Yes, this can be closed as we have ways to handle schema updates now.

PR #1864, opened 2018-08-07
Added resource_name column to all the scanners, added schema update handles in scanner dao

### LINKED ISSUE #1828: Add resource_name column in the violation reports
Resource_id is too hard to read but is useful to identify the resource in the gcp_inventory table. We should add another resource_name column to contain the human readable names.

related to: #1800

joecheuk:
Resolves #1828 , #109


google/cloud/forseti/services/scanner/dao.py
```diff
@@ -377,4 +408,30 @@ def initialize(engine):
...
+    tables = BASE.metadata.tables
+
+    schema_update_method_name = 'schema_update'
+
+    for subclass in base_subclasses:
+        schema_update = getattr(subclass, schema_update_method_name, None)
+        if callable(schema_update) and subclass.__tablename__ in tables:
+            LOGGER.info('Updating table %s', subclass.__tablename__)
+            # schema_update will require the Table object.
+            schema_update(tables.get(subclass.__tablename__))
```
blueandgold: As discussed, it would be awesome if the running of the schema update can be moved outside of the forseti application, so that the db update operation doesn't have to run everytime the forseti application starts up, and users can have the chance to backup their database before their db schema is actually updated.  This will also require a documentation change to our upgrade instruction as well.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -183,6 +185,15 @@ def __repr__(self):
...
             self.violation_type, self.resource_type, self.rule_name)
 
+    def schema_update(table):
+        """Maintain all the schema changes for this table.
+
+        Args:
+            table (Table): The table object of this class.
+        """
+        col = Column('resource_name', String(256), default='')
+        create_column(table, col)
```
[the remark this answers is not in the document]
joecheuk: Added error handling in db_migrator.py.

install/gcp/upgrade_tools/db_migrate.py
```diff
@@ -0,0 +1,62 @@
+import sys
```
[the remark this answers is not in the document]
joecheuk: Renamed to db_migrator.py

deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -193,6 +193,11 @@ def GenerateConfig(context):
 echo "Starting services."
 systemctl start cloudsqlproxy
 sleep 5
+
+echo "Attempting to upgrade database."
+python $USER_HOME/forseti-security/install/gcp/upgrade_tools/db_migrator.py
+sleep 5
```
blueandgold: Why `sleep 5`.... can it take more than 5 seconds?

deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -193,6 +193,11 @@ def GenerateConfig(context):
 echo "Starting services."
 systemctl start cloudsqlproxy
 sleep 5
+
+echo "Attempting to upgrade database."
```
blueandgold: I would be clear that this updating the `schema`, and add `if necessary` at the end.

```
echo "Attempting to update database schema. if necessary."
```

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,71 @@
...
+        if callable(schema_update) and subclass.__tablename__ in tables:
+            LOGGER.info('Updating table %s', subclass.__tablename__)
+            # schema_update will require the Table object.
+            try:
+                schema_update(tables.get(subclass.__tablename__))
+            except OperationalError:
+                LOGGER.info('Failed to update db schema, table=%s',
+                            subclass.__tablename__)
+            except Exception as e:
+                LOGGER.error(e)
```
blueandgold: How about using LOGGER.exception() here?

LOGGER.exception('Unexpected error happened when attempting to update database schema.')

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,71 @@
...
+
+    for subclass in base_subclasses:
+        schema_update = getattr(subclass, schema_update_method_name, None)
+        if callable(schema_update) and subclass.__tablename__ in tables:
+            LOGGER.info('Updating table %s', subclass.__tablename__)
+            # schema_update will require the Table object.
+            try:
+                schema_update(tables.get(subclass.__tablename__))
+            except OperationalError:
+                LOGGER.info('Failed to update db schema, table=%s',
```
blueandgold: This should be LOGGER.error(), rather than info()

If this is `OperationalError`, can we be more specific in the log message what might have happened?

install/gcp/upgrade_tools/db_migrator.py
blueandgold: Should we also add the table name here?  Or is that already in the stacktrace?

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -183,6 +188,16 @@ def __repr__(self):
         return string.format(
             self.violation_type, self.resource_type, self.rule_name)
 
+    @staticmethod
+    def schema_update(table):
```
blueandgold: Method name should start with verb first.  So, `update_schema()`, which is also consistent with `migrate_schema()`.  Or is there a reason to call this `schema_update()`?

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,71 @@
...
+    # If the DB connection string is passed in, use that, otherwise
+    # fall back to the default DB connection string.
+    print (sys.argv)
+    DB_CONN_STR = sys.argv[1] if len(sys.argv) > 1 else DEFAULT_DB_CONN_STR
+
+    sql_engine = general_dao.create_engine(DB_CONN_STR,
+                                           pool_recycle=3600)
+
+    # Upgrade Scanner tables.
+    migrate_schema(sql_engine, scanner_dao.BASE)
```
blueandgold: Can we do this?  Which will make it easier to add more tables in the future?

```
declaritive_bases = [scanner_dao.BASE, inventory_dao.BASE]

for base in declaritive_bases:
    migrate_schema(sql_engine, base)
```

deployment-templates/compute-engine/server/forseti-instance-server.py
joecheuk: Updated.

deployment-templates/compute-engine/server/forseti-instance-server.py
joecheuk: The migrator script should block until the upgrade is finished, removed the sleep statement.

google/cloud/forseti/services/scanner/dao.py
joecheuk: Renamed to update_schema

install/gcp/upgrade_tools/db_migrator.py
joecheuk: This is expected when the table already has that column, I think info makes more sense so user won't need to worry about seeing the message at error level. Added log message in the update_schema method so we know what is failing.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: Updated.

joecheuk: Updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: Updated.

joecheuk:
Hi @ahoying , can you take a look at this PR again? This would be the final version if no new bugs were found.

deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -193,6 +193,10 @@ def GenerateConfig(context):
 echo "Starting services."
 systemctl start cloudsqlproxy
 sleep 5
+
+echo "Attempting to update database schema. if necessary."
```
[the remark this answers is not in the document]
joecheuk: Thanks for pointing that out, that should definitely be a comma, updated!

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -183,6 +188,18 @@ def __repr__(self):
...
             self.violation_type, self.resource_type, self.rule_name)
 
+    @staticmethod
+    def update_schema(table):
+        """Maintain all the schema changes for this table.
+
+        Args:
+            table (Table): The table object of this class.
+        """
+        col_name = 'resource_name'
```
blueandgold: qq, in the future, as we have more columns to add, how would they be handled?  This time, we want to add `resource_name`, next time we want to add `xyz`. It's hard to see how `xyz` would fit here.

install/gcp/upgrade_tools/db_migrator.py
blueandgold: There is no log message in the `update_schema()`

google/cloud/forseti/services/scanner/dao.py
joecheuk: You will add the column as an attribute to the class and add the column creation logic in this method, 

i.e.
class Violation(BASE):
      xyz = Column(...)

      def update_schema(table):
             col = Column('resource_name', String(256), default='')
             col.create(table, populate_default=True)

             col = Column('xyz', String(256), default='')
             col.create(table, populate_default=True)

install/gcp/upgrade_tools/db_migrator.py
joecheuk: There is an info log in the update_schema method - LOGGER.info('Attempting to create column: %s', col_name)

google/cloud/forseti/services/scanner/dao.py
blueandgold: Right, that's what I thought.  I think you should make `col_name` into a list of tuples `[(name, type, default), ....]`

And then just loop?

Or else just hardcode the name into the Column(), instead of having a variable `col_name` to track it.

joecheuk: Updated to hard code the column name

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -183,6 +188,17 @@ def __repr__(self):
...
+    @staticmethod
+    def update_schema(table):
+        """Maintain all the schema changes for this table.
+
+        Args:
+            table (Table): The table object of this class.
+        """
+        col = Column('resource_name', String(256), default='')
+        LOGGER.info('Attempting to create column: %s', col.name)
+        col.create(table, populate_default=True)
```
blueandgold: if the pattern is to add the column as you mentioned, then should we surround try/except around each `col.create()`?  Otherwise, when adding `xyz` in the future, would the execution reach `xyz` if `resource_name` already exists?
```
         col = Column('resource_name', String(256), default='')
         col.create(table, populate_default=True)

         col = Column('xyz', String(256), default='')
         col.create(table, populate_default=True)
```

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -183,6 +184,18 @@ def __repr__(self):
...
             self.violation_type, self.resource_type, self.rule_name)
 
+    @staticmethod
+    def update_schema():
+        """Maintain all the schema changes for this table.
+
+        Returns:
+            dict: A mapping of Action: Column.
+        """
+        columnMapping = {'CREATE': Column('resource_name',
```
blueandgold: What if we want to add more columns in the future?  Shouldn't we use a list?:
```
columnMapping = {'CREATE', [column1, column2, ..]}
```

blueandgold: Why camel casing?

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,113 @@
...
+LOGGER = logger.get_logger(__name__)
+
+
+class ColumnAction:
+    """Column action class."""
+    DROP = 'DROP'
+    CREATE = 'CREATE'
+
+
+def create_col(table, col):
```
blueandgold: Can we avoid abbreviation here?
```
def create_column(table, column):
```

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,113 @@
...
+    """Create Column.
+
+    Args:
+        table (Table): The table object.
+        col (Column): The column object."""
```
blueandgold: Can you be more specific on the `Column` and `Table` object's type here?  Other than just 'Column` and Tables.

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,113 @@
...
+    LOGGER.info('Attempting to create column: %s', col.name)
+    col.create(table, populate_default=True)
+
+
+def drop_col(table, col):
```
blueandgold:
```
def drop_column(table, column):
```

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Update schema changes.
+    # Find all the child classes inherited from declarative base class.
+    base_subclasses = _find_subclasses(base)
+
+    # Find all the Table objects for each of the classes.
+    # The format of tables is: {table_name: Table object}.
+    tables = base.metadata.tables
+
+    schema_update_method_name = 'update_schema'
```
blueandgold: Be consistent with the `update_schema()` method, so `update_schema_method_name` would be better.
```
update_schema_method_name = 'update_schema'
```

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,113 @@
...
+
+    for subclass in base_subclasses:
+        schema_update = getattr(subclass, schema_update_method_name, None)
```
blueandgold: Better as:
```
update_schema = getattr(subclass, update_schema_method_name, None)
```

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,113 @@
...
+        if callable(schema_update) and subclass.__tablename__ in tables:
+            LOGGER.info('Updating table %s', subclass.__tablename__)
+            # schema_update will require the Table object.
+            try:
+                table = tables.get(subclass.__tablename__)
+                columnMapping = schema_update()
```
blueandgold: Again, why camel casing?

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -183,6 +184,18 @@ def __repr__(self):
         return string.format(
             self.violation_type, self.resource_type, self.rule_name)
 
+    @staticmethod
+    def update_schema():
```
blueandgold: This method does not do any update.  It would be better to be called as `get_update_actions()` or `get_migrate_actions()`.

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,113 @@
...
+                for columnAction, column in columnMapping.iteritems():
+                    columnAction = columnAction.upper()
+                    if columnAction in ColumnActionMap:
+                        ColumnActionMap.get(columnAction)(table, column)
```
blueandgold: Should there be a `try/except` around this line, so that if a column exists, then it can move to the next column?

google/cloud/forseti/services/scanner/dao.py
joecheuk: Updated.

google/cloud/forseti/services/scanner/dao.py
joecheuk: Updated.

joecheuk: Updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: Updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: Updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: Updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: Updated.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -183,6 +184,18 @@ def __repr__(self):
...
             self.violation_type, self.resource_type, self.rule_name)
 
+    @staticmethod
+    def get_schema_update_actions():
+        """Maintain all the schema changes for this table.
+
+        Returns:
+            dict: A mapping of Action: Column.
+        """
+        column_mapping = {'CREATE': [Column('resource_name',
```
blueandgold: Can we move the list out, so that it's easier to read?

```
columns_to_create = [
    Column('resource_name',  String(256), default=''),
    Column('abc',  String(256), default=''),
    Column('xyz',  String(256), default=''),
]

column_mapping = {
    'CREATE': columns_to_create,
    DELETE: columns_to_delete,
}
```

blueandgold: Instead of `column_mapping`, it would document better if the naming is `schema_update_mapping` or `schema_action_mapping`

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,118 @@
...
+    base_subclasses = _find_subclasses(base)
+
+    # Find all the Table objects for each of the classes.
+    # The format of tables is: {table_name: Table object}.
+    tables = base.metadata.tables
+
+    schema_update_actions_method = 'get_schema_update_actions'
+
+    for subclass in base_subclasses:
+        update_actions = getattr(subclass, schema_update_actions_method, None)
```
blueandgold: This line here needs to be consistent with the naming comment above.

google/cloud/forseti/services/scanner/dao.py
joecheuk: Updated

joecheuk: Updated

install/gcp/upgrade_tools/db_migrator.py
joecheuk: Updated

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -0,0 +1,151 @@
...
+        if (not callable(get_schema_update_actions) or
+                subclass.__tablename__ not in tables):
+            LOGGER.warn('Method: %s is not callable or Table: %s doesn\t '
+                        'exist', schema_update_actions_method,
+                        subclass.__tablename__)
+            continue
+        LOGGER.info('Updating table %s', subclass.__tablename__)
+        # schema_update will require the Table object.
+        table = tables.get(subclass.__tablename__)
+        schema_update_mapping = get_schema_update_actions()
```
blueandgold: Just one more nit: this line is a bit inconsistent.  We should make it read more smoothly as:
```
schema_update_actions = get_schema_update_actions()
```

joecheuk: Updated

PR #1873, opened 2018-08-10
Re-enable groups scanner test

### LINKED ISSUE #1672: Scanner G Suite groups unit tests are skipped
- [x] Which version of Forseti Security you're using (look in `google/cloud/security/__init__.py` in your installation).
2.0-RC3
 - [x] Which module(s) (inventory, scanner, enforcer, explain) you're having trouble with.
Scanner. groups scanner.
 - [ ] Include errors, log output, and host operating system, including installed packages.
 - [x] Note how you deployed (via Deployment Manager, local system, etc.).
Depolyment Manager

G Suite groups scanner doesn't have unit tests running. All the tests are being been skipped: https://github.com/GoogleCloudPlatform/forseti-security/blob/2c0c5754ca5d38d75774e6ecac8c6ea37ef7ac6c/tests/scanner/groups_scanner_test.py#L70

blueandgold:
This also consolidates the 3 separate test cases into one, since the test flow is sequential, reduces repetitive code, and allows the pickled files to be removed.

Fixes #1672


tests/scanner/test_data/fake_groups_scanner_data.py
```diff
@@ -36,144 +37,87 @@
...
-     'id': 3L,
-    },
-    {'direct_member_count': 3L,
-     'group_email': 'ddddd@mycompany.com',
-     'group_id': 'ddddd',
-     'group_kind': 'admin#directory#group',
-     'id': 4L,
-    },
+    FakeGroup('aaaaa', 'aaaaa@mycompany.com', 'admin#directory#group', 3L),
+    FakeGroup('bbbbb', 'bbbbb@mycompany.com', 'admin#directory#group', 3L),
```
joecheuk: The direct_member_count doesn't seem right compare with the data below, BBB and DDD have more direct_member_count than 3.

blueandgold: This is not used, so removing.  The exact members will be checked with a direct comparison.

PR #1875, opened 2018-08-10
Add the -r flag for requirements.txt

carise:


blueandgold:
The link is that the html-proofer is complaining about is working. Opening #1881 to resolve this, so this can be merged.

Issue #1869, opened 2018-08-09
Violations not firing


bkudzia:
I am having some issues with violations firing.

 - using 2.0
 - Scanner, notifier not triggering violations
 - no errors, but trying to configure bucket ACL violations with allUsers is not working (an example)
 - deployed via the 2.0 installer

Even though my bucket rule is defined this way (the default):
rules:
  - name: sample bucket acls rule to search for public buckets
    bucket: '*'
    entity: allUsers
    email: '*'
    domain: '*'
    role: '*'
    resource:
        - resource_ids:
          - <removed>
  - name: sample bucket acls rule to search for exposed buckets
    bucket: '*'
    entity: allAuthenticatedUsers
    email: '*'
    domain: '*'
    role: '*'
    resource:
        - resource_ids:
          - <removed>

And using Explainer shows the "allUsers" group in two roles on my test bucket:
/home/ubuntu/forseti-security/rules$ forseti explainer get_policy bucket/anothertest1234
{
  "bindings": [
    {
      "role": "roles/storage.legacyBucketOwner", 
      "members": [
        "projecteditor/new-test-212622", 
        "projectowner/new-test-212622"
      ]
    }, 
    {
      "role": "roles/storage.objectViewer", 
      "members": [
        "allusers"
      ]
    }, 
    {
      "role": "roles/storage.legacyObjectReader", 
      "members": [
        "allusers"
      ]
    }, 
    {
      "role": "roles/storage.legacyBucketReader", 
      "members": [
        "projectviewer/new-test-212622"
      ]
    }
  ], 
  "etag": "8203811fee8d842975797df5594e09ed"
}

It continues to tell me there are no violations.  What am I missing?  I've tried all flavors of allusers, wildcards, and quotes within the bucket yaml file, but nothing triggers.  If i change the entity to '*', Forseti reports violations, but in no row does it list the all users group.  I've been at this for several hours now, and I can't figure this out.  My guess is it's something really stupid, but i can't find it.  Let me know your thoughts.

blueandgold:
Hi @bkudzia, the `bucket_rules_scanner` works on the legacy ACL.  So, it is not meant for the IAM policies that you listed.  Instead, you will have to use the iam_scanner

https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/rules/iam_rules.yaml

And specify `bucket` as the resource type.

bkudzia:
[...]

PR #1878, opened 2018-08-10
Fix iam scanner so that it audits allUsers correctly

blueandgold:
Fix a bug that user has found in #1869 

This is how I think that `allUsers` should be audited.  Previously, it was looking at any member, which does not seem right.


blueandgold:
@ahoying there is already unit tests in place that covers the `allUsers` is matched correctly.  I've updated the test, so please take a look.

blueandgold:
Thanks, opened an issue so that improving the test is remembered.  :)

PR #1884, opened 2018-08-13
[Fixes #1865] Fix bigquery scanner to respect resources

### LINKED ISSUE #1865: BigQuery rules engine does not respect resources
Currently, the engine only checks for the existence of resources (https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/scanner/audit/bigquery_rules_engine.py#L131) but does not store or use them any further. 

Multiple resources defined in the rule simply duplicate the rule.

When applying a bigquery acl to a rule, the resource is unchecked, thus the rule applies to any matching acl.

umairidris:
Prior to this change, the bigquery scanner would match rules against all resources, even if it did not apply.

For example, a rule set on organization 'foo' would be incorrectly matched against a bigquery dataset under organization 'bar'.


google/cloud/forseti/scanner/scanners/bigquery_scanner.py
```diff
@@ -14,16 +14,22 @@
...
 from google.cloud.forseti.common.util import logger
 from google.cloud.forseti.scanner.audit import bigquery_rules_engine
 from google.cloud.forseti.scanner.scanners import base_scanner
 
 
 LOGGER = logger.get_logger(__name__)
 
+BigqueryAccessControlsData = collections.namedtuple(
+    'BigqueryAccessControlsData',
+    ['resource', 'bigquery_acl'])
```
dufton: (here and in function parameters of find_policy_violations). To avoid ambiguity, I think we should either rename resource to something like parent_project, or change resource to storing the dataset resource.

A possible advantage of using the dataset as the resource is it allows the bigquery scanner to examine other properties of the dataset (e.g. location) in the future.

tests/scanner/scanners/bigquery_scanner_test.py
```diff
@@ -0,0 +1,132 @@
...
+            full_name=resource['full_project_name'],
+            type='project',
+            name='projects/' + resource['project_id'],
+            parent_type_name='',
+            parent=None,
+            data='',
+        )
+
+        dataset= Resource(
+            full_name=resource['full_name'],
```
dufton: This full_name is the name of the dataset_policy. Trim the end off: '/'.join(full_name.split('/')[:-3]) + '/'

google/cloud/forseti/scanner/scanners/bigquery_scanner.py
```diff
@@ -115,32 +121,61 @@ def _retrieve(self):
...
                 # do not use project_id = policy.parent.parent.name
                 # which will cause db session conflict.
                 # Instead, parse the project_id from the full_name.
-                bq_acls.extend(
-                    BigqueryAccessControls.from_json(
-                        project_id=None,
-                        dataset_id=dataset_id,
-                        full_name=policy.full_name,
-                        acls=policy.data))
+                gen = BigqueryAccessControls.from_json(
```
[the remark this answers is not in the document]
umairidris: Simplified to converting to list + rename to  bq_acls.

I could replace project_id with parent_project, if that's what you mean? We need the parent Resouce class so just the id alone won't be enough.

The reason I made BigqueryAccessControlsData was to match other scanners. For example, all scanners (such as iam), return data which is usually a tuple of (parent_resource, policy) and then that is passed to find_violations. Instead of doing a tuple, I just made it a namedtuple. I feel that is easier to follow as I was getting a bit confused by the tuple in some parts. If you disagree I can make it a tuple like the rest.

Also, in bigquery_rules_engine.py, the BigqueryAccessControls object is also used as the rule itself. So while we could just omit the project for that, it may be cleaner to keep these two concepts separated.

What are your thoughts?

google/cloud/forseti/scanner/scanners/bigquery_scanner.py
umairidris: I went with option 1 since there is no gcp_type Dataset defined yet, and I don't want to make this PR even bigger :) We can change it fairly easily down the line if it's needed!

tests/scanner/scanners/bigquery_scanner_test.py
umairidris: Done.

google/cloud/forseti/scanner/scanners/bigquery_scanner.py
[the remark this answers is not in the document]
joecheuk: Sounds good, thanks for the clarifications!

google/cloud/forseti/scanner/scanners/bigquery_scanner.py
```diff
@@ -115,32 +122,61 @@ def _retrieve(self):
...
+                    dataset_id=dataset.name,
+                    full_name=policy.full_name,
+                    acls=policy.data))
+
+                for bq_acl in bq_acls:
+                    data = BigqueryAccessControlsData(
+                        parent_project=proj,
+                        bigquery_acl=bq_acl,
+                    )
+                    bq_acl_data.append(data)
```
nishchalc: nit: You can use list comprehension here like so: (will give a little speed boost as well)

return [BigqueryAccessControlsData(parent_project=proj, bigquery_acl=acl)
            for acl in bq_acls]

umairidris: We won't be able to return here as the bq_acl_data is two levels out :( and doing 

bq_acl_data.extend([ [BigqueryAccessControlsData(parent_project=proj, bigquery_acl=acl)
for acl in bq_acls])

seems a bit harder to read, so will leave it as is for now.

PR #1887, opened 2018-08-15
Templatize Forseti server region and Zone

johnrevans6:
Currently, The deployment template for the Forseti server explicitly references ref.cloudsql-instance.region as the server's subnet region, and likewise, references ref.cloudsql-instance.region-c as the server's GCE zone. Abstracted this away into FORSETI_SERVER_REGION, and FORSETI_SERVER_ZONE template variables to better match the Forseti client template.

### DIFF
--- a/deployment-templates/compute-engine/server/forseti-instance-server.py.schema
+++ b/deployment-templates/compute-engine/server/forseti-instance-server.py.schema
@@ -25,7 +25,7 @@ required:
 - zone
 - service-account
 - service-account-scopes
-- service-account-gsuite
+# - service-account-gsuite
 - scanner-bucket
 - database-name
 - src-path
--- a/install/gcp/installer/forseti_server_installer.py
+++ b/install/gcp/installer/forseti_server_installer.py
@@ -273,7 +273,7 @@ class ForsetiServerInstaller(ForsetiInstaller):
             'BUCKET_LOCATION': self.config.bucket_location,
             'GCP_SERVER_SERVICE_ACCOUNT': self.gcp_service_acct_email,
             'FORSETI_SERVER_REGION': self.config.cloudsql_region,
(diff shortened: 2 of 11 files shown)


install/gcp/installer/forseti_server_installer.py
```diff
@@ -272,6 +272,8 @@ def get_deployment_values(self):
             'FORSETI_BUCKET': bucket_name[len('gs://'):],
             'BUCKET_LOCATION': self.config.bucket_location,
             'GCP_SERVER_SERVICE_ACCOUNT': self.gcp_service_acct_email,
+            'FORSETI_SERVER_REGION': self.config.cloudsql_region,
```
blueandgold: This is an improvement, and will be accepted.  But the inherent oddity is still present... that is we are re-using a value for the cloudsql for the VM.  Either that flag should be renamed `--cloudsql-region` to just `--deployment-region`, or a new flag for the VM should be added `--vm_region`.  I realized this is a bigger change, so thoughts?

deployment-templates/compute-engine/server/forseti-instance-server.py.schema
```diff
@@ -25,7 +25,7 @@ required:
 - zone
 - service-account
 - service-account-scopes
-- service-account-gsuite
+# - service-account-gsuite
```
blueandgold: Why is this commented out?

rules/bigquery_rules.yaml
```diff
@@ -24,7 +24,7 @@ rules:
     resource:
       - type: organization
         resource_ids:
-          - {ORGANIZATION_ID}
+          - 826592752744
```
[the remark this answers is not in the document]
johnrevans6: Are you saying these files can be removed?

PR #1889, opened 2018-08-15
Fixing CrawlerTest to use Forseti Logging infrastructure

EricBeach:
All logging should be done through the central Forseti logging infrastructure so that we can selectively control which logs go to the console or not in an effort to fix test log pollution.

### DIFF
--- a/tests/services/inventory/crawling_test.py
+++ b/tests/services/inventory/crawling_test.py
@@ -13,10 +13,10 @@
 # limitations under the License.
 """Unit Tests: Inventory crawler for Forseti Server."""
 
-import logging
 import unittest
 from tests.services.inventory import gcp_api_mocks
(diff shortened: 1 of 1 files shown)


tests/services/inventory/crawling_test.py
```diff
@@ -28,6 +28,7 @@ class NullProgresser(Progresser):
 
     def __init__(self):
         super(NullProgresser, self).__init__()
+        self.LOGGER = logger.get_logger(__name__)
```
blueandgold: Can you please make the initialization of the `LOGGER` to be consistent with all the other files?  i.e. do it globally within the module, instead of class?

https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/services/explain/explainer.py#L19
```
```

Issue #1848, opened 2018-08-03
./install/scripts/docker_unittest_forseti.sh results in tests passing, but lots of errors being outputted


EricBeach:
To reproduce:

```bash
$ ./install/scripts/docker_unittest_forseti.sh
```

EXPECTED
Something similar to this:
```bash
$ ./install/scripts/docker_unittest_forseti.sh
Running unittests... 
----------------------------------------------------------------------
Ran 909 tests in 43.074s
OK (skipped=3)
```

ACTUAL
For example, from https://travis-ci.org/GoogleCloudPlatform/forseti-security/builds/411520053
```bash
$ ./install/scripts/docker_unittest_forseti.sh
Running unittests... 
...................................................................................................................................................................................................................................................................................................................................................................................................................ERROR:root:Progressor Warning: "key: owner, data: {'name': 'organizations/111222333'}"
ERROR:root:Progressor Warning: "key: owner, data: {'name': 'organizations/111222333'}"
.ERROR:root:Progressor Warning: u'1014'
.........sqlite:////tmp/tmpC_hnNJ
.................ERROR:google.cloud.forseti.services.dao:Grant not found: (user/u3,r/res1,admin)
ERROR:google.cloud.forseti.services.dao:Grant not found: (user/u3,r/res1,delete)
................ERROR:google.cloud.forseti.services.dao:Etags distinct, stored=075d4bdfea47783994ce1213828f581c, provided=somehash
/forseti-security/.eggs/SQLAlchemy-1.2.10-py2.7-linux-x86_64.egg/sqlalchemy/orm/strategies.py:781: SAWarning: Multiple rows returned with uselist=False for lazily-loaded attribute 'Binding.resource' 
/forseti-security/.eggs/SQLAlchemy-1.2.10-py2.7-linux-x86_64.egg/sqlalchemy/orm/strategies.py:781: SAWarning: Multiple rows returned with uselist=False for lazily-loaded attribute 'Binding.role' 
..ERROR:google.cloud.forseti.services.scanner.dao:Cannot create hash for a violation: 
...ERROR:google.cloud.forseti.services.scanner.dao:Cannot create hash for a violation with algorithm: sha512
......................ERROR: "/what/ever" is not readable.
.ERROR: "/this/does/not/exist" is not a file.
.ERROR: please specify the Forseti config file.
.ERROR: please specify the Forseti database string.
[...]

PR #1890, opened 2018-08-15
[Issue #1848] Fixing logging to use Forsetting logging infrastructure.

EricBeach:
This is setup work to enable us to turn off console logging for unit tests that are designed to emit errors. This will prevent us from polluting logs.

### DIFF
--- a/tests/services/util/db.py
+++ b/tests/services/util/db.py
@@ -17,8 +17,8 @@
 import os
 import tempfile
 
-from google.cloud.forseti.services.dao import create_engine
 from google.cloud.forseti.common.util import logger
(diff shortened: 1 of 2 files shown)


tests/services/util/db.py
```diff
@@ -14,11 +14,13 @@
 
 """Utils for Forseti Server services testing."""
 
-import logging
 import os
 import tempfile
 
 from google.cloud.forseti.services.dao import create_engine
+from google.cloud.forseti.common.util import logger
```
blueandgold: Sort this to be above line 20:
```
from google.cloud.forseti.services.dao
```

PR #1893, opened 2018-08-16
Inventory and model compute snapshots

jceresini:
We expect to need to perform checks on the backup/snapshot status of existing disks. In order to do so we'll need an inventory of the snapshots from compute engine.


blueandgold:
Awesome, thanks for this!  Please give me a chance to review this closely tomorrow!

google/cloud/forseti/services/inventory/base/gcp.py
```diff
@@ -667,6 +667,19 @@ def iter_networks(self, projectid):
         for network in self.compute.get_networks(projectid):
             yield network
 
+    @create_lazy('compute', _create_compute)
+    def iter_computesnapshots(self, projectid):
```
blueandgold: To be consistent, can we just call this `iter_snapshots`?  As other methods here, do not incorporate the api naming, i.e. `compute`.

tests/common/gcp_api/test_data/fake_compute_responses.py
```diff
@@ -1350,6 +1350,107 @@
 
 EXPECTED_NETWORK_NAME = [u"default1", u"default2"]
 
+SNAPSHOTS_LIST_PAGE1 = """
+{
+ "kind": "compute#snapshotList",
+ "id": "projects/cleardata-lab-jpc/global/snapshots",
```
blueandgold: Can we please anonymize this and everything else in the test data?

google/cloud/forseti/services/inventory/base/gcp.py
jceresini: Happy to change it, for what its worth i used compute disks as a reference for much of what I did here, and that calls the function `iter_computedisks`.

tests/common/gcp_api/test_data/fake_compute_responses.py
jceresini: Doh, cant believe I missed that. Thanks for catching it!

google/cloud/forseti/services/inventory/base/gcp.py
blueandgold: Good point, opening an issue to us to make the naming consistent here.   :)

blueandgold: https://github.com/GoogleCloudPlatform/forseti-security/issues/1901

PR #1895, opened 2018-08-16
Added try catch before uploading files to gcs bucket.

### LINKED ISSUE #1590: Add more error handling around saving CSV files into buckets.
Currently, there are no error handling code around saving CSV files, either at the API or at the caller.  This could potentially cause certain workflow to crash, for example the notification process as in #1577 .

joecheuk:
Resolves #1590

### DIFF
--- a/google/cloud/forseti/common/util/file_uploader.py
+++ b/google/cloud/forseti/common/util/file_uploader.py
@@ -17,8 +17,6 @@
 
 import tempfile
 
-from googleapiclient.errors import HttpError
-
 from google.cloud.forseti.common.data_access import csv_writer
 from google.cloud.forseti.common.gcp_api.storage import StorageClient
 from google.cloud.forseti.common.util import logger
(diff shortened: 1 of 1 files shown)


google/cloud/forseti/common/util/file_uploader.py
```diff
@@ -33,11 +35,16 @@ def upload_json(data, gcs_upload_path):
...
-        tmp_data.flush()
-        storage_client = StorageClient()
-        storage_client.put_text_file(tmp_data.name, gcs_upload_path)
+    try:
+        with tempfile.NamedTemporaryFile() as tmp_data:
+            tmp_data.write(parser.json_stringify(data))
+            tmp_data.flush()
+            storage_client = StorageClient()
+            storage_client.put_text_file(tmp_data.name, gcs_upload_path)
+    except HttpError as e:
```
blueandgold: To make sure this is really resilient, I think that we should consider a broad except here.

And if you use logger.exception, you can capture the stacktrace within the context of the exception handler.  Thus saving an extra logger.error line.

```
except Exception:
    logger.exception('Unable to upload json document to bucket %s:\n%s',
                                 gcs_upload_path, data)
```

PR #1896, opened 2018-08-17
Add a flake8 test (pycodestyle) to check for pep8 related stye

### LINKED ISSUE #1240: Check for PEP8 style (where it makes sense)
Support for this exists in the pycodestyle tool.

mirons-google:
This resolves #1240

### DIFF
--- a/google/cloud/forseti/common/gcp_type/firewall_rule.py
+++ b/google/cloud/forseti/common/gcp_type/firewall_rule.py
@@ -861,7 +861,7 @@ def ips_in_list(ips, ips_list):
     for ip_addr in ips:
         if not ips_list:
             return False
-        if not any([ip_in_range(ip_addr, ips) for ips in ips_list]):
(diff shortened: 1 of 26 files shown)


google/cloud/forseti/common/gcp_type/firewall_rule.py
```diff
@@ -861,7 +861,7 @@ def ips_in_list(ips, ips_list):
     for ip_addr in ips:
         if not ips_list:
             return False
-        if not any([ip_in_range(ip_addr, ips) for ips in ips_list]):
+        if not any([ip_in_range(ip_addr, addresses) for addresses in ips_list]):
```
blueandgold: Also change `ips_list`, since the other `ips` are changed?

mirons-google: We'd run into line length issues then. I improved the naming here.

FYI The reason I touched this is because the previously defined variable of `ips` was redefined in this list comprehension.

PR #1903, opened 2018-08-17
[Issue 1848] Mock out logger to fix almost all remaining instances of pollution of test logs. Remaining issuses involve server and will likely require some production refactoring.

EricBeach:


EricBeach:
@blueandgold - Thanks for the tip to mock! Much cleaner solution here.

EricBeach:
There are now only five errors left in the test logs and they are all in ``tests.services.server_test.ServerTest`` 
Source: https://travis-ci.org/GoogleCloudPlatform/forseti-security/builds/417454378

PR #1907, opened 2018-08-20
Improve CSCC usability

### LINKED ISSUE #1902: Improve CSCC Usability
Discussion with andychang:

- replace `UNKNOWN_RISK` with some summary information, perhaps the rules name
ideally - I'd like to have something that show's which scanner (e.g. firewall, blacklist, bq, etc.. and the top level issue - ie.  missing firewall,  blacklisted ip,  non-org user etc...
- populate the `finding_callback_url` with either the bucket/filename or the violations table & row id
- add a new property that gives the name of the db table

blueandgold:
Fixes #1902.

- added the new fields based on issue #1902
- did not add scanner type, and that will be a bigger change, and perhaps require a schema change
- made both the bucket output and api output to be consistent


google/cloud/forseti/notifier/notifiers/cscc_notifier.py
```diff
@@ -134,7 +139,9 @@ def _transform_for_api(self, violations):
                     'violation_data': violation.get('violation_data')
                 },
                 'source_id': 'FORSETI',
-                'category': 'UNKNOWN_RISK'
+                'category': violation.get('rule_name'),
+                'url': 'table:{}/id:{}'.format(
+                    VIOLATIONS_TABLE, violation.get('id'))
```
joecheuk: Is the 'url' property required? I think we can include the id inside 'properties:' above, having a url like this seems weird.

blueandgold: Moving this into the `properties` section so this is not clickable.

PR #1921, opened 2018-08-21
[Issue 1848] Fix more log pollution of tests.

EricBeach:


EricBeach:
@blueandgold - FYI - There is one more instance of log pollution, which is tricky as fixing it involves both mocking and parameterized

PR #1922, opened 2018-08-21
Add billing accounts to log sink scanner

dufton:
This adds support for billing account level log sinks in the log sink scanner.

Rules are specified using type: billing_account


tests/common/gcp_type/billing_account_test.py
```diff
@@ -0,0 +1,77 @@
...
+            parent=self.org_234)
+
+        self.assertEqual('000000-111111-222222', billing_acct.id)
+        self.assertEqual('billing_account', billing_acct.type)
+        self.assertEqual('billingAccounts/000000-111111-222222',
+                         billing_acct.name)
+        self.assertEqual('My Billing Account', billing_acct.display_name)
+        self.assertEqual(self.org_234, billing_acct.parent)
+
+    def test_create_from_resource_utils(self):
```
umairidris: This test doesn't seem to be testing anything from billing_account.py. Should it be in resources instead?

tests/scanner/audit/data/log_sink_test_valid_rules.yaml
```diff
@@ -67,3 +67,14 @@ rules:
       destination: '*'
       filter: ''
       include_children: '*'
+  - name: 'Only allow Billing Account sinks to audit logs project.'
+    mode: whitelist
+    resource:
+      - type: billing_account
+        applies_to: self
+        resource_ids:
+          - ABCD-1234
```
umairidris: nit: a more descriptive resource name may make it easier to understand the tests (e.g. FOO-BILLING, BILLING-ACCT-1, etc).

google/cloud/forseti/common/gcp_type/billing_account.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __init__(
+            self,
+            billing_account_id,
+            full_name=None,
+            data=None,
+            name=None,
+            display_name=None,
+            parent=None,
+            lifecycle_state=BillingAccountLifecycleState.UNSPECIFIED):
```
umairidris: Not a Python expert but it seems this is a mutable default arg and might be an anti-pattern?

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

tests/scanner/audit/data/log_sink_test_valid_rules.yaml
dufton: I was keeping these looking somewhat like a real billing account, which are hexadecimal strings with hyphens, but longer: 012345-6789AB-CDEF01

PR #1925, opened 2018-08-21
[Fixes #783] Support whitelist mode in Bigquery scanner

### LINKED ISSUE #783: Whitelisting mode for BigQuery scanner
Currently BigQuery scanner rules allow us only to blacklist entities that aren't allowed to access specific datasets. 

Having whitelisting mode (like in IAM policies scanner rules) would be very helpful.

umairidris:
Prior to this only blacklist was supported. This PR adds the whitelist mode.

Fixes #783


google/cloud/forseti/scanner/audit/bigquery_rules_engine.py
```diff
@@ -243,12 +266,33 @@ def __init__(self, rule_name, rule_index, rules):
...
+        rule_bigquery_acl = self.rules.bigquery_acl
+        rule_regex_to_val = {
+            rule_bigquery_acl.dataset_id: bigquery_acl.dataset_id,
+            rule_bigquery_acl.role: bigquery_acl.role,
+        }
+
+        return all([
+            re.match(rule_regex, acl_val)
+            for (rule_regex, acl_val) in rule_regex_to_val.iteritems()
+        ])
```
dufton: It was already an issue, but the list comprehension will undo the laziness of all, so this will always evaluate the second re.match. 

For here and all_matched below, it might be better to use a list of tuples instead of a dict, and a helper function
```
  def all_re_match(patterns_and_strings):
    for pattern, string in patterns_and_strings):
      if not re.match(pattern, string):
        return False
    return True
```
...
```
  return all_re_match([
      (rule_bigquery_acl.dataset_id, bigquery_acl.dataset_id),
      (rule_bigquery_acl.role, bigquery_acl.role)
  ])
```

It may even make sense to put the helper function in common/util/regular_exp.py

rules/bigquery_rules.yaml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# The BigQuery scanner rules are blacklists.
+# BigQuery scanner rules default to mode 'blacklist'.
 rules:
   - name: BigQuery rule to search for public datasets
     dataset_id: '*'
```
dufton: Add "mode: blacklist" to these default rules in case it becomes a required field later.

umairidris: Done.

google/cloud/forseti/scanner/audit/bigquery_rules_engine.py
umairidris: Done.

I was not sure if the list of tuples was required, so I just kept it as a dict as it looks cleaner.

google/cloud/forseti/scanner/audit/bigquery_rules_engine.py
```diff
@@ -147,17 +157,29 @@ def _build_rule(cls, rule_def, rule_index, raw_resource):
...
+            bigquery_acl=bq_acls.BigqueryAccessControls(
+                project_id='',
+                dataset_id=regular_exp.escape_and_globify(dataset_id),
+                full_name='',
+                special_group=regular_exp.escape_and_globify(special_group),
+                user_email=regular_exp.escape_and_globify(user_email),
+                domain=regular_exp.escape_and_globify(domain),
+                group_email=regular_exp.escape_and_globify(group_email),
+                role=regular_exp.escape_and_globify(role.upper()),
+                view={},
```
joecheuk: Is there any reason why we updated the view from string to dictionary?

umairidris: I may have (incorrectly) seen it as a dict instead of a string somewhere and changed it. If it's supposed to be a string I can revert it.

google/cloud/forseti/common/util/regular_exp.py
```diff
@@ -39,3 +39,19 @@ def escape_and_globify(pattern_string):
     if pattern_string == '*':
         return '^.*$'
     return '^{}$'.format(re.escape(pattern_string).replace('\\*', '.+?'))
+
+
+def all_match(pattern_to_val):
+    """Lazily determines whether the given regexes match the values.
```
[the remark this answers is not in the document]
umairidris: In the previous list comprehension method, all regex matches would be evaluated into a list and then all() would be called on that list. Now, as soon as the first regex fails to match it will return (so the rest don't get evaluated). I am not sure if lazy is the right word but the idea was it lazily evaluates each regex.. i.e. doesn't do it if not needed. Let me know if you prefer another description. If you don't care too much about this micro optimization I can put the list comprehension back. Doubt it will be any difference in speed in reality.

google/cloud/forseti/scanner/audit/bigquery_rules_engine.py
umairidris: Found it: https://github.com/GoogleCloudPlatform/forseti-security/blob/c8d1485e5c88673d8c6190ba0e9a7c22013db385/google/cloud/forseti/common/gcp_type/bigquery_access_controls.py

The ACL class lists it as a dict which my linter caught. Either we should change the callers to use dict, or change the class to say the view is a string. I don't even know if the view is used, so it may be possible to delete it entirely...

google/cloud/forseti/common/util/regular_exp.py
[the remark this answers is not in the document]
joecheuk: I was thinking lazy evaluation is more on a line of not creating all the values at the same time (maybe doing so through a generator). I am not concerned about the performance in this case, just a bit confused about the wording in the docstring.

umairidris:
Moving @nischalc to cc as he has been OOO recently. Feel free to merge when you're comfortable!

Issue #1845, opened 2018-08-03
Add benchmark to the Forseti processes with OpenCencus


blueandgold:
This came out of #1844, where we need to know more info on where exactly the slowness is coming from: whether it's the actual inventory, or building the data model, etc.

It would also be nice to benchmark the the individual steps, like the inventory of different resources (gsuite users, groups, etc), and the running of the scanners & notifiers.

mirons-google:
A follow-up here. This can actually be done a couple ways, AFAIK.

1. A deep integration with Stackdriver
2. Use opencensus.io (which allows other exporters besides Stackdriver).

mirons-google:
Questions we need answers for.

1. Where in each of the core modules (inventory, scanning, model building, etc.) are we spending the most time and should be optimized?
1. Where can I tell a user to go in Stackdriver to gather the appropriate profiling information to debug an issue they opened?
1. What are the performance expectations for various sizes of organizations (buckets, projects, etc.) and G Suite domains (groups, users) a person can expect and what implications does the machine type of the Forseti Server play in that?

red2k18:
Planning to use OpenCensus to collect metrics and traces and Stackdriver exporter to export the metrics to Google Cloud.

red2k18:
Forseti is a security suite composed of gRPC services written in Python. As part of discussions about the instrumentation for Forseti, it was decided to use OpenCensus to collect Forseti metrics, logs, traces and events to add monitoring, tracing and profiling to Forseti services. 

The integration with OpenCensus is out-of-the box and only requires to set up a trace interceptor that will intercept each gRPC call and add code to calculate latency of each request (gRPC calls), and build traces automatically. 

The OpenCensus Stackdriver Stats Exporter exports data to Stackdriver Monitoring which provides visibility into performance, uptime and overall health of cloud-powered applications. Stackdriver ingests that data and generates insights via dashboard, charts, alerts etc.

red2k18:
OpenCensus provides support for following exporters. We plan to give users the ability to choose the exporter. 
-[x] Stackdriver (Python supported for traces)
-[x] Zipkin (Python supported for traces)
-[x] Jaeger (Python supported for traces)
-[x] Instana
-[x] Datadog
-[x] Prometheus
-[x] SignalFX

red2k18:
Implementation plan:

- [ ] Core tracing for gRPC server/client
- [ ] Tracing integration libraries (requests, sqlalchemy, httplib, httplib2)
[...]

Issue #1844, opened 2018-08-03
Forseti cron job takes 8 hours to complete, after enabling GSuite Inventory of 100K users.


blueandgold:
As reported by user.  And half the time, there is no csv file in the buckets.

Issue #1438, opened 2018-04-19
Enable profiling to detect and track performance bottlenecks


mirons-google:
Here's an [example](https://docs.python.org/2/library/profile.html). Ideally, I can pass in an argument to any of the Forseti Entry [points](https://github.com/GoogleCloudPlatform/forseti-security/blob/2.0-dev/setup.py#L139) that enables profiling and it will be collected/logged for performance troubleshooting.

PR #1926, opened 2018-08-21
Add OpenCensus instrumentation for tracing

ocervell:
This PR brings support for OpenCensus tracing for all gRPC Forseti services:

* Add OpenCensus interceptors to Forseti's `gRPC` server and client
* Set up an exporter of traces to Stackdriver Trace (plan on supporting more via cfg changes)
* Add appropriate IAM role `cloudtrace.agent` to allow client/server VMs to write traces to Stackdriver (during deployment)
* Add 'opencensus' and 'google-cloud-trace' dependencies


ocervell:
Seems like the library fails to contact Stackdriver API on the Travis CI VM, do we have gcloud credentials set up on that VM ? 

Maybe we can just disable the Stackdriver exporter in testing mode and have the FileExporter that works locally instead.

blueandgold:
Relates to #1845, #1438

ocervell:
I signed it !

ahoying:
Please make this change aware of if the opencensus package is installed, if it is not installed it should be a no-op.

For changes like this, where a new third party library is introduced for an optional feature, or a feature that can be disabled without impacting the rest of forseti, the code needs to fail back cleanly when that library is not installed.

This happens to be the first new feature to be implemented where this would apply.

ocervell:
We're planning on using OpenCensus for our monitoring as well. 

While tracing could be made optional and configurable, monitoring (exposing OpenCensus metrics to monitoring systems like Stackdriver) seems required for Forseti, so the dependency would be as well.

Is there currently any way to add an optional dependency in Forseti deployments ?

mirons-google:
Since this is a first step in working to solve #1845 #1844 and #1438 I think it makes sense to consider the package as required.

The preferred way to do that is in setup.py currently.

ahoying:
Forseti is both a deployment but also a set of libraries that are used elsewhere. Introducing additional third party dependencies makes the use of forseti libraries more difficult when a full deployment is not desired.

I recommend wrapping the opencensus imports in the tracing module in a try catch, and setting a global to True or False depending on if they are available or not.

Then the code can handle the cases where opencensus is installed or not installed by checking the value of that global, and you can fall back gracefully.

ocervell:
@ahoying made some good points there. Users should be able to install the `forseti-security` library without extra dependencies. 

My last commit brings support for:

- More interceptors to be added (e.g a MetricsInterceptor)
- A server CLI flag `--enable-tracing` that's disabled by default.
- Added `--enable-tracing` to our `systemctl` startup script.
- Tracing libraries removed as required dependencies and added as `OPTIONAL_DEPENDENCIES` to `setup.py`. They can be installed using `pip install .[tracing]`, statement I've added to our Forseti deployments that will enable tracing by default.

Please review and let me know what you think.

google/cloud/forseti/services/utils.py
```diff
@@ -258,3 +259,20 @@ def get_resource_id_from_type_name(type_name):
...
+
+    Returns:
+        bool: True if check pass, False otherwise.
+    """
+    # Check if OpenCensus exist on system
+    try:
+        imp.find_module('opencensus')
+        return True
+    except ImportError:
+        logging.warning('Cannot enable tracing because the `opencensus` library\
```
blueandgold: Backslash should not be used for line cotinuations.
```
logging.warning(
    'Cannot enable tracing because the `opencensus` library'
    'was not found. Run `pip install .[tracing]` to install tracing libraries.')

File: deployment-templates/compute-engine/client/forseti-instance-client.py [hunk 2]
```diff
@@ -133,6 +133,9 @@ def GenerateConfig(context):
 pip install -q --upgrade setuptools wheel
 pip install -q --upgrade -r requirements.txt
 
+# Install instrumentation libs
+pip install .[tracing]
```
blueandgold: This is new for me.  But is this relative import the preferred way here?  Would it be better to do specify it as:
```
pip install <module>[tracing]
```
https://stackoverflow.com/a/24640526/2830207

google/cloud/forseti/services/utils.py
```diff
@@ -258,3 +259,20 @@ def get_resource_id_from_type_name(type_name):
     if '/' in type_name:
         return type_name.split('/')[-1]
     return type_name
+
+
+def opencensus_enabled():
```
blueandgold: This method's naming would be more accurate as `is_opencensus_installed()`, because that's what it's checking.

google/cloud/forseti/services/client.py
```diff
@@ -797,3 +799,17 @@ def delete_model(self, model_name):
...
+
+    @property
+    def interceptors(self):
+        """gRPC client interceptors.
+
+        Returns:
+            tuple: A tuple of interceptors.
+        """
+        interceptors = []
+        if opencensus_enabled():
```
blueandgold: What happens if `enable_tracing` is disabled on the server?  Is there any impact here in the client?

google/cloud/forseti/services/client.py
```diff
@@ -797,3 +799,17 @@ def delete_model(self, model_name):
         """
 
         return self.model.delete_model(model_name)
+
+    @property
+    def interceptors(self):
```
blueandgold: Why is this a property method here in the `services/client.py`, but `setup_interceptors()` in `services/server.py`?

google/cloud/forseti/services/server.py
```diff
@@ -170,6 +196,10 @@ def main():
         '--enable_console_log',
         action='store_true',
         help='Print log to console.')
+    parser.add_argument(
+        '--enable-tracing',
+        action='store_true',
```
blueandgold: Is this, and should be, enabled?  Per your PR description, this should be disabled by default?

google/cloud/forseti/services/tracing.py
```diff
@@ -0,0 +1,84 @@
...
+        OpenCensusClientInterceptor: a gRPC client-side interceptor.
+    """
+    exporter = setup_exporter()
+    tracer = Tracer(exporter=exporter)
+    return client_interceptor.OpenCensusClientInterceptor(
+        tracer,
+        host_port=endpoint)
+
+
+def trace_server_interceptor():
```
blueandgold: Method naming should generally start with verbs.  So, `create_trace_server_interceptor()`.

google/cloud/forseti/services/tracing.py
```diff
@@ -0,0 +1,84 @@
...
+from opencensus.trace.exporters import stackdriver_exporter, file_exporter
+from opencensus.trace.exporters.transports import background_thread
+from opencensus.trace.ext.grpc import client_interceptor, server_interceptor
+
+from google.cloud.forseti.common.util import logger
+
+LOGGER = logger.get_logger(__name__)
+
+
+def trace_client_interceptor(endpoint):
```
blueandgold: Method naming should generally start with verbs.  So, `create_trace_client_interceptor()`.

google/cloud/forseti/services/server.py
ocervell: Yes, `store_true` actually stores False by default in Python (cf: https://docs.python.org/2/library/argparse.html#action)

google/cloud/forseti/services/tracing.py
ocervell: I'd simplify it to `create_client_interceptor` and `create_server_interceptor` then.

I think it looks good that way:
```
from google.cloud.forseti.services import tracing
tracing.create_client_interceptor(endpoint='localhost:5001')
tracing.create_server_interceptor()
```

When we do the monitoring piece, we'll have similarly:
```
from google.cloud.foresti.services import monitoring
monitoring.create_client_interceptor(endpoint='localhost:5001')
monitoring.create_server_interceptor()
```

Side question: should we move `tracing.py` (and future `monitoring.py`) into `google/cloud/forseti/common/` or does it make sense to keep it next to the `client.py` and `server.py` ?

google/cloud/forseti/services/tracing.py
```diff
@@ -0,0 +1,84 @@
...
+    """Intercept gRPC calls on client-side and add tracing information
```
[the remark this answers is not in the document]
ocervell: No problem.

google/cloud/forseti/services/tracing.py
ocervell: Ok. See previous comment to drop the `_trace`

google/cloud/forseti/services/tracing.py
```diff
@@ -0,0 +1,84 @@
...
+            transport=background_thread.BackgroundThreadTransport)
+        LOGGER.info(
+            'StackdriverExporter set up successfully for project %s.',
+            exporter.project_id)
+    except Exception as e:
+        LOGGER.exception(e)
+        LOGGER.warning('StackdriverExporter set up failed. Using FileExporter.')
+        exporter = file_exporter.FileExporter(
+            transport=background_thread.BackgroundThreadTransport)
+        LOGGER.info(
```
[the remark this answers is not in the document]
ocervell: Right.

google/cloud/forseti/services/server.py
blueandgold: TIL  :)

google/cloud/forseti/services/client.py
```diff
@@ -718,7 +718,9 @@ def __init__(self, endpoint=DEFAULT_ENDPOINT, ping=False):
         Raises:
             Exception: gRPC connected but services not registered
         """
-        self.channel = grpc.insecure_channel(endpoint)
+        self.endpoint = endpoint
+        self.channel = grpc.insecure_channel(self.endpoint)
```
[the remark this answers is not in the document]
blueandgold: Yes, your proposal is much more clear for me:
```
self.channel = grpc.intercept_channel(
    grpc.insecure_channel(endpoint),
    *self.interceptors)
```

google/cloud/forseti/services/client.py
blueandgold: Got it, thanks.

google/cloud/forseti/services/client.py
blueandgold: Can we just leave a comment here?
```
# It's okay for this to be enabled on the client, even if the tracing flag is disabled on the server.
```

google/cloud/forseti/services/tracing.py
blueandgold: Your naming proposal sounds great!

I am neutral on moving this to the common package, as that to me is for sharing between different packages.  Do you think that will be the case in the future?

google/cloud/forseti/services/tracing.py
```diff
@@ -0,0 +1,84 @@
...
+        `StackdriverExporter`: A Stackdriver exporter.
+        `FileExporter`: A file exporter.
+    """
+    try:
+        exporter = stackdriver_exporter.StackdriverExporter(
+            transport=background_thread.BackgroundThreadTransport)
+        LOGGER.info(
+            'StackdriverExporter set up successfully for project %s.',
+            exporter.project_id)
+    except Exception as e:
```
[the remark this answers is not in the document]
blueandgold: We actually don't mind broad exceptions as a mean to keep the server from fatal crashes, and the exception event is handled appropriately and logged with stacktraces.

ocervell:
Yes, we need tests at least for the functions included in `tracing.py`.
Changed title to WIP since there is [a bug](https://github.com/census-instrumentation/opencensus-python/issues/270) blocking this to be merged.

- [ ] Add unit tests
- [ ] Include integration libraries tracing (requests, httplib, sqlalchemy)
- [ ] [Bugfix] Stackdriver exporter crash handler for OpenCensus

google/cloud/forseti/services/tracing.py
ocervell: Yes, I think one of the goals after this PR is to implement some kind of decorator to enable more tracing (sub-spans) on certain functions / classes. In that case those would be shared by other services.

google/cloud/forseti/common/opencensus/tracing.py
```diff
@@ -0,0 +1,108 @@
...
+    exporter = create_exporter()
+    sampler = always_on.AlwaysOnSampler()
+    if extras:
+        trace_integrations()
+    return server_interceptor.OpenCensusServerInterceptor(
+        sampler,
+        exporter)
+
+
+def trace_integrations(integrations=DEFAULT_INTEGRATIONS):
```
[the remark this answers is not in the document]
ocervell: isn't 'trace' a verb ? Like 'trace those integrations' ?

google/cloud/forseti/services/client.py
```diff
@@ -69,6 +70,24 @@ def wrapper(*args, **kwargs):
...
+        endpoint (str): The gRPC server endpoint. E.g: 'localhost:443'
+
+    Returns:
+        tuple: A tuple of interceptors.
+    """
+    interceptors = []
+    if is_opencensus_enabled():
+        # It's okay for this to be enabled on the client, even if the tracing
+        # flag is disabled on the server.
+        from google.cloud.forseti.common.opencensus import tracing
```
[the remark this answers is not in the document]
ocervell: We have to make an exception here: if we put the import at the top of the file, the import would fail if tracing is disabled because `opencensus` is not even installed. 

That's why the code checks if `is_opencensus_enabled()` before running this import, otherwise we would fail with an `ImportError`. 

I find this more elegant than `try`/`except` blocks for imports in `server.py` and `client.py`, what do you think ?

tests/common/opencensus/tracing_test.py
```diff
@@ -0,0 +1,89 @@
...
+    """Test Tracing resource."""
+
+    def setUp(self):
+        if not tracing_libs:
+            self.skipTest('Package `opencensus` not installed.')
+
+    @mock.patch('opencensus.trace.ext.grpc.client_interceptor.OpenCensusClientInterceptor')
+    def test_create_client_interceptor(self, mock):
+        create_client_interceptor('localhost')
+        self.assertTrue(mock.called)
```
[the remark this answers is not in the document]
ocervell: I need to check if the OpenCensusClientInterceptor can be instantiated without a gRPC server running.

google/cloud/forseti/common/opencensus/tracing.py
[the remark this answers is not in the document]
ocervell: Would it be better called `trace_integration_libraries` ?

google/cloud/forseti/common/opencensus/tracing.py
```diff
@@ -0,0 +1,108 @@
...
+        extras (bool): If set to True, also trace integration libraries.
+
+    Returns:
+        OpenCensusServerInterceptor: a gRPC server-side interceptor.
+    """
+
+    exporter = create_exporter()
+    sampler = always_on.AlwaysOnSampler()
+    if extras:
+        trace_integrations()
```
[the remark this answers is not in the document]
blueandgold: Would it be very hard to assert on the state of `config_integration` after `config_integration.trace_integrations()` is called?  I would think that it should be possible.

If that is too hard, can you please put a comment on the return statement in that the returned values are only used for testing?  So that in the future we can remember the return is there for testing purposes.

google/cloud/forseti/services/client.py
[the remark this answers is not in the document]
ocervell: Sure, but keep in mind that this means wherever we'll need to add spans to traces (in each different service, such as inventory), we'll have to put the same try / except at the top of each file that we add tracing too.
Is this something acceptable ?

ocervell:
Thanks for your in-depth review. I think I've addressed the above comments with my last commits.

PR #1929, opened 2018-08-22
[Issue 1848] Fixing parameterized test to mock out logger to fix log pollution

EricBeach:


tests/enforcer/gce_firewall_enforcer_test.py
```diff
@@ -1300,15 +1300,18 @@ def test_check_change_operation_order(self, name, quota, usage,
...
-        if expect_exception:
-          with self.assertRaises(fe.FirewallQuotaExceededError):
-            self.enforcer._check_change_operation_order(
-                insert_rule_count, delete_rule_count)
-        else:
-          delete_before_insert = self.enforcer._check_change_operation_order(
-              insert_rule_count, delete_rule_count)
 
-          self.assertEqual(expect_delete_before_insert, delete_before_insert)
+        with mock.patch('google.cloud.forseti.enforcer.gce_firewall_enforcer.LOGGER') as mock_logger:
```
blueandgold: Thanks for fixing this!  Out of curiosity, why do we need to patch in the mock loggers this way (with `with`), vs at the top of the method?

EricBeach: Great question. I am no Python expert, so there may be a way. However, this method, unlike the others, uses ``@parameterized.parameterized.expand(`` and I could not find a way to get the mock into those parameters. If you have a way, let me know and I'd be happy to do it.

blueandgold: Hmmm, let me play around with this, and see what is happening.

EricBeach: @blueandgold - any thoughts here?

EricBeach:
@blueandgold - friendly ping

tests/enforcer/gce_firewall_enforcer_test.py
ahoying: See https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/tests/common/gcp_api/compute_test.py#L226

EricBeach:
Fixed up and much cleaner now. I must have made a small syntax error when trying the simple mock. Thanks @ahoying !

Issue #1744, opened 2018-06-28
Document how to install Forseti with Terraform


blueandgold:
Recent work has enabled Forseti to be installed with Terraform but we still need to document how to do this.

gkuchta:
I don't suppose that there's a preview version of this documentation? I'm in the middle of deploying this via terraform for a PoC.

joecheuk:
Hi @gkuchta , @johnrevans6 is working on the Forseti/Terraform documentation, feel free to reach out to John if you have any questions!

gkuchta:
@johnrevans6 awesome! thank you!

joecheuk:
Hi @johnrevans6 , just a follow up here - Since the integration is already finished, would you mind updating the Forseti/Terraform documentation on the Forseti official website?

johnrevans6:
Hmmm not sure I have access to that... I linked to the repo above which has instructions. 

Here's the link again: https://github.com/terraform-google-modules/terraform-google-forseti

joecheuk:
The website is a branch (forsetisecurity.org-dev) in this repository, you can edit the corresponding files under **_docs/_latest**
https://github.com/GoogleCloudPlatform/forseti-security/tree/forsetisecurity.org-dev/_docs/_latest

Please let me know if you need more information!

johnrevans6:
Ah, good to know. Sure, I can do that for you. Do you just want a link back to the Forseti module in Terraform, or something more robust?

johnrevans6:
Works for me :)

joecheuk:
Thanks for handling this!

johnrevans6:
No problem. So for the FAQ, it looks like you're basically templating out read me files. Is this [1] where those files live?

[1] https://github.com/GoogleCloudPlatform/forseti-security/tree/forsetisecurity.org-dev/_faq

joecheuk:
Yes, I think it should fall under section "installation_and_deployment"

johnrevans6:
Agreed. Okay, cool. I'll get a PR out by eod hopefully.

joecheuk:
Resolved #1939

PR #1939, opened 2018-08-22
Add entry in FAQ for Terraform deployment

johnrevans6:
Resolving [1]. We have now open-sourced a Terraform module geared towards deploying Forseti. 

[1] https://github.com/GoogleCloudPlatform/forseti-security/issues/1744

### DIFF
--- a/_faq/installation_and_deployment/deploy-using-terraform.md
+++ b/_faq/installation_and_deployment/deploy-using-terraform.md
@@ -5,4 +5,5 @@ order: 8
 
 {::options auto_ids="false" /}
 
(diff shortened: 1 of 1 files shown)


_faq/installation_and_deployment/deploy-using-terraform.md
```diff
@@ -0,0 +1,8 @@
+---
+title: How can I deploy Forseti using Terraform?
+order: 8
+---
+
+{::options auto_ids="false" /}
+
+* There is now a [Terraform module](https://github.com/terraform-google-modules/terraform-google-forseti) in available for deploying Forseti.
```
blueandgold: Instead of `in available`, it might be better as `that is available`.

Issue #1937, opened 2018-08-22
Difficult to express specific rules for BigQuery


umairidris:
As a follow up to PR #1925, we would like to add specific whitelists for groups and roles on our projects.

e.g. One whitelist rule for project foo with multiple roles definitions for OWNER, WRITER and READER. In each role, I want to define multiple groups/users.

Currently, a bigquery acl rule only allows a single rule and single user/group. Since the rules work as an "AND" (i.e. no violation if ALL rules don't report a violation), it is impossible for us to define multiple specific groups/users in rules today.

We could use regexes, but regex get escaped and only globs are respected. So another option is to allow for regexes (at the cost of globs).

Looking at something like iam rules, the structure used there makes a lot more sense. I would like something like the following for bigquery acl rules:

```
  - name: Whitelist for project project_1
    mode: whitelist
    dataset_ids:
      - project_1:us_data
      - project_1:euro_data
    resource:
      - type: project
        resource_ids:
          - project_1
    bindings:
      - role: OWNER
        members:
          - group_email: 'project_1_owners@google.com'
      - role: WRITER
        members:
          - group_email: 'project_1-readwrite@google.com'
      - role: READER
        members:
          - group_email: 'project_1-readonly@google.com'
          - group_email: 'project_1-external@google.com'

```

Is it possible to add support for this? How important is backwards compatibility? I am willing to write up the code.

/cc @dufton @joecheuk

PR #1958, opened 2018-08-27
Default non-existent fields in bigquery acl to glob

umairidris:
Currently, a user needs to set each field in a bigquery rule that they don't care about to glob for the rule to work correctly. This is error prone and very verbose. Simplify by making the default glob. This behaviour matches other rules such as iam.

Part of Issue #1937.


rules/bigquery_rules.yaml
```diff
@@ -15,36 +15,21 @@
 rules:
   - name: BigQuery rule to search for public datasets
     mode: blacklist
-    dataset_id: '*'
```
joecheuk: Can we add a comment in this rule file saying that for the following fields, if not defined, the values will be defaulted to '*' so if user wants to define a rule on their own, they know what fields can be changed.

umairidris: Nice suggestion, done. It can be useful to have a more explicit schema to highlight these fields, filed #1962 for it.

PR #1964, opened 2018-08-28
Added a new rule to scan for bucket with allUsers in IAM policy

red2k18:
Added a new rule to scan for bucket with allUsers in IAM policy


rules/iam_rules.yaml
```diff
@@ -25,8 +25,22 @@ rules:
     bindings:
       - role: roles/resourcemanager.organizationAdmin
         members:
-          - user:*@{DOMAIN}
-          - group:*@{DOMAIN}
+          - user:*@henrychang.mygbiz.com
```
blueandgold: Please add the placeholders back.

rules/iam_rules.yaml
```diff
@@ -25,8 +25,22 @@ rules:
...
+          - group:*@henrychang.mygbiz.com
+
+  - name: Prevent allUsers access to buckets via IAM
```
blueandgold: I would describe this name as:

```
Prevent public users from having access to buckets via IAM
```

red2k18: Done

rules/iam_rules.yaml
```diff
@@ -28,6 +28,20 @@ rules:
...
+        applies_to: self
+        resource_ids:
+          - '*'
+    inherit_from_parents: true
+    bindings:
+      - role: '*'
+        members:
+          - allUsers
+
+
```
blueandgold: One newline here only.

PR #1967, opened 2018-08-28
more APIs require enabling

rllin-fathom:

- [ ] All of the [unit-tests](https://forsetisecurity.org/docs/latest/develop/dev/testing.html) still pass.


install/gcp/installer/util/constants.py
```diff
@@ -136,8 +136,14 @@ class DeploymentStatus(Enum):
      'service': 'compute.googleapis.com'},
     {'name': 'Deployment Manager',
      'service': 'deploymentmanager.googleapis.com'},
+    {'name': 'Google Cloud Storage JSON API',
```
blueandgold: I feel that we should comment these... why are they being explicitly enabled if they are enabled by default in the normal way of creating a new project, so that in the future we would remember.

ahoying: GCP changed what APIs are enabled by default recently, so that broke our assumptions (and probably lots of other peoples)

ahoying: Here's the document, it seems to be correct: https://cloud.google.com/service-usage/docs/enabled-service. Note that this has changed a few times, so we should be explicit for every API required.

Here's what I show for a fresh project:

bigquery-json.googleapis.com      BigQuery API
cloudapis.googleapis.com          Google Cloud APIs
clouddebugger.googleapis.com      Stackdriver Debugger API
cloudtrace.googleapis.com         Stackdriver Trace API
datastore.googleapis.com          Cloud Datastore API
logging.googleapis.com            Stackdriver Logging API
monitoring.googleapis.com         Stackdriver Monitoring API
servicemanagement.googleapis.com  Service Management API
serviceusage.googleapis.com       Service Usage API
sql-component.googleapis.com      Google Cloud SQL
storage-api.googleapis.com        Google Cloud Storage JSON API
storage-component.googleapis.com  Google Cloud Storage

rllin-fathom: @blueandgold I agree with @ahoying that due to changes in default project we should be explicit.

Another reason that my use case raised is that we base all of our projects off a base GDM project template where minimal APIs are enabled (amongst other changes like disabling all firewall allowances) for security transparency.  Open to feedback about this though.

rllin-fathom:
@joecheuk added Kube API
@blueandgold added a comment

open to any further feedback!

install/gcp/installer/util/constants.py
blueandgold: Cool, sounds good.  As the default APIs might shift, there is probably no value to comment these.

PR #1969, opened 2018-08-28
Handle SSH failure gracefully during the installation process

### LINKED ISSUE #1947: Installer to handle ssh failure gracefully
If there is an ssh problem, we can output a message and skip the initialization step instead of kept on re trying.

joecheuk:
Resolves #1947


install/gcp/installer/util/gcloud.py
```diff
@@ -803,13 +805,20 @@ def check_vm_init_status(vm_name, zone):
...
         ['gcloud', 'compute', 'ssh', vm_name,
          '--zone', zone, '--command', check_script_executed, '--quiet'])
     # --quiet flag is needed to eliminate the prompting for user input
     # which will hang the run_command function
     # i.e. It will create a folder at ~/.ssh and generate a new ssh key
 
+    if err:
+        print('Error occurred when sshing to the VM, will skip waiting for '
+              'the instance to be initialized and proceed with the rest of '
+              'the installation.')
```
blueandgold: I would add a message to tell the user that they might need to verify the VM is actually initialized themselves.

```
'Error occurred when sshing to the VM.  Unable to verify the VM is fully initialized.'
'This is a non-fatal error, but you should verify the VM is fully initialized before use (add how).'
'Will skip waiting for the instance to be initialized and proceed with the rest of '
'the installation.'
```

install/gcp/installer/forseti_installer.py
```diff
@@ -307,10 +308,14 @@ def wait_until_vm_initialized(self, vm_name):
...
-            max_loading_time=constants.MAXIMUM_LOADING_TIME_IN_SECONDS,
-            exit_condition_checker=status_checker,
-            message=loading_message)
+        try:
+            _ = utils.start_loading(
+                max_loading_time=constants.MAXIMUM_LOADING_TIME_IN_SECONDS,
+                exit_condition_checker=status_checker,
+                message=loading_message)
+        except installer_errors.SSHError:
+            # Expected error, don't need to do anything.
```
blueandgold: It is good to have a comment here.  But I would clarify why this is an expected error, and why we don't need to do anything.

install/gcp/installer/util/gcloud.py
```diff
@@ -803,13 +805,20 @@ def check_vm_init_status(vm_name, zone):
...
-    _, out, _ = utils.run_command(
+    _, out, err = utils.run_command(
         ['gcloud', 'compute', 'ssh', vm_name,
          '--zone', zone, '--command', check_script_executed, '--quiet'])
     # --quiet flag is needed to eliminate the prompting for user input
     # which will hang the run_command function
     # i.e. It will create a folder at ~/.ssh and generate a new ssh key
 
+    if err:
+        print('Error occurred when sshing to the VM, will skip waiting for '
```
blueandgold: Should this be in constants?

install/gcp/installer/forseti_installer.py
joecheuk: Updated.

install/gcp/installer/util/gcloud.py
joecheuk: Updated.

install/gcp/installer/util/gcloud.py
joecheuk: Updated.

PR #1984, opened 2018-09-05
Update rules.md with new BigQuery syntax

### LINKED ISSUE #1937: Difficult to express specific rules for BigQuery

umairidris:
Partially fix #1937.

### DIFF
--- a/_docs/_latest/configure/scanner/rules.md
+++ b/_docs/_latest/configure/scanner/rules.md
@@ -335,21 +335,21 @@ rules:
 
 ### Rule definition
 
-BigQuery scanner rules serve as blacklists, for example:
+BigQuery scanner rules can be blacklists or whitelists, for example:
 
 ```yaml
 rules:
   - name: sample BigQuery rule to search for public datasets
-    dataset_id: '*'
-    special_group: 'allAuthenticatedUsers'
-    user_email: '*'
-    domain: '*'
-    group_email: '*'
-    role: '*'
+    mode: blacklist
     resource:
       - type: organization
         resource_ids:
           - YOUR_ORG_ID / YOUR_PROJECT_ID
+    dataset_ids: ['*']
+    bindings:
+      - role: '*'
+        members:
+          - special_group: 'allAuthenticatedUsers'
 ```
 
 * `name`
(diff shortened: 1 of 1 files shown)


_docs/_latest/configure/scanner/rules.md
```diff
@@ -365,35 +365,58 @@ rules:
...
+        * `user_email`
+          * **Description**: User email.
+          * **Valid values**: String.
+        * `special_group`
+          * **Description**: Special group.
+          * **Valid values**: String.
+
 * `special_group`
-  * **Description**: The special group.
+  * **Description**: The special group. ***DEPRECATED, prefer to use bindings.***
```
joecheuk: What do you think about "DEPRECATED, please use bindings instead" ?

umairidris: Done.

PR #1989, opened 2018-09-06
Stop matching unset member fields in BigQuery ACLs 

umairidris:
This is the final part of fixing #1937.

Currently, we cannot blacklist all values of a member (if set). For example

...
members:
  - user_email: *
...

to blacklist all users. In Iam rules, member "user:*" does blacklist all users. 

This PR makes the behaviour consistent by not defaulting values to None and not checking fields that were unset.

This also uncovered a bug in my previous code which has made old style configs not work correctly. By using a dict to map rule regex to values, having two rules with the same regex would have one overriding the other. Now, all fields are checked as they should be.


umairidris:
Note: this PR also contains #1986. Once that is submitted I will rebase to remove the unrelated changes.

umairidris:
Rebased on dev so should no longer see unrelated changes.

PR #1990, opened 2018-09-06
kubernetes rule updated to address the latest vulnerabilities 

joecheuk:
CVE-2018-5390 describes a kernel-level networking vulnerability that increases the effectiveness of denial of service (DoS) attacks against vulnerable systems over TCP connections.

CVE-2018-5391 describes a kernel-level networking vulnerability that increases the effectiveness of denial of service (DoS) attacks against vulnerable systems over IP connections.


rules/ke_rules.yaml
```diff
@@ -61,13 +61,25 @@ rules:
     check_serverconfig_valid_master_versions: false
     allowed_nodepool_versions:
       - major: '1.8'
-        minor: '12-gke.1'
+        minor: '10-gke.2'
+        operator: '='
```
blueandgold: Can we do a `>=` operator here, so that it will capture `1.8.12-gke.3` and not repeat the major versions?

blueandgold: Same with the ones below too.

blueandgold: After clarifying this offline, can we just a comment here to explain the gap, so that it's easier to understand?

PR #1992, opened 2018-09-07
Adjust Dockerfiles to speed up builds during development

rvandegrift:
Currently, any change to the forseti code causes a rebuild of both docker images.  This is slow, since docker can't cache intermediate work.  This PR fixes that by only copying the requirements files when they are used.  Copying the full source tree is delayed to the end.

### DIFF
--- a/install/docker/base
+++ b/install/docker/base
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Do not add forseti source in this file to avoid rebuilding base during development.
+
 FROM ubuntu:18.04
 
-# Add only the apt dependencies to avoid rebuilding base during development
+# Add the apt dependencies
 ADD install/dependencies/apt_packages.txt /tmp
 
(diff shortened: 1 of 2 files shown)


install/docker/base
```diff
@@ -14,13 +14,12 @@
 
 FROM ubuntu:18.04
 
-# Expose our source so we can read in dependencies.
-ADD . /forseti-security/
-WORKDIR /forseti-security/
+# Add only the apt dependencies to avoid rebuilding base during development
```
blueandgold: This is a super tiny nit, but if we are going to comment about avoid rebuilding during development, we should move it out of this comment line and put it at the top-level of this file.

```
# Do not add forseti source in this file to avoid rebuilding base during development.

FROM ubuntu:18.04

# Add the apt dependencies.
ADD install/dependencies/apt_packages.txt /tmp
```


```

rvandegrift: Makes sense to me, fixed

PR #2005, opened 2018-09-12
Clean up and improve firewall rules

red2k18:
- [x] Added functionality to delete duplicate firewall rules
- [x] Created service account rule to allow users to SSH into client
- [x] Created service account rule to deny all ingress which overrides the implied rule for deny
- [x] Refactored Installer code


install/gcp/installer/forseti_server_installer.py
```diff
@@ -36,6 +35,7 @@ class ForsetiServerInstaller(ForsetiInstaller):
     access_target = None
     target_id = None
     user_can_grant_roles = True
+    firewall_rules_to_be_deleted = ["default-allow-icmp", "default-allow-internal", "default-allow-rdp", "default-allow-ssh"]
```
blueandgold: This exceeds the 80-char limiit.

You can do:
```
firewall_rules_to_be_deleted = [
    'default-allow-icmp',
    'default-allow-internal',
    'default-allow-rdp',
    'default-allow-ssh',
]
```

blueandgold: Is `ForsetiServerInstaller` the best place to maintain these firewalls? Since these are not specific to the server. How about moving them to the `ForsetiInstaller` class?

install/gcp/installer/forseti_server_installer.py
```diff
@@ -145,6 +148,28 @@ def create_firewall_rules(self):
             self.config.vpc_host_network,
             '0.0.0.0/0')
 
+        # Create firewall rule to open only port tcp:22 (ssh)
+        # to all the external traffics from the internet to ssh into client VM.
+        gcloud.create_firewall_rule(
+            self.format_firewall_rule_name(
+                'forseti-client-allow-ssh-external'),
+            [self.gcp_service_acct_email],
```
blueandgold: This looks like the server's service account, not the client.

Also, should this be somewhere else?  What about the client installer?  How do you think?

install/gcp/installer/forseti_server_installer.py
red2k18: Done

red2k18: Was thinking the same. Moved it to `ForsetiInstaller` class

install/gcp/installer/forseti_server_installer.py
```diff
@@ -24,7 +24,6 @@
 from util import gcloud
 from util import utils
```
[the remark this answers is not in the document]
red2k18: Ok

install/gcp/installer/forseti_client_installer.py
```diff
@@ -94,3 +96,38 @@ def get_deployment_values(self):
...
+    def format_firewall_rule_name(self, rule_name):
+        """Format firewall rule name.
+
+        Args:
+            rule_name (str): Name of the firewall rule.
+
+        Returns:
+            str: Firewall rule name.
+        """
+        return '{}-{}'.format(rule_name, self.config.identifier)
```
blueandgold: This is repeated from the method in the server installer, and should be refactored.

install/gcp/installer/forseti_server_installer.py
```diff
@@ -145,17 +153,15 @@ def create_firewall_rules(self):
...
-        return '{}-{}'.format(rule_name, self.config.identifier)
+    def delete_firewall_rules(self):
+        """Deletes default firewall rules as the forseti service account rules
+        serves the purpose"""
+        if len(self.firewall_rules_to_be_deleted) > 0:
+            for rule in self.firewall_rules_to_be_deleted:
+                gcloud.delete_firewall_rule(rule)
+            print('Firewall rules were successfully deleted')
+        else:
+            print('Couldn\'t find rules to delete')
```
blueandgold: This message would be better as: `No firewall rules to delete.`  Thinking on this further, do we really need to display this to the user?

install/gcp/installer/forseti_server_installer.py
```diff
@@ -145,17 +153,15 @@ def create_firewall_rules(self):
...
-            str: Firewall rule name.
-        """
-        return '{}-{}'.format(rule_name, self.config.identifier)
+    def delete_firewall_rules(self):
+        """Deletes default firewall rules as the forseti service account rules
+        serves the purpose"""
+        if len(self.firewall_rules_to_be_deleted) > 0:
+            for rule in self.firewall_rules_to_be_deleted:
+                gcloud.delete_firewall_rule(rule)
+            print('Firewall rules were successfully deleted')
```
blueandgold: I think that we should include what firewall rules where deleted for the users to see.

install/gcp/installer/forseti_server_installer.py
```diff
@@ -101,16 +106,19 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
             instance_name = 'forseti-{}-vm-{}'.format(
                 self.config.installation_type,
                 self.config.identifier)
-            self.wait_until_vm_initialized(instance_name)
 
             # Create firewall rules.
             self.create_firewall_rules()
+            self.wait_until_vm_initialized(instance_name)
```
blueandgold: Why is this moved?

install/gcp/installer/forseti_server_installer.py
```diff
@@ -101,16 +106,19 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
...
+
+            # Delete firewall rules.
+            self.delete_firewall_rules()
```
blueandgold: These default rules are not server specific, and should be moved outside of the server installer.

install/gcp/installer/forseti_client_installer.py
```diff
@@ -94,3 +96,38 @@ def get_deployment_values(self):
...
+                'forseti-client-allow-ssh-external'),
+            [self.gcp_service_acct_email],
+            constants.FirewallRuleAction.ALLOW,
+            ['tcp:22'],
+            constants.FirewallRuleDirection.INGRESS,
+            0,
+            self.config.vpc_host_network,
+            '0.0.0.0/0')
+
+        # Rule to block out all the ingress traffic on client VM.
```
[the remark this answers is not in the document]
red2k18: Done

install/gcp/installer/forseti_server_installer.py
red2k18: I think we should let the user know whether we deleted it or not.

install/gcp/installer/forseti_server_installer.py
red2k18: Consider it done.

install/gcp/installer/forseti_server_installer.py
red2k18: It gives VM initialization error if logic to delete default rules is in forseti_installer.py because the rules will be deleted before the server and the client rules are created.

install/gcp/installer/forseti_server_installer.py
red2k18: The service account rule to ssh into the VM needs to be created so VM initialization is successful.

install/gcp/installer/forseti_client_installer.py
red2k18: Server installer doesn't have create or delete functionality now. Can be refactored in the future though.

install/gcp/installer/forseti_server_installer.py
red2k18: On second thoughts, I deleted the print statement in the else block as it is understood that the rules weren't deleted if they don't see the deleted message.

install/gcp/installer/forseti_client_installer.py
```diff
@@ -15,8 +15,8 @@
 """Forseti CLI installer."""
 
 from forseti_installer import ForsetiInstaller
-
 from util import gcloud
+from util import constants
```
blueandgold: alphasort this import

install/gcp/installer/forseti_server_installer.py
blueandgold: Ah, that makes sense.  Great move!

install/gcp/installer/forseti_server_installer.py
blueandgold: Hmmm, okay.  I suppose we can can say that the correct firewalls are created only if the server is updated.

Otherwise, the relevant place to do this more centrally would be here.
https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/install/gcp/gcp_installer.py#L122-L137

install/gcp/installer/forseti_server_installer.py
```diff
@@ -101,16 +106,22 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
...
 
         return success, deployment_name
 
     def create_firewall_rules(self):
         """Create firewall rules for Forseti server instance."""
-        # Rule to block out all the ingress traffic.
+        # Rule to block out all the ingress traffic on server VM.
+        # We need a service account deny for ingress that is more
```
blueandgold: How does this rephrase look to you?
```
        # This rule overrides the implied deny for ingress
        # because we want a rule with a higher priority
        # that would be harder to be overridden.
```

red2k18: We need to mention the second reason as well, changed to below
# This rule overrides the implied deny for ingress
# because it has higher priority and is specific to service account.

red2k18: We need to mention the second reason as well, changed to below
   This rule overrides the implied deny for ingress
   because it has higher priority and is specific to service account.

install/gcp/installer/forseti_client_installer.py
red2k18: Consider it done.

install/gcp/installer/forseti_server_installer.py
red2k18: Based on the flow of control, I have put the logic in server as all the service account rules has to be created before default rules are deleted.

install/gcp/installer/forseti_client_installer.py
blueandgold: Okay

install/gcp/installer/forseti_server_installer.py
blueandgold: Okay

install/gcp/installer/forseti_server_installer.py
```diff
@@ -101,16 +106,20 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
...
+            # Delete firewall rules.
+            self.delete_firewall_rules()
 
         return success, deployment_name
 
     def create_firewall_rules(self):
         """Create firewall rules for Forseti server instance."""
-        # Rule to block out all the ingress traffic.
+        # This rule overrides the implied deny for ingress
+        # because it has higher priority and is specific to service account.
```
blueandgold: I want to add to this comment, the "why", which is to make it harder to override.  Can you please add this?  Here and in the client?

install/gcp/installer/forseti_client_installer.py
```diff
@@ -94,3 +97,40 @@ def get_deployment_values(self):
...
+        # Create firewall rule to open only port tcp:22 (ssh)
+        # to all the external traffic from the internet to ssh into client VM.
+        gcloud.create_firewall_rule(
+            self.format_firewall_rule_name(
+                'forseti-client-allow-ssh-external'),
+            [self.gcp_service_acct_email],
+            constants.FirewallRuleAction.ALLOW,
+            ['tcp:22'],
+            constants.FirewallRuleDirection.INGRESS,
+            0,
```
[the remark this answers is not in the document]
red2k18: Done

install/gcp/installer/forseti_client_installer.py
```diff
@@ -94,3 +97,40 @@ def get_deployment_values(self):
...
+            '0.0.0.0/0')
+
+        # This rule overrides the implied deny for ingress
+        # because it is specific to service account with a higher priority
+        # that would be harder to be overriden.
+        gcloud.create_firewall_rule(
+            self.format_firewall_rule_name('forseti-client-deny-all'),
+            [self.gcp_service_acct_email],
+            constants.FirewallRuleAction.DENY,
+            ['icmp', 'udp', 'tcp'],
```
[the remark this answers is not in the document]
red2k18: Done

PR #2006, opened 2018-09-12
Properly trap and return errors correctly in travis scripts

mirons-google:
This should ensure that if any command in any of the bash script test runners fail we still report the return code correctly, not just the last one.

This also defaults the docker commands back to a default log level.


mirons-google:
I amended this PR with changes to two offending files to have them pass lint.

PR #2011, opened 2018-09-14
Add project Liens to inventory

### LINKED ISSUE #2010: Add support for project liens
We would like Forseti to be able to check for the existence and correct configuration of project liens.

umairidris:
Partially fix #2010.


umairidris:
Added tests and regenerated db

google/cloud/forseti/services/inventory/base/resources.py
```diff
@@ -1210,6 +1210,27 @@ def type():
...
+class Lien(Resource):
+    """The Resource implementation for Lien"""
+
+    def key(self):
+        """Get key of this resource
+
+        Returns:
+            str: key of this resource
+        """
+        return self['name'].split('/')[-1]
```
ahoying: If the same lien name is used on multiple projects, this will break the assumption that the key field is always unique. I think this needs to use a hash like the appengine resources to ensure uniqueness.

umairidris: It seems lien names are unique and a single lien can only be applied to a single project.

https://cloud.google.com/resource-manager/reference/rest/v1/liens
name: A system-generated unique identifier for this Lien.

https://cloud.google.com/resource-manager/reference/rest/v1/liens/create
Create a Lien which applies to the resource denoted by the parent field.
Setting a name causes a field violation:
{
  "field": "name",
  "description": "must not be set for a new Lien"
}

In this case do we still to use the hash?

ahoying: Hrm...it's not clear in what context it is unique. Is it guaranteed globally unique, unique only within the project? I wish the documentation was a bit better. I've reached out to the team that owns the service to find out the answer.

PR #2016, opened 2018-09-18
[WIP] Document the new sequential step-upgrade process.

blueandgold:
This is to document how we will support the upgrade process by doing sequentially doing step-upgrades of through each minor version.

### DIFF
--- a/_docs/_latest/setup/upgrade.md
+++ b/_docs/_latest/setup/upgrade.md
@@ -7,6 +7,13 @@ order: 002
 
 This guide explains how to upgrade your Forseti instance.
 
+For version 2.0.0 and later, we will provide upgrade instructions from one minor
+version to the next minor version. This means, if you want to upgrade from
+version 2.2.0 to 2.4.0, you should follow the upgrade instruction from
+version 2.2.0 to 2.3.0 first, and then follow the upgrade instruction from
+version 2.3.0 to 2.4.0.  This ensures the upgrade process is easier to manage
+and to test.
+
 ---
 
 {% capture 1x_upgrade %}
@@ -153,6 +160,31 @@ Learn more about [Updating a Deployment](https://cloud.google.com/deployment-man
 
 {% endcapture %}
 {% include site/zippy/item.html title="Upgrading 2.X installations" content=2x_upgrade uid=1 %}
+
+{% capture upgrading_2_0_0_to_2_1_0 %}
+TBD
+{% endcapture %}
+{% include site/zippy/item.html title="Upgrading 2.0.0 to 2.1.0" content=2_0_0_to_2_1_0 uid=1 %}
+
+{% capture upgrading_2_1_0_to_2_2_0 %}
+TBD
+{% endcapture %}
+{% include site/zippy/item.html title="Upgrading 2.1.0 to 2.2.0" content=2_1_0_to_2_2_0 uid=1 %}
+
+{% capture upgrading_2_2_0_to_2_3_0 %}
+TBD
+{% endcapture %}
+{% include site/zippy/item.html title="Upgrading 2.2.0 to 2.3.0" content=2_2_0_to_2_3_0 uid=1 %}
+
+{% capture upgrading_2_3_0_to_2_4_0 %}
+2.4.0 does not contain any schema changes in deployment manager templates,
+server configurations, or scanner rule files.
+
+You can update by modifying `forseti-version` in your server and client
+deployment manager templates to version 2.4.0, and run the [deployment manager
+update command](https://forsetisecurity.org/docs/latest/setup/upgrade.html#zippy-1).
+{% endcapture %}
+{% include site/zippy/item.html title="Upgrading 2.3.0 to 2.4.0" content=2_3_0_to_2_4_0 uid=1 %}
 
 ## What's next
 
--- a/_docs/_latest/setup/upgrade.md
+++ b/_docs/_latest/setup/upgrade.md
@@ -158,33 +158,176 @@ change a certain resource, you'll need to create a new deployment of Forseti.
 
 Learn more about [Updating a Deployment](https://cloud.google.com/deployment-manager/docs/deployments/updating-deployments).
 
-{% endcapture %}
-{% include site/zippy/item.html title="Upgrading 2.X installations" content=2x_upgrade uid=1 %}
-
 {% capture upgrading_2_0_0_to_2_1_0 %}
-TBD
+
+1. Open cloud shell when you are in the Forseti project on GCP.
+1. Checkout forseti with tag v2.1.0 by running the following commands:
+    1. If you already have the forseti-security folder under your cloud shell directory, 
+    run command `rm -rf forseti-security` to delete the folder.
+    1. Run command `git clone` to clone the forseti-security directory to cloud shell.
+    1. Run command `cd forseti-security` to navigate to the forseti-security directory.
+    1. Run command `git checkout tags/v2.1.0` to checkout version `v2.1.0` of Forseti Security.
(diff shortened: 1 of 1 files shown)


_docs/_latest/setup/upgrade.md
```diff
@@ -151,6 +158,173 @@ change a certain resource, you'll need to create a new deployment of Forseti.
 
 Learn more about [Updating a Deployment](https://cloud.google.com/deployment-manager/docs/deployments/updating-deployments).
 
+{% capture upgrading_2_0_0_to_2_1_0 %}
+
+1. Open cloud shell when you are in the Forseti project on GCP.
+1. Checkout forseti with tag v2.1.0 by running the following commands:
+    1. If you already have the forseti-security folder under your cloud shell directory, 
+    run command `rm -rf forseti-security` to delete the folder.
+    1. Run command `git clone` to clone the forseti-security directory to cloud shell.
```
blueandgold: Do you feel we should give the full instruction here?  It seems odd that this is abbreviated here, but the full commands are given elsewhere.

_docs/_latest/setup/upgrade.md
```diff
@@ -151,6 +158,173 @@ change a certain resource, you'll need to create a new deployment of Forseti.
...
+    1. Run command `cd forseti-security` to navigate to the forseti-security directory.
+    1. Run command `git checkout tags/v2.1.0` to checkout version `v2.1.0` of Forseti Security.
+1. Download the latest copy of your Forseti server deployment template file from the Forseti server GCS 
+bucket (located under `forseti-server-xxxxxx/deployment_templates`).
+1. Update the `forseti-version` inside the deployment template to `v2.1.0`.
+1. Create file `forseti_server_v2_1_0.yaml` under `forseti-security/deployment-templates`.
+1. Copy and paste the content of the deployment template to file `forseti_server_v2_1_0.yaml`.
```
blueandgold: Instead of copy/paste the content, how about just rename the file that is downloaded to forseti_server_v2_1_0?

_docs/_latest/setup/upgrade.md
```diff
@@ -151,6 +158,173 @@ change a certain resource, you'll need to create a new deployment of Forseti.
...
+1. Upload file `forseti_server_v2_1_0.yaml` back to the GCS bucket (`forseti-server-xxxxxx/deployment_templates`)
```
blueandgold: Do we tell users what and how to update any changes in the `forseti_conf_server.yaml`?  Which in 2.1.0 is the cscc notifier.  Or other rule file changes?

_docs/_latest/setup/upgrade.md
```diff
@@ -151,6 +158,173 @@ change a certain resource, you'll need to create a new deployment of Forseti.
...
+1. Checkout forseti with tag v2.1.0 by running the following commands:
+    1. If you already have the forseti-security folder under your cloud shell directory, 
+    run command `rm -rf forseti-security` to delete the folder.
+    1. Run command `git clone` to clone the forseti-security directory to cloud shell.
+    1. Run command `cd forseti-security` to navigate to the forseti-security directory.
+    1. Run command `git checkout tags/v2.1.0` to checkout version `v2.1.0` of Forseti Security.
+1. Download the latest copy of your Forseti server deployment template file from the Forseti server GCS 
+bucket (located under `forseti-server-xxxxxx/deployment_templates`).
+1. Update the `forseti-version` inside the deployment template to `v2.1.0`.
+1. Create file `forseti_server_v2_1_0.yaml` under `forseti-security/deployment-templates`.
```
blueandgold: The `forseti_server_v2_1_0.yaml`  syntax is a departure from the existing `deploy-forseti-server-xxxxx.yaml`.  is it because of character limit?  
Also, `forseti_server_v2_1_0.yaml`  is very similar to `forseti_server_conf.yaml` file.
Should we also make changes to the Installer to generate the new file name?

_docs/_latest/setup/upgrade.md
```diff
@@ -151,6 +158,271 @@ change a certain resource, you'll need to create a new deployment of Forseti.
...
+
+1. Open cloud shell when you are in the Forseti project on GCP.
+1. Checkout forseti with tag v2.1.0 by running the following commands:
+    1. If you already have the forseti-security folder under your cloud shell directory, 
+    run command `rm -rf forseti-security` to delete the folder.
+    1. Run command `git clone https://github.com/GoogleCloudPlatform/forseti-security.git` to 
+    clone the forseti-security directory to cloud shell.
+    1. Run command `cd forseti-security` to navigate to the forseti-security directory.
+    1. Run command `git checkout tags/v2.1.0` to checkout version `v2.1.0` of Forseti Security.
+1. Download the latest copy of your Forseti server deployment template file from the Forseti server GCS
```
blueandgold: We should mention where the downloaded the file should be placed.  In the `forseti-security/deployment-templates` directory?

_docs/_latest/setup/upgrade.md
```diff
@@ -151,6 +158,271 @@ change a certain resource, you'll need to create a new deployment of Forseti.
...
+    run command `rm -rf forseti-security` to delete the folder.
+    1. Run command `git clone https://github.com/GoogleCloudPlatform/forseti-security.git` to 
+    clone the forseti-security directory to cloud shell.
+    1. Run command `cd forseti-security` to navigate to the forseti-security directory.
+    1. Run command `git checkout tags/v2.1.0` to checkout version `v2.1.0` of Forseti Security.
+1. Download the latest copy of your Forseti server deployment template file from the Forseti server GCS 
+bucket (located under `forseti-server-xxxxxx/deployment_templates`).
+1. Update the `forseti-version` inside the deployment template to `tags/v2.1.0`.
+1. Create file `deploy-forseti-server-xxxxx-2-1-0.yaml` under `forseti-security/deployment-templates`.
+1. Copy and paste the content of the deployment template to file `deploy-forseti-server-xxxxx-2-1-0.yaml`.
```
blueandgold: Per offline discussion, it would be better to just use gsutil command which can create the new file with the correct name in one shot in cloud shell env, and save the copy and paste step from local env to cloud shell.

_docs/_latest/setup/upgrade.md
```diff
@@ -151,6 +158,271 @@ change a certain resource, you'll need to create a new deployment of Forseti.
...
+        
+    ```
+1. Update the `forseti-version` inside the deployment template to `tags/v2.2.0`.
+1. Create file `deploy-forseti-server-xxxxx-2-2-0.yaml` under `forseti-security/deployment-templates`.
+1. Copy and paste the content of the deployment template to file `deploy-forseti-server-xxxxx-2-2-0.yaml`.
+1. Upload file `deploy-forseti-server-xxxxx-2-2-0.yaml` back to the GCS bucket (`forseti-server-xxxxxx/deployment_templates`)
+1. Navigate to [Deployment Manager](https://console.cloud.google.com/dm/deployments) and 
+copy the deployment name for Forseti server.
+1. Run command `gcloud deployment-manager deployments update DEPLOYMENT_NAME --config deploy-forseti-server-xxxxx-2-2-0.yaml`  
+If you see errors while running the deployment manager update command, please refer to below section
```
blueandgold: This is not an `if`, as they will see this problem for sure.  So, how about moving the section from below, and put them here?

joecheuk: The workaround we put in before in the upgrade instructions will  get user to v2.2.0 without updating the vm instance (staying on the old branch to do the update). They will still have the same problem updating to v2.3.0/v2.4.0 that's why I am trying to move the error instruction out so it can be used for other versions.

PR #2017, opened 2018-09-19
[docs] Recommend latest minor version release in setup docs

erikwebb:
The docs for versions >v2.0 still recommend using `v2.0.0` in their setup guide. This PR updates more recent versions to use their respective minor release.


erikwebb:
I'm also tempted to update `scripts/create_new_version_from_latest.sh` to update this reference automatically, but not sure of the most maintainable way to do this.

_docs/v2.1/setup/install.md
```diff
@@ -51,14 +51,14 @@ steps below:
       git clone https://github.com/GoogleCloudPlatform/forseti-security.git
       ```
 
-  1. Check out the specific version of Forseti you want to install by using a tag like `v2.0.0.`:
+  1. Check out the specific version of Forseti you want to install by using a tag like `v2.1.0.`:
```
blueandgold: Hmmm, this was meant to serve as a generic example.  But I can see how there can be a problem for copy/paste.  However, this change would force us to update the version here on every release.  So how about this, like below?
```
Check out the specific version of Forseti you want to install by using a tag like `v2.1.0, v2.2.0, v2.3.0, etc.`.  You can visit the release page (add link) to see which versions are available to install:

git checkout tags/v<X.Y.Z>
```

erikwebb: That makes sense to me. I was actually hoping to use another `sed` command in `create_new_version_from_latest.sh` to change this to the latest version automatically. I couldn't think of a way that didn't end up with a nasty `%VERSION%` macro or something like that.

blueandgold: Cool, will you change the PR here to make it more generic?

erikwebb: Will do. Try to get to it later today.

erikwebb:
@blueandgold Rather than overthinking this more, I just changed the line to read `git checkout tags/<tag_number>` like in `_docs/_latest/setup/manual.md` already. I squashed into a single commit also.

_docs/_latest/setup/install.md
```diff
@@ -60,7 +60,7 @@ steps below:
       # If the tag exists in the remote repository but you are unable to checkout the tag,
       # run command `git fetch --all` to fetch all the latest branch/tag information and run
       # the checkout command again.
-      git checkout tags/v2.0.0
+      git checkout tags/<tag_number>
```
blueandgold: I think that should include the small `v` as that is part of the whole tag (v2.0.0, v2.1.0).  I am worried that people might miss that, when they paste in just the version number, i.e. 2.0.0.
```
git checkout tags/v<tag_number>
```
i.e. it will be
```
git checkout tags/v2.0.0
```

erikwebb:
Added new changes to the v2.5 folder.

PR #2018, opened 2018-09-19
[docs] Add missing cd command to setup guide

erikwebb:
Copying install commands as-is misses the step to `cd` into the
freshly cloned `forseti-security` directory after a `git clone`. This
PR adds that steps for easier copy-and-paste setup.

### DIFF
--- a/_docs/_latest/setup/install.md
+++ b/_docs/_latest/setup/install.md
@@ -49,13 +49,14 @@ steps below:
 
       ```bash
       git clone https://github.com/GoogleCloudPlatform/forseti-security.git
-      cd forseti-security
       ```
 
   1. Check out the specific version of Forseti you want to install by using a tag like `v2.0.0.`:
 
       ```bash
       # Make sure you are in the forseti-security folder.
+      cd forseti-security
+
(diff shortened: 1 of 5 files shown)


_docs/_latest/setup/install.md
```diff
@@ -49,6 +49,7 @@ steps below:
 
       ```bash
       git clone https://github.com/GoogleCloudPlatform/forseti-security.git
+      cd forseti-security
```
blueandgold: This step is focused on how to download the source code.  
So, the `cd` step should be better in the next next section, closer to the comment about `Make sure you are in the forseti-security folder`.

How about this?
```
       # Make sure you are in the forseti-security folder.
       # If the tag exists in the remote repository but you are unable to checkout the tag,
       # run command `git fetch --all` to fetch all the latest branch/tag information and run
       # the checkout command again.
       cd forseti-security
       git checkout tags/v2.0.0
```

erikwebb:
@blueandgold I moved it into the next section. A bit different than you requested, but I think it makes the most sense.

Issue #2023, opened 2018-09-20
Consider adding version number into logs.


blueandgold:
When troubleshooting the logs, the question that always come up is what what version of Forseti is being used.

So to make this easier, we could consider to put the forseti version into the log statements.

Current:
```
2018-09-14 03:28:52,591 WARNING [forseti-security] google.cloud.forseti.scanner.audit.blacklist_rules_engine(find_violations)
```

Proposed:
2018-09-14 03:28:52,591 WARNING [forseti-security-2.3.0] google.cloud.forseti.scanner.audit.blacklist_rules_engine(find_violations)

jjawad-google:
[forseti-security] [v2.3.0] instead of [forseti-security-2.3.0] so that existing log parsing is not broken.

PR #2037, opened 2018-09-24
Adding version number into logs

red2k18:
This resolves issue #2023

### DIFF
--- a/google/cloud/forseti/common/util/logger.py
+++ b/google/cloud/forseti/common/util/logger.py
@@ -21,11 +21,12 @@ import logging
 import logging.handlers
 import os
 
+from google.cloud.forseti import __version__
 
 DEFAULT_LOG_FMT = ('%(asctime)s %(levelname)s '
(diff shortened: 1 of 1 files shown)


google/cloud/forseti/common/util/logger.py
```diff
@@ -21,11 +21,12 @@
 import logging.handlers
 import os
 
+from google.cloud.forseti import __version__
 
 DEFAULT_LOG_FMT = ('%(asctime)s %(levelname)s '
                    '%(name)s(%(funcName)s): %(message).1024s')
 
-SYSLOG_LOG_FMT = ('%(levelname)s [forseti-security] '
+SYSLOG_LOG_FMT = ('%(levelname)s [forseti-security] %(__version__)s '
```
blueandgold: We said we would put `[ ]` around the version.
`[forseti-security] [v2.3.0]`

red2k18: Consider it done.

PR #2044, opened 2018-09-25
Model Delete returns result 

red2k18:


google/cloud/forseti/services/model/model.proto
```diff
@@ -45,10 +45,7 @@ message DeleteModelRequest {
 }
 
 message DeleteModelReply {
-  enum Status {
-    SUCCESS = 0;
-  }
-  Status status = 1;
+  string result = 1;
```
blueandgold: Instead of deleting the status field, can you please keep it and add a `FAIL = 1`.
Can you also rename the `result` field, call it `message = 2`, and use it to send details of the operation?  model name of what has been deleted, or what exception has occurred.

google/cloud/forseti/services/model/modeller.py
```diff
@@ -103,8 +103,20 @@ def delete_model(self, model_name):
...
+            result (str): Success or failure message based on whether model is
+            deleted or not
         """
 
         LOGGER.info('Deleting model: %s', model_name)
         model_manager = self.config.model_manager
-        model_manager.delete(model_name)
+        try:
+            model_manager.delete(model_name)
+            result = 'SUCCESS'
```
blueandgold: No need to have a return or the `finally`, just raise an exception and handle it from the caller method.

google/cloud/forseti/services/model/modeller.py
```diff
@@ -103,8 +103,20 @@ def delete_model(self, model_name):
...
+        except Exception as e:
+            result = 'FAILED'
+            LOGGER.exception(e)
```
blueandgold: Add a description, and drop the `e`... that will be captured :
```
LOGGER.exception('Unable to delete model: %s', model_name)
```

google/cloud/forseti/services/model/modeller.py
```diff
@@ -103,8 +103,20 @@ def delete_model(self, model_name):
...
+            raise
```
blueandgold: You should raise a custom exception here.
```
raise ModelDeletionError()
```

google/cloud/forseti/services/model/modeller.py
```diff
@@ -103,8 +103,20 @@ def delete_model(self, model_name):
...
+        finally:
```
blueandgold: Can drop this finally since we are not returning.

google/cloud/forseti/services/model/service.py
```diff
@@ -103,8 +103,9 @@ def DeleteModel(self, request, _):
         """
 
         model_name = request.handle
-        self.modeller.delete_model(model_name)
-        return model_pb2.DeleteModelReply()
+        result = self.modeller.delete_model(model_name)
```
blueandgold: Add a try / except here, handing `ModelDeletionError()`.
```
try:
    self.modeller.delete_model(model_name)
    reply = model_pb2.DeleteModelReply(status=SUCCESS_ENUM, message='Successfully deleted <model_name>)
except ModelDeletionError:
    reply = model_pb2.DeleteModelReply(status=FAIL_ENUM, message='Unable to delete <model_name>)

return reply
```

google/cloud/forseti/services/model/service.py
```diff
@@ -103,8 +103,15 @@ def DeleteModel(self, request, _):
         """
 
         model_name = request.handle
-        self.modeller.delete_model(model_name)
-        return model_pb2.DeleteModelReply()
+        try:
+            self.modeller.delete_model(model_name)
+            success = model_pb2.DeleteModelReply.Status.Value('SUCCESS')
```
blueandgold: nit: i would change this line:
```
status = model_pb2.DeleteModelReply.Status.Value('SUCCESS')
```
So that you can save two lines of code like this:
```
try:
    self.modeller.delete_model(model_name)
    status = model_pb2.DeleteModelReply.Status.Value('SUCCESS')
except Exception:
    LOGGER.exception('Unable to delete model: %s', model_name)
    status = model_pb2.DeleteModelReply.Status.Value('FAIL')

return model_pb2.DeleteModelReply(status=status)
```

red2k18: Done

Issue #2076, opened 2018-10-04
New feature bucket retention


gruihuang:
Forseti should check the life cycle and TTL for GCS buckets. e.g. For buckets holding logs, we want to make sure TTL is at least X days.

jjawad-google:
This is a great idea, would it be possible for you to contribute?

Issue #1962, opened 2018-08-28
Specify schema of YAML files


umairidris:
Today, it is not possible to know the schema of the rules yaml files unless one looks at examples or digs into the code.

It would be great if a schema such as https://github.com/GoogleCloudPlatform/healthcare/blob/master/deploy/templates/data_project.py.schema could be defined.

umairidris:
Possible solution: convert to json dict and use https://github.com/Julian/jsonschema.

PR #2077, opened 2018-10-04
Bucket retention

gruihuang:
New feature "bucket retention".
[#2076](https://github.com/GoogleCloudPlatform/forseti-security/issues/2076)


google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+            rule_index (int): The index of the rule from the rule definitions.
+                Assigned automatically when the rule book is built.
+        """
+        self._rules_sema.acquire()
+        try:
+            applies_to = rule_def.get('applies_to', None)
+            if applies_to is None:
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Lack of applies_to in rule {}'.format(rule_index))
+            if type(applies_to).__name__ != 'list':
```
dufton: We generally haven't been checking types of rule fields, so you can remove the check. I think type checks would best be done with a schema file: #1962 .

Also, the check can be written as: `isinstance(applies_to, list)`

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Miss dash (-) near applies_to in rule {}'.format(
+                        rule_index))
+
+            minimum_retention = rule_def.get('minimum_retention', None)
+            maximum_retention = rule_def.get('maximum_retention', None)
+            if minimum_retention is None and maximum_retention is None:
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Lack of minimum_retention and \
+maximum_retention in rule {}'.format(rule_index))
```
dufton: Use implicit line joining for long strings:

```
raise audit_errors.InvalidRulesSchemaError(
    'Lack of minimum_retention and '
    'maximum_retention in rule {}'.format(rule_index))
```

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+            elif minimum_retention != None and maximum_retention != None:
+                if minimum_retention > maximum_retention:
+                    raise audit_errors.InvalidRulesSchemaError(
+                        'minimum_retention larger than \
+maximum_retention in rule {}'.format(rule_index))
+
+            applies_to_history = {}
```
dufton: You can find duplicates in a single line:

```
if any(applies_to.count(x) > 1 for x in applies_to):
```

You can also check the types are valid:

```
if any(x not in _APPLY_TO_RESOURCES for x in applies_to):
```

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+                    self.add_bucket_rule(
+                        rule_def,
+                        rule_index,
+                        minimum_retention,
+                        maximum_retention)
+                # other appto add here
+        finally:
+            self._rules_sema.release()
+
+    def add_bucket_rule(self, rule_def, rule_index,
```
dufton: The rules for each applies_to type will probably look the same, so you can make this function general (with applies_to as a param) and then have

```
self.resource_rules_map[applies_to][gcp_resource].add(rule)
```

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+            for rid in resource_ids:
+                if rid == '*':
+                    raise audit_errors.InvalidRulesSchemaError(
+                        'The symbol * is not allowed in rule {}'.format(
+                            rule_index))
+
+                rule = Rule(rule_name=rule_def.get('name', 'no name'),
+                            rule_index=rule_index,
+                            min_retention=min_retention,
+                            max_retention=max_retention)
```
dufton: Create `rule` before the for loops, since it will be the same for each resource

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Miss dash (-) near resource_ids in rule {}'.format(
+                        rule_index))
+            for rid in resource_ids:
+                if rid == '*':
+                    raise audit_errors.InvalidRulesSchemaError(
+                        'The symbol * is not allowed in rule {}'.format(
+                            rule_index))
+
+                rule = Rule(rule_name=rule_def.get('name', 'no name'),
```
dufton: Rule name should be a required field

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+                            max_retention=max_retention)
+                gcp_resource = resource_util.create_resource(
+                    resource_id=rid,
+                    resource_type=resource_type)
+
+                self.resource_rules_map[_APPLY_TO_BUCKETS][gcp_resource].add(
+                    rule)
+
+    def get_resource_rules(self, applies_to):
+        """Get all the resource rules for (resource, RuleAppliesTo.*).
```
dufton: Update comment? (doesn't use resource, only applies_to)

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+        """
+        return self.resource_rules_map[applies_to]
+
+
+class Rule(object):
+    """Rule properties from the rule definition file.
+    Also finds violations.
+    """
+
+    rttRuleViolation = namedtuple(
```
dufton: just `RuleViolation`

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,318 @@
...
+the maximum retention %d' % (age, maxretention))
+                continue
+            if 'createdBefore' in lci.get('condition', {}):
+                continue
+            if 'matchesStorageClass' in lci.get('condition', {}):
+                continue
+            if 'numNewerVersions' in lci.get('condition', {}):
+                continue
+            if 'isLive' in lci.get('condition', {}):
+                continue
```
dufton: Move this to a helper function that works out the actual retention age (could be None if there is no delete rule). Then in this function can check that min_retention <= actual_retention <= max_retention

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,327 @@
...
+
+    def build_rule_book(self, global_configs=None):
+        """Build RetentionRuleBook from the rules definition file.
+
+        Args:
+            global_configs (dict): Global configurations.
+        """
+        self.rule_book = RetentionRuleBook(self._load_rule_definitions())
+
+    def find_buckets_violations(self, buckets_lifecycle, force_rebuild=False):
```
umairidris: Does bucket need to have its own specific function? I think we could write a single function that works for all resources?

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,327 @@
...
+            rule_index (int): The index of the rule from the rule definitions.
+                Assigned automatically when the rule book is built.
+        """
+        self._rules_sema.acquire()
+        try:
+            applies_to = rule_def.get('applies_to', None)
+            if applies_to is None:
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Lack of applies_to in rule {}'.format(rule_index))
+            # assert(isinstance(applies_to, list))
```
umairidris: remove?

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,327 @@
...
+            resource_type = res.get('type', None)
+            if resource_type is None:
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Lack of type in rule {}'.format(rule_index))
+
+            resource_ids = res.get('resource_ids', None)
+            if resource_ids is None:
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Lack of resource_ids in rule {}'.format(rule_index))
+            # assert(isinstance(resource_ids, list))
```
umairidris: remove?

tests/scanner/audit/data/bucket_retention_test_rules_0.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - name: exact retention 365
+    applies_to: 
+      - bucket
+    resource:
+      - type: organization
+        resource_ids:
+          - "433655558668"
+          - "433655558669"
+    minimum_retention: 365
+    maximum_retention: 365
```
umairidris: nit: add empty line at bottom of each file.

You can use the following vs code settings:

   "files.insertFinalNewline": true,
    "files.trimTrailingWhitespace": true,
    "files.trimFinalNewlines": true,

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,327 @@
...
+
+        Args:
+            rule_def (dict): A dictionary containing rule definition
+                properties.
+            rule_index (int): The index of the rule from the rule definitions.
+                Assigned automatically when the rule book is built.
+        """
+        self._rules_sema.acquire()
+        try:
+            applies_to = rule_def.get('applies_to', None)
```
umairidris: No need for None as that's the default.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,327 @@
...
+    def __init__(self, rule_defs=None):
+        """Initialization.
+
+        Args:
+            rule_defs (dict): rule definitons
+        """
+        super(RetentionRuleBook, self).__init__()
+        self._rules_sema = threading.BoundedSemaphore(value=1)
+
+        self.resource_rules_map = {
```
umairidris: so this is a dict of dicts? do we need that? would just a dict of resource type to list of rules work?

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,327 @@
...
+    """The RuleBook for Retention resources."""
+
+    def __init__(self, rule_defs=None):
+        """Initialization.
+
+        Args:
+            rule_defs (dict): rule definitons
+        """
+        super(RetentionRuleBook, self).__init__()
+        self._rules_sema = threading.BoundedSemaphore(value=1)
```
umairidris: I don't think you need this since forseti runs in a single threaded manner. If you do want to keep it to be safe, use a lock instead of a semaphore.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,327 @@
...
+            resource_type=buckets_lifecycle.type,
+            full_name=buckets_lifecycle.full_name,
+            rule_name=self.rule_name,
+            rule_index=self.rule_index,
+            violation_type=VIOLATION_TYPE,
+            violation_describe=describe
+        )
+
+    # TODO: The naming is confusing and needs to be fixed in all scanners.
+    def find_buckets_violations(self, buckets_lifecycle):
```
umairidris: Again, perhaps we can use just a single function to work for all resources. Else there will be a lot of unnecessary duplication.

google/cloud/forseti/scanner/scanners/retention_scanner.py
```diff
@@ -0,0 +1,141 @@
...
+        LOGGER.info('Finding retention violations...')
+
+        for single_info in bucket_lifecycle_info:
+            violations = self.rules_engine.find_buckets_violations(
+                single_info)
+            all_violations.extend(violations)
+
+        return all_violations
+
+    def _retrieve_bucket(self):
```
umairidris: You may be able to use some ideas from https://github.com/GoogleCloudPlatform/forseti-security/pull/2075 to not have to handle specific resources individually

tests/scanner/audit/data/bucket_retention_test_rules_21.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - name: retention of all buckets is 365 days
+    applies_to:
+      - bucket
+    resource:
+      - type: bucket
+        resource_ids:
+          - some-resource-id1
+    minimum_retention: 365
+    maximum_retention: 365
+  - name: retention of buckets in all projects must be 365 days
```
umairidris: the max_retention is 2147483647 not 365, am i missing something?

tests/scanner/scanners/data/fake_bucket_retention_scanner_rule.yaml
```diff
@@ -0,0 +1,74 @@
...
+      - bucket
+    resource:
+      - type: bucket
+        resource_ids:
+          - "p2bkt5"
+    minimum_retention: 100
+
+
+
+# o - f - p2 - [bkt3,bkt4,bkt5]
```
umairidris: rm?

tests/scanner/audit/data/bucket_retention_test_rules_1.yaml
```diff
@@ -0,0 +1,8 @@
+rules:
```
umairidris: nit: i am not a fan of test cases spread amongst files. consider using the test style from https://github.com/GoogleCloudPlatform/forseti-security/pull/2075 (once approved).

google/cloud/forseti/scanner/scanners/retention_scanner.py
```diff
@@ -116,9 +119,18 @@ def _retrieve_bucket(self):
...
-                lifecycle_info = retention_bucket.RetentionBucket.from_json(
-                    bucketinfo)
-                all_lifecycle_info.append(lifecycle_info)
+                bucketdatadict = json.loads(bucketinfo.data)
+                lifecycleitems = []
+                if ('lifecycle' in bucketdatadict and
+                        'rule' in bucketdatadict.get('lifecycle')):
+                    lifecycleitems = bucketdatadict['lifecycle']['rule']
+
+                new_bucket_res = resource_util.create_resource(
```
dufton: You can just create the Bucket resource directly without using resource_utils, then you can pass the lifecycle in the constructor.

Alternatively, add a from_json method (similar to other Resources), then you can just use something like:

`return bucket.Bucket.from_json(parent=parent_proj, bucket_info.data)`

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,327 @@
...
+from google.cloud.forseti.common.util import relationship
+from google.cloud.forseti.scanner.audit import base_rules_engine as bre
+from google.cloud.forseti.scanner.audit import errors as audit_errors
+from google.cloud.forseti.services.inventory.base import resources
+
+
+LOGGER = logger.get_logger(__name__)
+
+VIOLATION_TYPE = 'RETENTION_VIOLATION'
+# Applyto.
```
dufton: Remove (or expand) comment

google/cloud/forseti/scanner/scanners/retention_scanner.py
```diff
@@ -0,0 +1,130 @@
...
+        Args:
+            retention_info (list): Contains all supported resource in retention
+
+        Returns:
+            list: All violations.
+        """
+        all_violations = []
+
+        for single_info in retention_info:
+            violations = self.rules_engine.find_violations(
```
dufton: Does this fit on a one line? Maybe rename `single_info` to something like `resource`

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,406 @@
...
+from google.cloud.forseti.scanner.audit import base_rules_engine as bre
+from google.cloud.forseti.scanner.audit import errors as audit_errors
+
+
+LOGGER = logger.get_logger(__name__)
+
+SUPPORTED_RETENTION_RESOURCE_TYPES = frozenset(['bucket'])
+
+VIOLATION_TYPE = 'RETENTION_VIOLATION'
+_APPLY_TO_BUCKETS = 'bucket'
```
dufton: You can get rid of `_APPLY_TO_BUCKETS` and `_APPLY_TO_RESOURCES` and just use `SUPPORTED_RETENTION_RESOURCE_TYPES` instead.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,406 @@
...
+        if self.rule_book is None or force_rebuild:
+            self.build_rule_book()
+
+        violations = itertools.chain()
+        resource_rules = self.rule_book.get_resource_rules(resource.type)
+        resource_ancestors = (relationship.find_ancestors(
+            resource, resource.full_name))
+
+        for related_resources in resource_ancestors:
+            rules = resource_rules.get(related_resources)
```
dufton: `.get(related_resources, [])`

Then you don't need the `if not rules:` block

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,406 @@
...
+                Assigned automatically when the rule book is built.
+        """
+        self._rules_sema.acquire()
+        try:
+            applies_to = rule_def.get('applies_to')
+            if applies_to is None:
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Lack of applies_to in rule {}'.format(rule_index))
+
+            retention_range = get_retention_range(rule_def, rule_index)
```
dufton: `min_retention, max_retention = get_retention_range(...`

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,406 @@
...
+        return iter(())
+
+    def bucket_max_retention_violation(self, bucket):
+        """Get a generator for violations especially for maximum retention
+        Args:
+            bucket (bucket): Find violation from the bucket
+        Yields:
+            RuleViolation: All max violations of the bucket breaking the rule.
+        """
+        maxretention = self.max_retention
```
dufton: You can just use `self.max_retention` without creating the extra variable

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,406 @@
...
+            return
+
+        # There should be a condition which guarantees to delete data
+        exist_max_limit = False
+        bucket_lifecycle = bucket.get_lifecycle_rule()
+        if bucket_lifecycle is None:
+            yield self.generate_bucket_violation(bucket)
+        else:
+            for lc_item in bucket_lifecycle:
+                action = lc_item['action']
```
dufton: Try to re-work the conditionals so you don't need continues/breaks if possible:

```
  # Only look at delete actions
  if lc_item.get('action', {}).get('type') == 'Delete':
    conditions = ...
```

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,406 @@
...
+                if action is not None and 'type' in action:
+                    if action['type'] != 'Delete':
+                        continue
+                else:
+                    continue
+
+                conditions = lc_item['condition']
+                if conditions is not None:
+                    age = conditions.get('age')
+                    if age is not None and len(conditions) == 1:
```
dufton: Add a comment that this means the rule cannot have conditions other than age

tests/scanner/audit/data/bucket_retention_test_rules_1.yaml
gruihuang: From test_rules_0 to test_rules_13, each of them rises an error. That is the reason why I use different files. test_rules_2* covers all the other logic.

tests/scanner/audit/data/bucket_retention_test_rules_21.yaml
gruihuang: Yes. the maximum value of the age is the max value of Integer but this is not that meaning. This is a rule instead of a config.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,388 @@
...
+        Yields:
+            RuleViolation: All min violations of the bucket breaking the rule.
+        """
+        if self.min_retention is None:
+            return
+
+        bucket_lifecycle = bucket.get_lifecycle_rule()
+        if bucket_lifecycle is not None:
+            for lc_item in bucket_lifecycle:
+                if lc_item.get('action', {}).get('type') != 'Delete':
```
dufton: If you move the check on `age` into `bucket_conditions_guarantee_min`, then the body of this loop becomes:
```
if (lc_item.get('action', {}).get('type') == 'Delete' and
    not bucket_conditions_guarantee_min(lc_item.get('condition'), self.min_retention):
  yield self.generate_bucket_violation(bucket)
```

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,626 @@
...
+        resource_name=bucket_id,
+        resource_type=res_map.get(bucket_id).type,
+        full_name=res_map.get(bucket_id).full_name,
+        rule_name=rule_name,
+        rule_index=rule_index,
+        violation_type=rre.VIOLATION_TYPE,
+        violation_data=lifecycle_str,
+        resource_data=res_map.get(bucket_id).data)
+
+def GetLefecycleDict(action, age, created_before, matches_storage_class,
```
dufton: Lefe -> Life

tests/scanner/test_data/fake_retention_scanner_data.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2018 The Forseti Security Authors. All rights reserved.
```
dufton: I'd suggest copying the style of the other files in tests/scanner/test_data, where (as much as possible) the test resources are just defined as dictionaries or lists of dictionaries, setting all dictionary keys in the one statement, rather than one key at a time.

Maybe have some common Lifecycle dictionaries as constants that can be shared in different tests

tests/scanner/scanners/retention_scanner_test.py
```diff
@@ -0,0 +1,251 @@
...
+
+class RetentionScannerTest(ForsetiTestCase):
+
+    def setUp(self):
+        """Set up."""
+
+
+    #"lifecycle": {"rule": [{"action": {"type": "Delete"}, "condition": {"age": 20, "numNewerVersions": 5}}, {"action": {"type": "Delete"}, "condition": {"age": 10, "numNewerVersions": 99}}]},
+    @unittest.skipIf(do_not_test_old_cases, 'debug new test cases')
+    def test_scanner_01(self):
```
dufton: You can follow the pattern of other `*_scanner_test.py` files, where the scanner_test mocks out the rule engine and focuses more on the retrieve, find_violations and output_results methods. The rule_engine test focuses more on the logic of finding violations in resources.

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,626 @@
...
+# TODO: More tests need to be added that cover the rule attributes and how they
+#    are evaluated
+class RetentionRulesEngineTest(ForsetiTestCase):
+    """Tests for the BigqueryRulesEngine."""
+
+    def setUp(self):
+        """Set up."""
+
+    @unittest.skipIf(do_not_test_old_cases, 'debug new test cases')
+    def test_retention_retrieve_1(self):
```
dufton: It is useful when looking at test outputs to have more descriptive names, e.g.
test_invalid_rule_with_no_applies_to

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,626 @@
...
+
+        generate_res_for_22()
+        all_lifecycle_info = self.scanner._retrieve()
+        all_violations = self.scanner._find_violations(all_lifecycle_info)
+
+        res_map = {}
+        for i in _fake_bucket_list_22:
+            res_map[i.id] = i
+
+        expected_violations = set([
```
dufton: Break this up into multiple tests so that each test only needs a small number of test resources, and produces a small number (or no) violations.

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,626 @@
...
+from google.cloud.forseti.scanner.audit import errors as audit_errors
+from tests.scanner.test_data import fake_retention_scanner_data as frsd
+from datetime import datetime, timedelta
+
+
+import collections
+from tests.unittest_utils import ForsetiTestCase
+from google.cloud.forseti.scanner.scanners import retention_scanner
+
+do_not_test_old_cases = False
```
dufton: Remove this (and in scanner_test)?

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,626 @@
...
+        result['condition']['createdBefore'] = created_before
+    if matches_storage_class != None:
+        result['condition']['matchesStorageClass'] = matches_storage_class
+    if num_newer_versions != None:
+        result['condition']['numNewerVersions'] = num_newer_versions
+    if is_live != None:
+        result['condition']['isLive'] = is_live
+    return result
+
+_fake_bucket_list_22 = []
```
dufton: You shouldn't need these to be globals

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,626 @@
...
+    )
+
+    res = []
+
+    # MAX
+    normal = 365
+    larger = 366
+
+    data_creater = frsd.FakeBucketDataCreater('bkt_max_1', proj_max)
+    data_creater.AddLefecycleDict("Delete", normal, None, None, None, None)
```
dufton: For a method like this, give all the non-required arguments a default value of None, then use named arguments:
```
  AddLifecycleDict(action='Delete', age=499)
  AddLifecycleDict(action='Delete', age=500, num_newer_versions=5)
  ...
```

google/cloud/forseti/scanner/scanners/retention_scanner.py
```diff
@@ -0,0 +1,130 @@
...
+        Args:
+            violations (list): The RuleViolations to flatten.
+
+        Yields:
+            dict: Iterator of RuleViolations as a dict per member.
+        """
+        for violation in violations:
+            yield {
+                'resource_name': violation.resource_name,
+                'resource_id': violation.resource_name,
```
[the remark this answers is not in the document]
gruihuang: Since the resource name is exactly the same as the resource id for buckets, the violations do not have resource id member.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
gruihuang: I believe we need not only the type but also the id to identify the rules so the book should be dict[type][id] something like that.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,387 @@
...
+            res (Resource): A class derived from Resource
+        Returns:
+            Generator: All violations of the resource breaking the rule.
+        """
+
+        if res.type == 'bucket':
+            return self.find_violations_in_bucket(res)
+        return iter(())
+
+    def bucket_max_retention_violation(self, bucket):
```
umairidris: Make these functions generic. For now you can put a comment that it only will support bucket, and then work on generalizing in future PRs.

google/cloud/forseti/scanner/scanners/retention_scanner.py
```diff
@@ -0,0 +1,132 @@
...
+    def _retrieve(self):
+        """Retrieves the data for scanner.
+
+        Returns:
+            list: a list of Resources, with a type in
+                SUPPORTED_RETENTION_RES_TYPES
+        """
+        model_manager = self.service_config.model_manager
+        scoped_session, data_access = model_manager.get(self.model_name)
+        retention_info = []
```
umairidris: this should probably be called resources

tests/scanner/scanners/retention_scanner_test.py
```diff
@@ -0,0 +1,383 @@
...
+        resource_name=u'buckets/'+bucket_id,
+        resource_type=res_map.get(bucket_id).type,
+        full_name=res_map.get(bucket_id).full_name,
+        rule_name=rule_name,
+        rule_index=rule_index,
+        violation_type=rre.VIOLATION_TYPE,
+        violation_data=lifecycle_str,
+        resource_data=res_map.get(bucket_id).data)
+
+yaml_str_multi_buckets_in_a_rule = """
```
umairidris: a lot of these rules could be created via a helper function or formatted string to reduce duplication

google/cloud/forseti/common/gcp_type/bucket.py
```diff
@@ -90,3 +90,14 @@ def from_json(cls, parent, json_string):
             locations=[bucket_dict['location']],
             data=json_string,
         )
+
+    def get_lifecycle_rule(self):
+        """Create a bucket lifecycle's rules dict from its JSON string.
+
+        Returns:
+            dict: bucket lifecycle's rules.
```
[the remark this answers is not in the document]
gruihuang: I return an empty dictionary now

google/cloud/forseti/common/gcp_type/bucket.py
```diff
@@ -90,3 +90,14 @@ def from_json(cls, parent, json_string):
...
+        """
+        bucket_dict = json.loads(self.data)
+        if 'lifecycle' in bucket_dict and 'rule' in bucket_dict['lifecycle']:
+            return bucket_dict['lifecycle']['rule']
+        return None
```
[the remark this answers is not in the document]
gruihuang: Yes. I change it to an empty dictionary

google/cloud/forseti/common/gcp_type/resource_util.py
```diff
@@ -81,7 +81,6 @@ def create_resource(resource_id, resource_type, **kwargs):
 
 def create_resource_from_json(resource_type, parent, json_string):
     """Factory to create a certain kind of Resource from JSON data.
-
```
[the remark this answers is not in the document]
gruihuang: Done

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,387 @@
...
+    Args:
+        rule_def (dict): A dictionary containing rule definition
+            properties.
+        rule_index (int): The index of the rule from the rule definitions.
+            Assigned automatically when the rule book is built.
+
+    Returns:
+        pair: the minimum and maximum value of the Age.
+    """
+    minimum_retention = rule_def.get('minimum_retention', None)
```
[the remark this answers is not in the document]
gruihuang: Done

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,387 @@
...
+        try:
+            applies_to = rule_def.get('applies_to')
+            if applies_to is None:
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Lack of applies_to in rule {}'.format(rule_index))
+
+            minimum_retention, maximum_retention = get_retention_range(
+                rule_def, rule_index)
+
+            if any(x not in SUPPORTED_RETENTION_RES_TYPES for x in applies_to):
```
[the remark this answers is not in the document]
gruihuang: Done. But I prefer to keep resource_rules_map because many other scanners have this name.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
gruihuang: Done

google/cloud/forseti/scanner/scanners/retention_scanner.py
gruihuang: I change it to retention_res. Because I think the return value should be something relevant to the scanner's name based on other scanners.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
gruihuang: Since many other scanners have the same name with the same purpose, I want to keep the consistency.

tests/scanner/scanners/retention_scanner_test.py
gruihuang: I tried to use a template string but they have more different than common. If I use templates, it could be a little complex. In this way, it might be easier to understand.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
gruihuang: It makes sense but we would better keep that for the consistency

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,392 @@
...
+            bucket (bucket): Find violation from the bucket
+        Yields:
+            RuleViolation: All max violations of the bucket breaking the rule.
+        """
+        if self.max_retention is None:
+            return
+
+        # There should be a condition which guarantees to delete data
+        bucket_lifecycle = bucket.get_lifecycle_rule()
+        if not bucket_lifecycle:
```
dufton: If you invert this condition to just `if bucket_lifecycle` with the `for` loop inside, then you can have the yield of violation only place, at the end of the function

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,392 @@
...
+            Generator: All violations of the buckets breaking rules.
+        """
+
+        violation_max = self.bucket_max_retention_violation(bucket)
+        violation_min = self.bucket_min_retention_violation(bucket)
+        return itertools.chain(violation_max, violation_min)
+
+
+def bucket_conditions_guarantee_min(conditions, min_retention):
+    """check if other conditions can guarantee minimum retention
```
dufton: Style nit: Capital letter at start of comment, and add a blank line between the first line (summary) and rest of the docstring.

(here and elsewhere)

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,392 @@
...
+        created_before = conditions['createdBefore']
+        dt_cfg = dt.get_datetime_from_string(created_before,
+                                             '%Y-%m-%d')
+        dt_now = dt.get_utc_now_datetime()
+        day_diff = (dt_now - dt_cfg).days
+        if day_diff >= min_retention:
+            return True
+
+    # if number of new version is larger than 0, OK.
+    if 'numNewerVersions' in conditions:
```
dufton:
```
if conditions.get('numNewerVersions', 0) >= 1:
  return True
```

rules/retention_rules.yaml
```diff
@@ -0,0 +1,44 @@
...
+#   - name: Buckets in Organization must have a retention between 180 and 730 days.
+#     applies_to:
+#       - bucket
+#     resource:
+#       - type: organization
+#         resource_ids:
+#           - "123456789012"
+#     minimum_retention: 180
+#     maximum_retention: 730
+#   - name: Buckets in Project must have a retention between 100 and 365 days.
```
dufton: It might be good to have for the 3 examples in this file:
1 with a min and a max
1 with min only
1 with max only

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,767 @@
...
+#    are evaluated
+class RetentionRulesEngineTest(ForsetiTestCase):
+    """Tests for the BigqueryRulesEngine."""
+
+    def setUp(self):
+        """Set up."""
+
+    def test_invalid_rule_with_no_applies_to(self):
+        yaml_str_no_applies_to="""
+rules:
```
dufton: YAML should handle the extra indentation fine if you add 12 spaces to keep everything looking more consistent.

(here and other constants)

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,767 @@
...
+        self.assertEqual(1, len(rules_engine.rule_book.resource_rules_map))
+
+        data_creater = frsd.FakeBucketDataCreater('fake_bucket', frsd.PROJECT1)
+        data_creater.AddLefecycleDict(action="Delete", age=89, num_newer_versions=1)
+
+        fake_bucket = data_creater.get_resource()
+        got_violations = list(rules_engine.find_violations(fake_bucket))
+        self.assertEqual(got_violations, [])
+
+    def test_only_min_lessver0_del(self):
```
dufton: Please add a comment/docstring to each of these tests to explain what they are testing.

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,767 @@
...
+        resource_type=res_map.get(bucket_id).type,
+        full_name=res_map.get(bucket_id).full_name,
+        rule_name=rule_name,
+        rule_index=rule_index,
+        violation_type=rre.VIOLATION_TYPE,
+        violation_data=lifecycle_str,
+        resource_data=res_map.get(bucket_id).data)
+
+
+# TODO: More tests need to be added that cover the rule attributes and how they
```
dufton: Are more tests still needed? What is missing?

tests/scanner/scanners/retention_scanner_test.py
dufton: These YAML strings seem to each only be used in once place. Can you move them inside the function that uses them.

tests/scanner/scanners/retention_scanner_test.py
```diff
@@ -0,0 +1,383 @@
...
+        resource_ids:
+          - fake-bucket-1
+          - fake-bucket-2
+          - fake-bucket-3
+    minimum_retention: 90
+
+"""
+
+
+def _mock_bucket_retention_multi_buckets_in_a_rule(_, resource_type):
```
dufton: Add a docstring/comment describing what these functions are doing/making.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,392 @@
...
+                    resource_id=rid,
+                    resource_type=resource_type)
+
+                self.resource_rules_map[apply_to][gcp_resource].add(
+                    rule)
+
+    def get_resource_rules(self, applies_to):
+        """Get all the rules for the resource "applies_to".
+        Args:
+            applies_to (str): The name of the resource
```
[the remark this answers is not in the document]
gruihuang: It is a string whose value is the type of the resource applied to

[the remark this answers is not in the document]
dufton: Update the comment (implies resource name rather than type)

tests/scanner/audit/retention_rules_engine_test.py
gruihuang: It would come out with errors if I add additional indentations. Should I write a function to delete extra indentations?

tests/scanner/scanners/retention_scanner_test.py
gruihuang: These yaml string are used inside the test cases but these strings come along with the _mock_bucket_retention... functions below them. I put them together to make it a bit more clear.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+                        'redundant applies_to in rule {}'.format(rule_index))
+                added_applies_to.add(appto)
+
+                self.create_and_add_rule(
+                    rule_def,
+                    rule_index,
+                    appto,
+                    minimum_retention,
+                    maximum_retention)
+        finally:
```
umairidris: remove sem

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+        Args:
+            rule_def (dict): A dictionary containing rule definition
+                properties.
+            rule_index (int): The index of the rule from the rule definitions.
+                Assigned automatically when the rule book is built.
+            apply_to (str): The resource type that the rule is applied to
+            min_retention(int): minimum value of the age in lifecycle
+            max_retention(int): maximum value of the age in lifecycle
+        """
+        resource = rule_def.get('resource', None)
```
umairidris: default for .get is None, no need to put ", None"

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+        if resource is None:
+            raise audit_errors.InvalidRulesSchemaError(
+                'Lack of resource in rule {}'.format(rule_index))
+
+        for res in resource:
```
umairidris: change resource to resources so we know it's a list

google/cloud/forseti/scanner/audit/retention_rules_engine.py
umairidris: consider making the default [] and moving the "if resource is None" check above and making it "if 'resource' not in rule_def"

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+            resource_type = res.get('type', None)
```
umairidris: no need for None here.

Consider:

resource_type = res.get('type')
if not resource_type:

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+            if resource_type is None:
+                raise audit_errors.InvalidRulesSchemaError(
+                    'Lack of type in rule {}'.format(rule_index))
+
+            resource_ids = res.get('resource_ids', None)
```
umairidris: ditto

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+class Rule(object):
+    """Rule properties from the rule definition file.
+    Also finds violations.
+    """
+
+    def __init__(self, rule_name, rule_index, min_retention, max_retention):
+        """Initialize.
+
+        Args:
+            rule_name (str): Name of the loaded rule
```
umairidris: end docstring sentences with periods

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+
+        Returns:
+           defaultdict:  A list of ResourceRules.
+        """
+        return self.resource_rules_map[applies_to]
+
+
+class Rule(object):
+    """Rule properties from the rule definition file.
+    Also finds violations.
```
umairidris: Add an empty new line between these sentences

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+            min_retention(int): minimum value of the age in lifecycle
+            max_retention(int): maximum value of the age in lifecycle
+        """
+        self.rule_name = rule_name
+        self.rule_index = rule_index
+        self.min_retention = min_retention
+        self.max_retention = max_retention
+
+    def generate_bucket_violation(self, bucket):
+        """generate a violation.
```
umairidris: Capitalize docstring sentences

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+
+        Args:
+            res (Resource): A class derived from Resource
+        Returns:
+            Generator: All violations of the resource breaking the rule.
+        """
+
+        if res.type == 'bucket':
+            return self.find_violations_in_bucket(res)
+        return iter(())
```
umairidris: maybe raise ValueError('only bucket is currently supported')

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,391 @@
...
+            Generator: All violations of the buckets breaking rules.
+        """
+
+        violation_max = self.bucket_max_retention_violation(bucket)
+        violation_min = self.bucket_min_retention_violation(bucket)
+        return itertools.chain(violation_max, violation_min)
+
+
+def bucket_conditions_guarantee_min(conditions, min_retention):
+    """Check if other conditions can guarantee minimum retention
```
umairidris: period

google/cloud/forseti/scanner/scanners/retention_scanner.py
```diff
@@ -0,0 +1,132 @@
...
+    def _output_results(self, all_violations):
+        """Output results.
+
+        Args:
+            all_violations (list): All violations
+        """
+        all_violations = self._flatten_violations(all_violations)
+        self._output_results_to_db(all_violations)
+
+    def _find_violations(self, retention_info):
```
umairidris: rename retention_info to resources

tests/scanner/scanners/retention_scanner_test.py
umairidris: Why not put them in the mock functions then?

tests/scanner/scanners/retention_scanner_test.py
```diff
@@ -0,0 +1,396 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for RetentionScanner."""
+
+import collections
+import json
+import unittest
+import mock
+from collections import namedtuple
```
umairidris: collections is already imported, so just do collections.namedtuple

tests/scanner/scanners/retention_scanner_test.py
```diff
@@ -0,0 +1,396 @@
...
+        resource_ids:
+          - fake-bucket-1
+          - fake-bucket-2
+          - fake-bucket-3
+    minimum_retention: 90
+
+"""
+
+
+def _mock_bucket_retention_multi_buckets_in_a_rule(_, resource_type):
```
umairidris: There is still A LOT of duplication between tests. Use closure to simplify:

Consider:

FakeBucketData = collections.namedtuple('FakeBucketData', ['bucket_id', 'project', 'age'])

```
def _get_mock_iter(fake_bucket_data):
    
    def _mock_iter(_, resource_type):
        if resource_type != 'bucket':
            raise ...
        resources = []
        for data in fake_bucket_data:
            data_creator = frsd.FakeBucketDataCreator(data.bucket_id, data.project)
          if data.age:
            data_creator.AddLifecycleDict(action='Delete', age=data.age)
          resources.append(data_creator.get_resource()) 
        return resources

    return _mock_iter

```

tests/scanner/scanners/retention_scanner_test.py
gruihuang: Oh, maybe I used the wrong word. Both Yaml and  _mock_bucket_retention are used by a test case, while they do not use each other. The test case itself does not carry any useful information so I put yaml and _mock_bucket_retention... together, which makes it easier to design the test data.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
gruihuang: I believe the _rules_sema is not necessary but "audit_logging", "enabled_apis", "iam", "iap", and "log_sink" all use it in this way.

tests/scanner/scanners/retention_scanner_test.py
gruihuang: Done

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,856 @@
...
+        if its age condition is larger than maximum_retention"""
+        rules_engine = get_rules_engine_with_rule(RetentionRulesEngineTest.yaml_str_only_max_retention)
+        self.assertEqual(1, len(rules_engine.rule_book.resource_rules_map))
+
+        data_creater = frsd.FakeBucketDataCreater('fake_bucket', frsd.PROJECT1)
+        data_creater.AddLifecycleDict(action="Delete", age=366)
+
+        fake_bucket = data_creater.get_resource()
+        got_violations = list(rules_engine.find_violations(fake_bucket))
+        expect_violations = frsd.build_bucket_violations(
```
umairidris: nit: expected_violations is more grammatically correct

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -0,0 +1,856 @@
...
+        self.assertEqual(1, len(rules_engine.rule_book.resource_rules_map))
+
+        data_creater = frsd.FakeBucketDataCreater('fake_bucket', frsd.PROJECT1)
+        data_creater.AddLifecycleDict(action="Delete", age=100)
+
+        fake_bucket = data_creater.get_resource()
+        got_violations = list(rules_engine.find_violations(fake_bucket))
+        self.assertEqual(got_violations, [])
+
+    def test_both_min_max_3_conditions(self):
```
umairidris: what do the numbers in the test names mean? can we use descriptive names instead of numbers?

tests/scanner/scanners/retention_scanner_test.py
```diff
@@ -0,0 +1,401 @@
...
+
+            all_lifecycle_info = self.scanner._retrieve()
+            all_violations = self.scanner._find_violations(all_lifecycle_info)
+
+            res_map = {}
+            for i in _fake_bucket_list:
+                res_map[i.id] = i
+
+            expected_violations = set([
+                get_expect_violation_item(res_map, 'fake-bucket-1',
```
umairidris: use 'expected' here too

tests/scanner/audit/retention_rules_engine_test.py
gruihuang: It means there are 3 conditions.

tests/scanner/scanners/retention_scanner_test.py
gruihuang: Do you mean rename "expected_violations" to "expected"?

gruihuang: OK, get it.

gruihuang:
To correct the user email

PR #2099, opened 2018-10-16
Added alter column functionality and increased the size of asset_data column

joecheuk:


install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -63,62 +84,92 @@ def drop_column(table, column):
...
+def _alter_columns(column_action, columns, table):
+    """Alter columns.
+
+    Args:
+        column_action (str): Column Action.
+        columns (dict): A dictionary of old_column: new_column.
+        table (sqlalchemy.schema.Table): The sql alchemy table object.
+    """
+    column_action = column_action.upper()
+    for old_column, new_column in columns.iteritems():
```
blueandgold: Is there any check to make sure the alter command doesn't run all the time?  If there are data in the tables, the alter command might take a while.

install/gcp/upgrade_tools/db_migrator.py
```diff
@@ -63,62 +84,92 @@ def drop_column(table, column):
...
-    for subclass in base_subclasses:
-        get_schema_update_actions = getattr(subclass,
+    for dao_class in dao_classes:
+        get_schema_update_actions = getattr(dao_class,
                                             schema_update_actions_method,
                                             None)
         if (not callable(get_schema_update_actions) or
-                subclass.__tablename__ not in tables):
+                dao_class.__tablename__ not in tables):
             LOGGER.warn('Method: %s is not callable or Table: %s doesn\t '
```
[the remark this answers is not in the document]
joecheuk: Updated.

install/gcp/upgrade_tools/db_migrator.py
joecheuk: The same ALTER command will be executed but mysql will detect if there is any changes before actually applying it

Issue #2135, opened 2018-10-24
Fix Flake8 styling issues in dev causing Travis build to fail


hshin-g:
Travis build is failing due to the many styling issues caught by Flake8. 

Will go through and fix each of the issues to remedy the problem.

@joecheuk @blueandgold

hshin-g:
Majority of issues are `W504` Line break after binary operator.
Once you change so that the line break is before the binary operator, get `W503` Line break before binary operator.

Need to change the config file or settings so that it is aligned on only one.

kevensen:
FWIW, I'm seeing this too.

hshin-g:
PR created and merged into dev to address issue.

PR #2137, opened 2018-10-24
Fix Flake8 styling issues in dev causing Travis build to fail

hshin-g:
Github Issue #2135 

+ Put break before binary operator and instead of before for Flake8 - first test


google/cloud/forseti/common/util/regular_exp.py
```diff
@@ -32,7 +32,7 @@ def escape_and_globify(pattern_string):
         pattern_string (str): The pattern string of which to make a regex.
 
     Returns:
-        str: The pattern string, escaped except for the "*", which is
+        str: The pattern string, escaped except for the "\*", which is
```
blueandgold: Why is this changed to `\*`?  The behavior of the code on line 42, shows that `*` is what's handled?

hshin-g: Got a warning on that line and wanted to test to see if that was what it was flagging- changed it back.

hshin-g:
Fixed out majority of Flake8 warnings/errors by adding `W504` in the ignore flag.

Last (hopefully) issue is dealing with `W605 invalid escape sequence '\*'`, which revolves around the usage of \* in docstrings to represent hierarchy/path, e.g. `Org 1234, bindings: roles/\*, members: user:\*@company.com` in `google/cloud/forseti/scanner/audit/iam_rules_engine.py`. Adding `r'` in front probably doesn't mesh well with the documentation, and adding this warning to the ignore flag might lead to other code falling through the cracks, so would be great to get your take on what to do here @blueandgold @joecheuk

hshin-g:
Latest commit was to test if we can take out the backslash `\\*` in the docstrings and it worked - Travis build passed. Please look over @blueandgold @joecheuk.

blueandgold:
Is there any use for the escaped asterisk in the method docs?  I am wondering if it's so that the `*` shows up properly in the website api references?

For example:
https://forsetisecurity.org/docs/latest/develop/reference/google.cloud.forseti.scanner.audit.iam_rules_engine.html#module-google.cloud.forseti.scanner.audit.iam_rules_engine
image

Since this only happens in only a couple of places, we might be better off to just add lint ignore in those couple of places.  How do you think?

Issue #2128, opened 2018-10-22
Location scanner cannot express specific locations for resources 


umairidris:
e.g. I want bucket 'foo' to be in 'us'.

Currently the location scanner applies to all buckets in a project, folder or organization. We want it to apply to specific resource ids as well.

Proposal: change applies_to from just a list of strings to a list of resources.

e.g. the rule above would be expressed as:

```
rules:
  - name: rule.
    mode: whitelist
    resource:
      - type: organization
        resource_ids:
          - org-1
    applies_to:
      - type: bucket
        resource_ids:
           - foo
    locations:
      - 'us'
```

resource_ids can be * as well.

PR #2142, opened 2018-10-25
Support specific ids in location rules applies_to

umairidris:
Code for #2128.

Still need to update docs.

### DIFF
--- a/google/cloud/forseti/scanner/audit/location_rules_engine.py
+++ b/google/cloud/forseti/scanner/audit/location_rules_engine.py
@@ -184,12 +184,18 @@ class LocationRuleBook(base_rules_engine.BaseRuleBook):
                 raise errors.InvalidRulesSchemaError(
                     'Missing field "{}" in rule {}'.format(field, rule_index))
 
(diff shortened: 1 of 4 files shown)


google/cloud/forseti/scanner/audit/location_rules_engine.py
```diff
@@ -184,13 +184,18 @@ def _build_rule(cls, rule_def, rule_index):
...
+
+        for applies_dict in rule_def.get('applies_to'):
+            resource_type = applies_dict['type']
+
             if resource_type not in SUPPORTED_LOCATION_RESOURCE_TYPES:
                 raise errors.InvalidRulesSchemaError(
                     'Unsupported applies to type "{}" in rule {}'.format(
                         resource_type, rule_index))
 
+            applies_to[resource_type] = applies_dict['resource_ids']
```
joecheuk: Can we make this backward compatible? e.g. if 'resource_ids' is not defined then we can use '*' as the default value.

umairidris: Done.

PR #2173, opened 2018-11-01
Update apt_packages to include new mysql name

umairidris:
libmysqlclient-dev is no longer available after Debian Jessie (oldstable). It is now default-libmysqlclient-dev.

https://packages.debian.org/search?searchon=names&keywords=libmysqlclient-dev


umairidris:
Both are available in Ubuntu 18.04 LTS as well (https://packages.ubuntu.com/bionic/default-libmysqlclient-dev)

Issue #2189, opened 2018-11-05
Wildcard behaviour is unconventional


umairidris:
In Forseti rules the wildcard means "1 or more" whereas in Linux and other systems, it means "0 or more" [1]. This inconsistency causes surprising behaviour and can mislead customers. For example, today we are advertising blocking all europe locations by doing 'eu*' in the location rules, but this does not actually block "eu" itself!

Please consider changing the behaviour of * to be 0 or more.

[1]
```
$ echo "" > foo
$ ls
foo
$ ls f*
foo
$ ls foo*
foo
$ ls food*
ls: cannot access 'food*': No such file or directory
```

umairidris:
Weirdly enough just * alone matches 0 or more, but * with a prefix requires there to be at least 1 character after the prefix...

```
if pattern_string == '*':
        return '^.*$'

return '^{}$'.format(re.escape(pattern_string).replace('\\*', '.+?'))
```

I also don't think we need the ? in the regex, it seems to be a no-op

jjawad-google:
We will keep this in mind as we continue to design scanner 2.0.

umairidris:
@jjawad-google do you think it's worth fixing the behaviour for all scanners sooner? The code change is trivial. I wonder how many people are misinterpreting what the wildcard does and misusing it. There is no documentation on this behaviour outside of looking at the code.

I coded a workaround specifically for the location scanner as being able to do us* is very important for the locations (match both locations US and US-WEST1), but others still remain.

Issue #2027, opened 2018-09-21
Another DB Session out of sync: "Commands out of sync; you can't run this command now"


blueandgold:
We have fixed a similar issue before in #1718. We can not have multiple simultaneous queries in in the same session.

2018-09-12 13:21:29,472 ERROR [forseti-security] google.cloud.forseti.scanner.scanner(run): Error running scanner: BigqueryScanner
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.2.0-py2.7.egg/google/cloud/forseti/scanner/scanner.py", line 111, in run
    scanner.run()
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.2.0-py2.7.egg/google/cloud/forseti/scanner/scanners/bigquery_scanner.py", line 179, in run
    bigquery_acl_data = self._retrieve()
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.2.0-py2.7.egg/google/cloud/forseti/scanner/scanners/bigquery_scanner.py", line 144, in _retrieve
    if dataset.parent.type != 'project':
  File "build/bdist.linux-x86_64/egg/sqlalchemy/orm/attributes.py", line 242, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "build/bdist.linux-x86_64/egg/sqlalchemy/orm/attributes.py", line 599, in get
    value = self.callable_(state, passive)
  File "build/bdist.linux-x86_64/egg/sqlalchemy/orm/strategies.py", line 631, in _load_for_state
    session, state, primary_key_identity, passive)
  File "<string>", line 1, in <lambda>
  File "build/bdist.linux-x86_64/egg/sqlalchemy/orm/strategies.py", line 724, in _emit_lazyload
    primary_key_identity)
  File "build/bdist.linux-x86_64/egg/sqlalchemy/ext/baked.py", line 508, in _load_on_pk_identity
    result = list(bq.for_session(self.session).params(**params))
  File "build/bdist.linux-x86_64/egg/sqlalchemy/ext/baked.py", line 353, in __iter__
    return q._execute_and_instances(context)
  File "build/bdist.linux-x86_64/egg/sqlalchemy/orm/query.py", line 3011, in _execute_and_instances
    result = conn.execute(querycontext.statement, self._params)
  File "build/bdist.linux-x86_64/egg/sqlalchemy/engine/base.py", line 948, in execute
[...]

PR #2203, opened 2018-11-07
Add location support for cloud sql instance, gce instance, bigquery datasets and gke clusters

### LINKED ISSUE #2053: Support validation of GCP resource locations
We would like to validate GCP resources are in the regions are they are meant to be in.

Test case: Where (geographically) should specific resources be hosted?

Resources:
- GCS bucket
- GKE cluster
- GCE VM
- BQ Dataset
- Cloud SQL Table

umairidris:
Also:
- create resource classes for missing resource types
- fix instance and ke cluster type to inherit from resource, make their from_json functions uniform with other resources
- add support for applying loc rules to all resources via type wildcard
- add workaround for #2189: add boolean kwarg to treat wildcards as 0 or more. 

Fix #2053.


google/cloud/forseti/scanner/scanners/blacklist_scanner.py
```diff
@@ -99,11 +100,15 @@ def get_instance_networks_interfaces(self):
...
                     session, 'instance'):
-                instance = Instance.from_json(
-                    instance_from_data_model.full_name,
-                    instance_from_data_model.data,
-                    instance_from_data_model.parent.name)
-                network_interfaces.append(instance.create_network_interfaces())
+
+                proj = project.Project(
+                    project_id=instance_from_data_model.parent.name,
+                    full_name=instance_from_data_model.parent.full_name,
```
joecheuk: We ran into a problem when executing queries in a yield_per session and that will cause the query to throw exception.
In this case if count(instance_from_data_model) > the PER_YIELD threshold (1024), the session will get only 1024 items and in the first query (but the session is still being used) and when you do 'instance_from_data_model.parent.name / full_name', a new query is generated and run using the same session which will cause the command out of sync error.

A similar issue:
https://github.com/GoogleCloudPlatform/forseti-security/issues/2027
Do you mind spliting that into 2 queries? This will be the same for other scanners.

google/cloud/forseti/common/util/regular_exp.py
```diff
@@ -41,4 +43,8 @@ def escape_and_globify(pattern_string):
     # pylint: enable=anomalous-backslash-in-string
     if pattern_string == '*':
         return '^.*$'
-    return '^{}$'.format(re.escape(pattern_string).replace('\\*', '.+?'))
+
+    # TODO(2189): Make wildcards always 0 or more.
+    replace = '.*' if wildcard_is_zero_or_more else '.+'
```
joecheuk: Do we need the '?' at the end? e.g. '.*?' and '.+?' (Looks like it was there from before)

umairidris: I think it's a no-op. Do you know if an example where it did anything?

PR #2245, opened 2018-11-13
bucket retention all codes

gruihuang:
New feature "bucket retention". Add GCS Bucket Retention to Forseti bucket scanner.
Close the old pull requests to correct email.


google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -0,0 +1,392 @@
...
+            return
+
+        # There should be a condition which guarantees to delete data
+        bucket_lifecycle = bucket.get_lifecycle_rule()
+        if bucket_lifecycle:
+            for lc_item in bucket_lifecycle:
+                if lc_item.get('action', {}).get('type') == 'Delete':
+                    conditions = lc_item.get('condition', {})
+                    age = conditions.get('age')
+                    if age is not None and len(conditions) == 1:
```
umairidris: this can be joined into one statement:
if age is not None and len(conditions) == 1 and age <= self.max_retention

PR #2273, opened 2018-11-19
Update Cloud Asset integration to skip resource with long names.

ahoying:
An analysis of data in the Cloud Asset backend showed a handful of resources with names longer than 512 characters, which the current data model in Forseti cannot handle. This PR skips resources with extra long names.


ahoying:
The problem is that impacts the size of the index key, which has a direct impact on all inserts and queries to the table. There is only a handful of resources longer than 512 chars and they are not on resources that we directly support in forseti yet, and mostly on testing resources created by google for UI validation (what happens when a really long resource name is created).

This is still a temporary work around while I explore longer term options like hashing the name to have a consistent key size, or using a different technology for the temporary table like memcache.

PR #2274, opened 2018-11-20
Consolidate rule description and specification

zachberger:


zachberger:
Agreed. I'll update the PR.

zachberger:
I'm not sure including all the rules in the left navigation makes sense. 

screen shot 2018-11-28 at 6 51 52 pm

I think there are a few things to consider to improve the appearance: 
* While its nice to include the full product name in the rule name, the product names take up quite a bit of space. Changing things like "Google Kubernetes Engine" to "GKE" will save quite a bit of space. We could probably drop "Cloud" from quite a few places. 
* Renaming the header from "Scanner" to "Scanner Rules" would allow us to take the word "Rule" out of each of the titles
* Make the fonts a bit smaller. GCE API docs use smaller fonts for their entire layout. Example: https://cloud.google.com/compute/docs/reference/rest/v1/firewalls


Even if I make the suggested changes the left nav is going to become crowded quickly.

zachberger:
@blueandgold  I've added a better description on rules.md. The page was also feeling a bit light so I brought in the default rules into this page.

zachberger:
Closing as it is too far out of date

PR #2281, opened 2018-11-21
Fix dependency in forseti.service

rvandegrift:
Wants= is only valid in the Unit section of a systemd unit.


install/gcp/scripts/initialize_forseti_services.sh
```diff
@@ -48,14 +48,14 @@ SQL_PROXY_COMMAND+=" -instances=${SQL_INSTANCE_CONN_STRING}=tcp:${SQL_PORT}"
 API_SERVICE="$(cat << EOF
 [Unit]
 Description=Forseti API Server
+Wants=cloudsqlproxy.service
```
joecheuk: What about using "Requires" instead? Since forseti service cannot start without cloudsql proxy

PR #2289, opened 2018-11-23
(3 in 1) External Project Access Scanner & Notifier Compatibility & Cron Job

angelsungoogle:

- [ ] All of the [unit-tests](https://forsetisecurity.org/docs/latest/develop/dev/testing.html) still pass.


angelsungoogle:
The code I submitted so far adds the `--scanner_index_id` parameter to `forseti notifier run`. I'm reaching a corner case here, and would like your input @blueandgold @joecheuk :
- If only scanner index id is provided: need to find inventory index id based on scanner index id, otherwise the program throws an error
- If both scanner and inventory index id are provided: need to handle a possible user error when the scanner index id provided doesn't match the inventory index id provided

I've pushed my tentative solution, please see the files changed, and feel free to comment.

install/gcp/scripts/run_external_project_access_scanner.sh
```diff
@@ -0,0 +1,85 @@
...
+forseti model use ${MODEL_NAME}
+# Sometimes there's a lag between when the model
+# successfully saves to the database.
+sleep 10s
+echo "Forseti config: $(forseti config show)"
+
+# Run scanner command
+echo "Running Forseti external project access scanner."
+scanner_return = `forseti scanner run --scanner external_project_access_scanner`
+scanner_index_id =  ${scanner_return} | grep '(?<="serverMessage": "Scanner Index ID: )(.*)(?= is created")"'
```
angelsungoogle: this cannot be tested locally, as some previous variables are not initiated. One way I can think of is to install a new forseti with this script. Do you have any recommendations?

angelsungoogle: In addition, I used regex here to get the scanner index id out of the return json. not sure if this is the best way to do it. please feel free to advice

google/cloud/forseti/services/scanner/service.py
```diff
@@ -82,7 +82,7 @@ def Run(self, request, context):
         Yields:
             Progress: The progress of the scanner.
         """
-        scanner_name = request.handle
+        scanner_name = request.scanner_name
```
angelsungoogle: this is to add back a fix we had before. somehow it's gone. so adding it back.

google/cloud/forseti/services/client.py
```diff
@@ -134,7 +134,7 @@ def run(self, scanner_name):
         Returns:
             proto: the returned proto message.
         """
-        request = scanner_pb2.RunRequest(handle=scanner_name)
+        request = scanner_pb2.RunRequest(scanner_name=scanner_name)
```
angelsungoogle: this was fixed before but somehow is now gone. adding it back.

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -906,6 +907,32 @@ def get_latest_inventory_index_id(cls, session):
             inventory_index.id)
         return inventory_index.id
 
+    @classmethod
+    def get_inventory_index_id_by_scanner(cls, session, scanner_index_id):
```
angelsungoogle: when the user runs `forseti notifier run --scanner_index_id XXXXX`, the inventory_index_id is not provided, but will be needed to complete the notifier. This function serves to fill the gap

google/cloud/forseti/notifier/notifier.py
```diff
@@ -95,11 +96,31 @@ def run(inventory_index_id, progress_queue, service_config=None):
 
     violations = None
     with service_config.scoped_session() as session:
-        if not inventory_index_id:
-            inventory_index_id = (
-                DataAccess.get_latest_inventory_index_id(session))
-        scanner_index_id = scanner_dao.get_latest_scanner_index_id(
-            session, inventory_index_id)
+        # if scanner index id is provided,
```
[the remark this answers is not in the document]
angelsungoogle: could you elaborate on "Do the disallow check in the service layer."?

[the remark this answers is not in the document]
angelsungoogle: for the change below, do we also need to do the same change in the notifier definitions?
`notifiers.append(chosen_pipeline(
    resource['resource'], scanner_index_id,
    violation_map[resource['resource']], global_configs,
    notifier_configs, notifier['configuration']))`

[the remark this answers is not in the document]
angelsungoogle: based on our discussion offline, we decided not to change the original inventory_index_id here. This inventory_index_id is later used in the notifier base class to generate filename. Making the change here will cause inconsistency in filename, which we are concerned might affect customers parsing the filename in any way.

angelsungoogle:
closing this pull request as it is now moved to another one

google/cloud/forseti/scanner/scanner_builder.py
```diff
@@ -51,60 +55,80 @@ def build(self):
...
-                            rules)
-
-                scanner = scanner_class(self.global_configs,
-                                        self.scanner_configs,
-                                        self.service_config,
-                                        self.model_name,
-                                        self.snapshot_timestamp,
-                                        rules)
-                runnable_scanners.append(scanner)
+        if self.scanner_name == 'external_project_access_scanner':
```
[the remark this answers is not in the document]
angelsungoogle: well actually, is this dictionary a repetition of the requirements_map?

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -930,6 +931,33 @@ def get_latest_inventory_index_id(cls, session):
...
+            .outerjoin(ScannerIndex,
+                       ScannerIndex.inventory_index_id == InventoryIndex.id)
+            .filter(
+                and_(ScannerIndex.id == scanner_index_id,
+                     or_(InventoryIndex.inventory_status == 'SUCCESS',
+                         InventoryIndex.inventory_status == 'PARTIAL_SUCCESS'))
+            ).order_by(InventoryIndex.id.desc()).first())
+        session.expunge(inventory_index)
+        LOGGER.info(
+            'Latest success/partial_success inventory index id is: %s',
```
blueandgold: This message would be more accurate as `Found inventory_index_id %s from scanner_index_id %s.'

google/cloud/forseti/services/notifier/service.py
```diff
@@ -81,25 +81,39 @@ def Run(self, request, _):
         """
         progress_queue = Queue()
 
-        LOGGER.info('Run notifier service with inventory index id: %s',
-                    request.inventory_index_id)
+        if request.scanner_index_id and request.inventory_index_id:
+            LOGGER.exception('Invalid input: supplying both inventory_index_id '
```
blueandgold: You should also say here, that the scanner command will not run.

google/cloud/forseti/services/notifier/service.py
```diff
@@ -81,25 +81,39 @@ def Run(self, request, _):
         """
         progress_queue = Queue()
 
-        LOGGER.info('Run notifier service with inventory index id: %s',
-                    request.inventory_index_id)
+        if request.scanner_index_id and request.inventory_index_id:
```
blueandgold: Just having the logger.exception message is not enough.  The notifier will still run at the bottom.  You should just return a response right away, to stop the execution here.

google/cloud/forseti/services/notifier/service.py
```diff
@@ -81,25 +81,39 @@ def Run(self, request, _):
...
+            LOGGER.exception('Invalid input: supplying both inventory_index_id '
+                             'and scanner_index_id is not supported.')
+        elif request.scanner_index_id:
```
blueandgold: This should become an `if`, if you return from above.

install/gcp/scripts/run_external_project_access_scanner.sh
```diff
@@ -0,0 +1,85 @@
...
+echo "Finished running Forseti scanner."
+sleep 10s
+
+# Run notifier command
+echo "Running Forseti notifier with specified scanner index id ${scanner_index_id}."
+forseti notifier run --scanner_index_id ${scanner_index_id}
```
blueandgold: Should also update the first cron job to also use specific `--scanner_index_id` as this here?  I am concerned that the first cron job will become broken, if it's running first, and this second cron job gets triggered.

deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -228,6 +230,12 @@ def GenerateConfig(context):
 (echo "{run_frequency} (/usr/bin/flock -n /home/ubuntu/forseti-security/forseti_cron_runner.lock $FORSETI_HOME/install/gcp/scripts/run_forseti.sh || echo '[forseti-security] Warning: New Forseti cron job will not be started, because previous Forseti job is still running.') 2>&1 | logger") | crontab -u $USER -
 echo "Added the run_forseti.sh to crontab under user $USER"
 
+# Adding a second cron job to run the External Project Access Scanner separately if it is enabled 
+if {external_project_access_scanner_cron_enabled}; then
+    (echo "{external_project_access_scanner_cron_schedule} (/usr/bin/flock -n /home/ubuntu/forseti-security/forseti_cron_runner.lock $FORSETI_HOME/install/gcp/scripts/run_external_project_access_scanner.sh || echo '[forseti-security] Warning: New Forseti cron job will not be started, because previous Forseti job is still running.') 2>&1 | logger") | crontab -u $USER -
```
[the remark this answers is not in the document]
angelsungoogle: Is this .lock file a file that gets created here to run the cron job? just double check

install/gcp/scripts/run_external_project_access_scanner.sh
angelsungoogle: where does the task of "getting the last scanner_index_id" happen during the first cron job? is it a one time deal at the beginning, or something that every step runs to get an answer? 
if it's latter, then what if some one manually ran a scan at the same time the cron job is running?

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -930,6 +931,33 @@ def get_latest_inventory_index_id(cls, session):
...
+            scanner_index_id (int): id of the scanner in scanner_index table
+
+        Returns:
+            int64: inventory index id
+        """
+
+        inventory_index = (
+            session.query(InventoryIndex)
+            .outerjoin(ScannerIndex,
+                       ScannerIndex.inventory_index_id == InventoryIndex.id)
```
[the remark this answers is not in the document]
angelsungoogle: uh I totally didn't see the inventory_index_id column before. let me get rid of the join

google/cloud/forseti/scanner/scanner.py
```diff
@@ -86,13 +89,12 @@ def run(model_name=None, progress_queue=None, service_config=None):
         model_name (str): The name of the data model.
         progress_queue (Queue): The progress queue.
         service_config (ServiceConfig): Forseti 2.0 service configs.
-
+        scanner_name (str): name of the scanner that runs separately.
```
blueandgold: Proper case for `Name`

google/cloud/forseti/scanner/scanner_builder.py
```diff
@@ -51,60 +55,87 @@ def build(self):
...
-                                        rules)
-                runnable_scanners.append(scanner)
+        # please note that the scanner name passed in CLI must match the
+        # corresponding key in this dictionary
+        long_running_scanner_factory = {
+            'external_project_access_scanner': (
+                external_project_access_scanner.ExternalProjectAccessScanner,
+                'external_project_access_rules.yaml')
+        }
+        if self.scanner_name in long_running_scanner_factory:
```
blueandgold: Hmmm.  I think that we should check if `self.scanner_name` is populated (i.e. not None) like below.  Otherwise, if someone makes a typo of a scanner name, all the other scanners will run.
```
if self.scanner_name:
    if self.scanner_name in long_running_scanner_factory:
else
    for scanner in self.scanner_configs.get('scanners'):
```

google/cloud/forseti/scanner/scanner_builder.py
```diff
@@ -51,60 +55,87 @@ def build(self):
...
-                scanner = scanner_class(self.global_configs,
-                                        self.scanner_configs,
-                                        self.service_config,
-                                        self.model_name,
-                                        self.snapshot_timestamp,
-                                        rules)
-                runnable_scanners.append(scanner)
+        # please note that the scanner name passed in CLI must match the
+        # corresponding key in this dictionary
+        long_running_scanner_factory = {
```
blueandgold: Can you please put a `TODO` comment, that we will make this use the requirements map in a follow-up PR?
```
# TODO: Use the SCANNER_REQUIREMENTS_MAP instead of the custom map here.
```

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -930,6 +931,32 @@ def get_latest_inventory_index_id(cls, session):
...
+        Args:
+            session (object): Database session
+            scanner_index_id (int): id of the scanner in scanner_index table
+
+        Returns:
+            int64: inventory index id
+        """
+
+        inventory_index = (
+            session.query(InventoryIndex).filter(
```
blueandgold: A couple of comments, where this query should be much simpler:
1: We should query on the `ScannerIndex` table.
2: We should not care about the `inventory_status`, since there is a 1:1 relationship of the `scanner_index_id` to `inventory_index_id`, which is the contract of what this method is suppose to return.

What do you think?

install/gcp/scripts/run_external_project_access_scanner.sh
```diff
@@ -0,0 +1,85 @@
...
+echo "Using model ${MODEL_NAME} to run scanner"
+forseti model use ${MODEL_NAME}
+# Sometimes there's a lag between when the model
+# successfully saves to the database.
+sleep 10s
+echo "Forseti config: $(forseti config show)"
+
+# Run scanner command
+echo "Running Forseti external project access scanner."
+scanner_return=`forseti scanner run --scanner external_project_access_scanner`
```
blueandgold: nit: a better naming here is `scanner_command` instead of `scanner_return`

install/gcp/scripts/run_forseti.sh
```diff
@@ -69,13 +69,14 @@ echo "Forseti config: $(forseti config show)"
 
 # Run scanner command
 echo "Running Forseti scanner."
-forseti scanner run
+scanner_return=`forseti scanner run`
```
blueandgold: ditto on `scanner_command`

angelsungoogle:
Another thing I realized, we are asking users to enable external project access scanner cron job in deployment-forseti-server.yaml, but the notification of this scanner needs to be added to forseti_conf_server.yaml. @blueandgold

google/cloud/forseti/scanner/scanner_builder.py
angelsungoogle: this is actually handled in [the other PR](https://github.com/GoogleCloudPlatform/forseti-security/pull/2324/files#diff-b3ab31b8737b069d164ab56066436a0b). Well I was just having some issue making that PR specific to one file...

google/cloud/forseti/scanner/scanner_builder.py
angelsungoogle: handled[ in the other PR](https://github.com/GoogleCloudPlatform/forseti-security/pull/2324/files#diff-b3ab31b8737b069d164ab56066436a0b)

google/cloud/forseti/services/inventory/storage.py
angelsungoogle: agree!

install/gcp/scripts/run_external_project_access_scanner.sh
angelsungoogle: done

install/gcp/scripts/run_forseti.sh
angelsungoogle: done

angelsungoogle:
@blueandgold I manually ran the two unit tests that failed in Travis, they seemed to pass in my local environment. Is there any way we could identify what's causing it to fail in Travis?

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -930,6 +931,30 @@ def get_latest_inventory_index_id(cls, session):
             inventory_index.id)
         return inventory_index.id
 
+    @classmethod
+    def get_inventory_id_by_scan_id(cls,
```
angelsungoogle: this name change is to conform with the 30 characters limit by pylint....

google/cloud/forseti/scanner/scanner.py
```diff
@@ -86,13 +89,12 @@ def run(model_name=None, progress_queue=None, service_config=None):
         model_name (str): The name of the data model.
         progress_queue (Queue): The progress queue.
         service_config (ServiceConfig): Forseti 2.0 service configs.
-
+        scanner_name (str): Name of the scanner that runs separately.
     Returns:
```
blueandgold: Add a blank line to separate the `Args` section from the `Returns` section.

google/cloud/forseti/services/cli.py
```diff
@@ -334,6 +342,14 @@ def define_notifier_parser(parent):
               'will be used.')
     )
 
+    create_notifier_parser.add_argument(
+        '--scanner_index_id',
+        default=0,
```
blueandgold: What does a default of `0` do?  Would a `None` be more appropriate?

angelsungoogle: agree, however the previous test case, which is for inventory_index_id, defaults to 0. Hmm would it be nice to be consistent?

PR #2290, opened 2018-11-24
Visualize with annotation

joecheuk:


blueandgold:
Adding `cla: yes` label, as the PR & commit authors are googlers at the time of this PR was created.

PR #2297, opened 2018-11-26
Add Docker entrypoint script to start Forseti Server in a container

Red-Five:

- [x ] I signed Google's [Contributor License Agreement](https://opensource.google.com/docs/cla/)
- [ n.a.] My code conforms to Google's [python style](https://google.github.io/styleguide/pyguide.html).
- [ n.a.] My PR at a minimum doesn't decrease unit-test coverage (if applicable).
- [ n.a.] My PR has been functionally tested.
- [ n.a.] All of the [unit-tests](https://forsetisecurity.org/docs/latest/develop/dev/testing.html) still pass.
- [ n.a.] Running `pylint --rcfile=pylintrc` passes.


install/scripts/docker_entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+# This script assumes it's running in a container on a GCE Container Optimized OS VM
+# that is already authorized with the forseti service account.
+# gsutil will automatically use the hosting VM's service account.
+# TODO Other scenarios.
+
+# TODO named arg and arg validation
+# TODO arg to control which services to start (default to server for now)
+BUCKET=$1
+
+# Download config files
```
angelsungoogle: Probably need to include the creation of the buckets in the entrypoint, if we are really automating this. @blueandgold please advice

angelsungoogle: upon further research, I realized that Forseti installation does a lot more than just GCS bucket. Service account, role assignment, firewall rules, etc, all needs to be done during the installation. We definitely want to automate all these steps in the end, and need more discussion regarding what is the best way to automate. For a first step, I suggest that we keep this entrypoint file as is, and advice on using the manual instructions on forseti website to create the resounces.

angelsungoogle: [instruction for manual resource creation](https://forsetisecurity.org/docs/latest/develop/dev/setup.html)

Red-Five: My 2c. Keep the entrypoint focused on just starting Forseti services.

Keep all deployment as part of the CI\CD process.

Forseti runtime deployment of code is not ideal as would prefer to scan images for vulnerabilities before deployment.

PR #2302, opened 2018-11-27
Update BufferedDbWriter to be aware of the size of data sent to SQL.

### LINKED ISSUE #2292: [2-8-0]-90% of ressources are missing in the inventory
Since 2-8-0 inventory takes 5min (30min before)
got only 60 project ( 1250 before) and same for all ressources, violations .....

permissions seems ok, the service account is ok , i have add project viewer on organization level + org admin viewer to test = > same result ...

any idea ?

ahoying:
Resolves #2292.

* Ensure that the size of a flush is never larger than MAX_ALLOWED_PACKET for Cloud Asset data.

### DIFF
--- a/google/cloud/forseti/services/inventory/storage.py
+++ b/google/cloud/forseti/services/inventory/storage.py
@@ -52,6 +52,7 @@ LOGGER = logger.get_logger(__name__)
 BASE = declarative_base()
 CURRENT_SCHEMA = 1
 PER_YIELD = 1024
+MAX_ALLOWED_PACKET = 32*1024*1024  # 32 Mb is the default mysql max packet size
 
 
 class Categories(enum.Enum):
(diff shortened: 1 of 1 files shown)


google/cloud/forseti/services/inventory/storage.py
```diff
@@ -52,6 +52,7 @@
 BASE = declarative_base()
 CURRENT_SCHEMA = 1
 PER_YIELD = 1024
+MAX_ALLOWED_PACKET = 32 * 1024 * 1024  # 32 Mb default mysql max packet size
```
joecheuk: I think the default max allowed packet is 4mb in mysql 5.7
https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_max_allowed_packet

ahoying: That is the default, but this is how CloudSQL sqlproxy is configured:

```
mysql> select @@max_allowed_packet;
+----------------------+
| @@max_allowed_packet |
+----------------------+
|             33554432 |
+----------------------+
1 row in set (0.00 sec)

mysql> select @@max_allowed_packet/1024/1024;
+--------------------------------+
| @@max_allowed_packet/1024/1024 |
+--------------------------------+
|                    32.00000000 |
+--------------------------------+
1 row in set (0.00 sec)
```

PR #2305, opened 2018-11-27
When nothing is available to scan return empty set

knitterb:
This resolves a problem where the Enabled API and IAM Rules scanners have no results.  Previously they issued a system exit, now they return empty results.


knitterb:
I signed it!

knitterb:
I signed it again!

knitterb:
I signed it!

knitterb:
I will resubmit with the proper author

PR #2306, opened 2018-11-27
Bigquery retention inventory part (for review)

gruihuang:
Add Bigquery table into the GCP inventory, which enables the model to get expiration Time data that would be used in the Bigquery retention.


umairidris:
Please rename the PR name.

tests/common/gcp_api/test_data/fake_bigquery.py
```diff
@@ -118,3 +118,110 @@
...
+  ],
+  "totalItems": 2
+}
+"""
+
+GET_TABLES_RESPONSES = [GET_TABLES_PAGE_1,
+                        GET_TABLES_PAGE_2]
+
+TABLES_GET_EXPECTED = [
+    {u'kind': u'bigquery#table',
```
dufton: do you need the unicode prefix on all these strings? If not, remove it to have less visual clutter.

PR #2307, opened 2018-11-27
Make scanner exit when no results are found by scanner.

knitterb:
This resolves an issue where no results are found and the scanner would hang.   The scanner will no longer exit when there are no results, instead returning an empty set.


google/cloud/forseti/scanner/scanners/enabled_apis_scanner.py
```diff
@@ -135,7 +134,7 @@ def _retrieve(self):
 
         if not enabled_apis_data:
             LOGGER.warn('No Enabled APIs found. Exiting.')
-            sys.exit(1)
+            return iter([])
```
blueandgold: This should raise a `NoDataError` exception, 
https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/common/util/errors.py#L42-L44

Then in the `run()` method, you can do this, which will exit cleanly and early.
```
try:
    enable_apis_data = self._retrieve()
except NoDataError:
    return
all_violations = self._find_violations(enabled_apis_data)
self._output_results(all_violations)
```

knitterb: I've made the modifications to the two scanners.  Please let me know if you have further suggestions.

google/cloud/forseti/scanner/scanners/enabled_apis_scanner.py
```diff
@@ -15,7 +15,6 @@
 """Scanner for Enabled APIs."""
 
 import json
-import sys
 
 from google.cloud.forseti.common.gcp_type.project import Project
 from google.cloud.forseti.common.util import logger
```
blueandgold: Hi Brandon, you will have to import like this.  Thanks.
```
from google.cloud.forseti.common.util import errors as util_errors

raise util_errors.NoDataError

except util_errors.NoDataError
```

https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/common/util/errors.py#L42-L44

knitterb: Just caught that.  New build coming.  Sorry about that.

google/cloud/forseti/scanner/scanners/enabled_apis_scanner.py
```diff
@@ -20,6 +20,7 @@
...
 from google.cloud.forseti.scanner.audit import enabled_apis_rules_engine
 from google.cloud.forseti.scanner.scanners import base_scanner
+from google.cloud.forseti.common.util import errors as util_errors
```
blueandgold: Can you please alpha-sort this import?

google/cloud/forseti/scanner/scanners/enabled_apis_scanner.py
```diff
@@ -134,15 +135,15 @@ def _retrieve(self):
 
         if not enabled_apis_data:
             LOGGER.warn('No Enabled APIs found. Exiting.')
-            raise NoDataError('No enabled APIs found. Exiting')
+            raise util_errors.NoDataError('No enabled APIs found. Exiting')
```
blueandgold: Can you please document the raise in the method doc? (underneath the return)
```
Raises:
    NoDataError: If enabled_api data is not found.
```

knitterb:
@blueandgold I rebased to upstream.  Looks like everything is passing but the code coverage, did you need that resolved as well?  I ask because it was not marked as "required".

PR #2323, opened 2018-11-30
Fix blacklist scanner: make the scanner handle no retrieve-able data

### LINKED ISSUE #2294: Blacklist scanner crashes if no network interfaces found
> [forseti-security][2.7.0] google.cloud.forseti.scanner.scanners.blacklist_scanner(get_instance_networks_interfaces): No VM network interfaces found. Exiting.
> [forseti-security][2.7.0] google.cloud.forseti.scanner.scanner(run): Error running scanner: BlacklistScanner Traceback (most recent call last): File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.7.0-py2.7.egg/google/cloud/forseti/scanner/scanner.py", line 111, in run scanner.run() File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.7.0-py2.7.egg/google/cloud/forseti/scanner/scanners/blacklist_scanner.py", line 144, in run self._find_violations(instances_network_interface_data)) File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.7.0-py2.7.egg/google/cloud/forseti/scanner/scanners/blacklist_scanner.py", line 137, in _find_violations all_violations.extend(violations) File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.7.0-py2.7.egg/google/cloud/forseti/scanner/audit/blacklist_rules_engine.py", line 216, in find_violations for network_interface in instance_network_interface: TypeError: 'NoneType' object is not iterable

The _retrieve code in the rule generator is hitting this line:

```
        if not network_interfaces:
            LOGGER.warn('No VM network interfaces found. Exiting.')
            return None, 0
```

I think just making this return [], 0 should work? Though I haven't dug through the rule generator code to see what would happen in this case.

blueandgold:
Fixes #2294 

Similar to #2307


google/cloud/forseti/scanner/scanners/blacklist_scanner.py
```diff
@@ -113,8 +118,9 @@ def get_instance_networks_interfaces(self):
             network_interfaces.append(ins.create_network_interfaces())
 
         if not network_interfaces:
-            LOGGER.warn('No VM network interfaces found. Exiting.')
-            return None, 0
+            error_message = 'No VM network interfaces found. Exiting.'
```
umairidris: nit: 'Exiting' can be a bit misleading as this isn't doing a fatal log or os.exit here. The exception in a normal situation is handled and ignored in the 'run' function so the binary didn't 'Exit'.

google/cloud/forseti/scanner/scanners/blacklist_scanner.py
```diff
@@ -146,7 +152,11 @@ def _find_violations(self, instances_networks_data):
 
     def run(self):
         """Runs scanning."""
-        instances_network_interface_data = self._retrieve()
+        try:
```
umairidris: optional: I might be missing some background here, but is this behaviour consistent with other scanners? Most scanners just seem to call _retrieve and blindly pass in the result into the rule engine (even if it's empty), and the retrieve function does not check or raise any exceptions either. I wonder if we can do that? Calling the rule engine can still be desirable if there are rules with 'mode: required', etc.

If not, maybe just return the empty list from retrieve and make the check here
```
if not instances_network_data:
   LOGGER.info('No VM network interfaces found, nothing to do'.)
   return
```

PR #2336, opened 2018-12-03
Add a deploy condition so that we don't deploy for forks.

mirons-google:
The conditions placed here should have the consequence of only building when a merge is made from the development branch of the website to the production version of the website by pull request where the travis encrypted variables are present.


mirons-google:
This change should have that consequence as a side effect. I'll add it to the description, but leave the subject the same as that's what it is specifically changing.

mirons-google:
Done. PTAL.

PR #2349, opened 2018-12-05
Update group_rules.yaml to have google related service account by default

BrunoReboul:
update version of group scanner default yaml file to have in the default service accounts


rules/group_rules.yaml
```diff
@@ -19,6 +19,10 @@
       # Must use quotes if the condition starts with @ symbol.
     - member_email: "@{DOMAIN}"
     - member_email: "@gmail.com"
+    # Service Accounts
```
blueandgold: Can we make this comment more clear and descriptive of what the `gserviceaccount.com` are?  Such as:
```
GCP service accounts
```

rules/group_rules.yaml
```diff
@@ -19,6 +19,10 @@
...
+    - member_email: "gserviceaccount.com"
```
blueandgold: Can we make this opt-in by default, i.e. by commenting this out?  Then if people want it, then can uncomment it.
```
# - member_email: "gserviceaccount.com"
```

rules/group_rules.yaml
```diff
@@ -19,6 +19,10 @@
...
+    # Big Query Transfer Service
+    - member_email: "@bqdts.google.baggins"
```
blueandgold: ditto on making this opt-in

BrunoReboul:
Thanks you for the feedback.
I have make the 3 changes

PR #2369, opened 2018-12-06
Updated KE Cluster to use unique ID instead of display name.

### LINKED ISSUE #2367: [v2.9.0] KeVersion Scanner error
```
google.cloud.forseti.scanner.scanner(run): Error running scanner: KeVersionScanner Traceback (most recent call last): 
File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.9.0-py2.7.egg/google/cloud/forseti/scanner/scanner.py", line 117, 
in run scanner.run() File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.9.0-py2.7.egg/google/cloud/forseti/scanner/scanners/ke_base_scanner.py", line 214,
in run ke_clusters = self._retrieve() File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.9.0-py2.7.egg/google/cloud/forseti/scanner/scanners/ke_base_scanner.py", line 206, 
in _retrieve parent_type_name=ke_cluster_type_name))[0] IndexError: list index out of range
```

joecheuk:
resolves #2367

### DIFF
--- a/google/cloud/forseti/common/gcp_type/ke_cluster.py
+++ b/google/cloud/forseti/common/gcp_type/ke_cluster.py
@@ -22,6 +22,7 @@ import json
 from google.cloud.forseti.common.gcp_type import resource
 from google.cloud.forseti.services.inventory.base.resources import size_t_hash
 
+
 # pylint: disable=too-many-arguments,too-many-instance-attributes
 # pylint: disable=too-many-locals,missing-param-doc,missing-type-doc
 # pylint: disable=invalid-name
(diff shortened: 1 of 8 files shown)


google/cloud/forseti/common/gcp_type/ke_cluster.py
```diff
@@ -89,22 +91,24 @@ def __init__(self, cluster_id, parent=None, full_name=None, locations=None,
         self._dict = None
 
     @classmethod
-    def from_json(cls, parent, json_string):
+    def from_json(cls, parent, json_string, cluster_id=None):
```
umairidris: Unfortunately  adding cluster_id here will not work too well for resource_util that calls from_json in a generic way: https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/common/gcp_type/resource_util.py#L106

Can we deduce the cluster_id from the JSON response? The cluster.name being passed in as a kwarg must be originally have been created from the JSON response, so it should be possible.

joecheuk: Updated to generate the key on the fly using hash(selfLink)

Issue #2372, opened 2018-12-07
FR - Lien scanner - enable * wildcard in project ids


BrunoReboul:
Feature Request

Scanner: lien
Request:
- enable usage of * wild card in resource_ids list

User story:
As an operation manager, I want to control that all the production projects in my organization are protected from deletion by a lien using the naming convention of project when all production projects are not below one folder in the GCP hierarchy.

Example: 
- naming convention:  production project must have a suffix '-prod'
- scanner resource_ids will be: - '*-prod'

jjawad-google:
Hi @umairidris could you take a look?

umairidris:
@BrunoReboul most scanners don't support wildcard with text in resource ids. It's either '*' or a non regex string.

It doesn't seem like that would be sufficient for your use case, but I wanted to confirm.

umairidris:
Wildcard only resource id support is added in #2380.

PR #2380, opened 2018-12-10
Support wildcard in lien resources

umairidris:
This PR adds supports for a wildcard only resource id for lien rules.

#2372


google/cloud/forseti/scanner/audit/lien_rules_engine.py
```diff
@@ -197,6 +197,10 @@ def find_violations(self, parent_resource, liens):
 
         for res in resource_ancestors:
             applicable_rules.extend(self.resource_to_rules.get(res, []))
+            wildcard_res = resource_util.create_resource(
```
joecheuk: if we put this in the loop, will the same rule gets added multiple times to applicable_rules?

joecheuk:
Thanks for updating this!

PR #2396, opened 2018-12-13
Bigquery retention model and scanner part

gruihuang:


google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -28,7 +28,7 @@
 
 LOGGER = logger.get_logger(__name__)
 
-SUPPORTED_RETENTION_RES_TYPES = frozenset(['bucket'])
+SUPPORTED_RETENTION_RES_TYPES = frozenset(['bucket', 'bigquery_table'])
```
umairidris: nit: can you use the resource constants to reduce duplication?

https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/scanner/audit/location_rules_engine.py#L33

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -353,15 +377,70 @@ def find_violations_in_bucket(self, bucket):
...
+        if self.max_retention is None:
+            return
+
+        table_dict = json.loads(table.data)
+        table_expiration = table_dict.get('expirationTime')
+        if not table_expiration:
+            self.generate_table_violation(table)
+
+        table_creation = table_dict.get('creationTime')
+        diff = long(table_expiration) - long(table_creation)
```
umairidris: won't this fail if there is table_expiration is None? you might need to make the .get call above return 0 by default

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -353,15 +377,70 @@ def find_violations_in_bucket(self, bucket):
...
+        """
+        if self.max_retention is None:
+            return
+
+        table_dict = json.loads(table.data)
+        table_expiration = table_dict.get('expirationTime')
+        if not table_expiration:
+            self.generate_table_violation(table)
+
+        table_creation = table_dict.get('creationTime')
```
umairidris: if you're assuming creationTime is set then use table_dict['creationTime']. That way if this ever fails this line will make it clear which key is missing

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -960,6 +962,15 @@ def _convert_service_config(self, service_config):
 
         self.session.add(resource)
 
+    def _convert_bigquery_table(self, table):
+        """Convert a table to a database object.
+
+        Args:
+            table (object): table to store.
+        """
+        # cached should be True or False?
```
umairidris: make consistent with other bq resources?

google/cloud/forseti/scanner/audit/retention_rules_engine.py
gruihuang: Good shot.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -353,15 +379,70 @@ def find_violations_in_bucket(self, bucket):
...
 
+    def table_max_retention_violation(self, table):
+        """Get a generator for violations especially for maximum retention
+           It only supports bucket for now, and will work on generalizing
+           in future PRs.
+
+        Args:
+            table (Table): Find violation from the table.
+        Yields:
+            RuleViolation: All max violations of the table breaking the rule.
```
joecheuk: "Retention violation of bigquery tables." might read better

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -243,7 +243,7 @@ def test_only_max_normal_delete(self):
         """Test that a bucket's rule can guarantee the maximum_retention if its
         action is 'Delete' and the only condition is an age(<= maximum_retention)"""
         rules_engine = get_rules_engine_with_rule(RetentionRulesEngineTest.yaml_str_only_max_retention)
-        self.assertEqual(1, len(rules_engine.rule_book.resource_rules_map))
+        self.assertTrue(1 <= len(rules_engine.rule_book.resource_rules_map))
```
dufton: This should be a failure if we don't get the exact number of expected rules.

You can use a constant if you don't want to update several lines any time you update the test rule strings.

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -851,6 +851,130 @@ def test_bucket_on_multi_project_has_vio(self):
...
+        data_creater = frsd.FakeTableDataCreater('fake_bqtable', frsd.DATASET1)
+
+        fake_table = data_creater.get_resource()
+        got_violations = list(rules_engine.find_violations(fake_table))
+        expected_violations = frsd.build_table_violations(
+            fake_table, 'bigquery retention on projects')
+        self.assertEqual(got_violations, expected_violations)
+
+    def test_bigquery_retention_on_project_too_big(self):
+        """Test that a rule with a resource.type equal to 'project'"""
```
dufton: Update doc strings (here and below) to cover what "too big", "too small", etc mean.

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -353,15 +377,70 @@ def find_violations_in_bucket(self, bucket):
...
+            return
+
+        table_dict = json.loads(table.data)
+        table_expiration = table_dict.get('expirationTime')
+        if not table_expiration:
+            yield self.generate_table_violation(table)
+        else:
+            table_creation = table_dict['creationTime']
+            diff = long(table_expiration) - long(table_creation)
+            if diff > self.max_retention * 24 * 3600000:
```
dufton: Make this a constant, e.g.
`_MS_PER_DAY = 24 * 60 * 60 * 1000`

Same in the test files

google/cloud/forseti/scanner/audit/retention_rules_engine.py
```diff
@@ -79,8 +81,9 @@ def find_violations(self, resource, force_rebuild=False):
 
         violations = itertools.chain()
         resource_rules = self.rule_book.get_resource_rules(resource.type)
+        res = resource_util.create_resource(resource.id, resource.type)
         resource_ancestors = (relationship.find_ancestors(
-            resource, resource.full_name))
+            res, resource.full_name))
```
dufton: This won't work with `resource` instead of creating a new `res` object?

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -571,6 +573,7 @@ def _store_resource(self, resource):
             'spanner_database': self._convert_spanner_db_resource,
             'spanner_instance': self._convert_spanner_resource,
             'subnetwork': self._convert_computeengine_resource,
+            'bigquery_table': self._convert_bigquery_table,
```
dufton: Place in alphabetical order

google/cloud/forseti/services/model/importer/importer.py
```diff
@@ -960,6 +963,15 @@ def _convert_service_config(self, service_config):
 
         self.session.add(resource)
 
+    def _convert_bigquery_table(self, table):
```
dufton: It seems like we could remove a lot of these "convert" functions that just directly call `_convert_resource`. Those that only call with `cached=False` can simply use `_convert_resource` directly in the `handlers` dictionary. For those with `cached=True`, we could define a method like:
```
  def _convert_resource_cached(self, resource):
    self._convert_resource(resource, cached=True)
````

This can be used by bucket, dataset, gae_resource, kubernetes cluster, etc.

It also means you can get rid of the `pylint: disable=too-many-lines`

File: tests/scanner/audit/retention_rules_engine_test.py [hunk 12]
```diff
@@ -239,11 +239,17 @@ def test_invalid_rule_with_no_res_id(self):
     maximum_retention: 365
 
 """
+
+    def test_number_of_rules(self):
+        """The number of rules should be exactly the same as the length of SUPPORTED_RETENTION_RES_TYPES."""
+        rules_engine = get_rules_engine_with_rule(RetentionRulesEngineTest.yaml_str_only_max_retention)
+        self.assertTrue(len(rre.SUPPORTED_RETENTION_RES_TYPES) == len(rules_engine.rule_book.resource_rules_map))
```
dufton: Use a literal constant here, and use `assertEqual`, e.g. `self.assertEqual(2, len(...))`

You should also check that the number of rules is correct, i.e. 1 rule for buckets, 0 rules for tables.

tests/scanner/audit/retention_rules_engine_test.py
```diff
@@ -851,6 +821,128 @@ def test_bucket_on_multi_project_has_vio(self):
             fake_bucket, 'bucket retention on multi projects')
         self.assertEqual(got_violations, expected_violations)
 
+    yaml_str_bigquery_retention_on_projects = """
```
dufton: Also check that this gets parsed into the correct number of rules (1 for tables, 0 for buckets)

tests/scanner/scanners/retention_scanner_test.py
```diff
@@ -399,3 +425,58 @@ def test_bucket_retention_on_multi_rules_on_a_project(self):
...
+          - "def-project-5:ds01.fake-table-01"
+    maximum_retention: 100
+
+"""
+
+        table_test_data=[
+            frsd.FakeTableDataInput(
+                table_id='fake-table-01',
+                dataset=frsd.DATASET1,
+                expiration_time=frsd.DEFAULT_TABLE_CREATE_TIME+110*3600000*24
```
dufton: Use a `_MS_PER_DAY = 24 * 60 * 60 * 1000` constant in this file too

google/cloud/forseti/services/model/importer/importer.py
gruihuang: Should I make this modification in another pull request?

dufton: Up to you

PR #2416, opened 2018-12-15
Docker poc refactoring

Red-Five:
Do not merge yet. Placeholder to track refactoring and cleanup etc.
Main purpose is to support various ways the docker container will be run
e.g.
GKE CronJob, short life
GKE Service, long life
Control which services run in the container
Externally trigger CronJob when running as GKE Service
Control log level
Error handling


angelsungoogle:
Thanks for the improvements! I'm working on putting an FAQ in the website to document the resources available here. A couple questions:
- do you plan to add cos option as well? if not I can do that in a separate branch
- what's the purpose of "Control which services run in the container"?

Red-Five:
> Thanks for the improvements! I'm working on putting an FAQ in the website to document the resources available here. A couple questions:
> 
> * do you plan to add cos option as well? if not I can do that in a separate branch
> * what's the purpose of "Control which services run in the container"?

1. Yes, docker_entrypoint.sh will (shortly) work on cos and gke 

2. Was thinking to re-use same image (for now at least) for both server and client.

angelsungoogle:
got it. I've started #2420 to update the documentation. Please feel free to comment.

install/scripts/docker_entrypoint.sh
```diff
@@ -16,97 +16,168 @@
...
+RUN_SERVER=true
+RUN_CLIENT=false
+
+# Note
+# CLOUDSQLPROXY_SERVICE_HOST
+# CLOUDSQLPROXY_SERVICE_PORT
+# are environment variables set by k8s
+# TODO process these as command line args if not running on k8s
+
+download_configuration_files(){
```
angelsungoogle: what will be the service account here to access the configuration files? the compute engine default service account, or the service account specified by the `GOOGLE_APPLICATION_CREDENTIALS` environment variable?

angelsungoogle: If its the default service account, then it might not have the authorization to access the bucket unless the bucket is in the same project. would it be good to do something like [this](https://cloud.google.com/sdk/gcloud/reference/auth/activate-service-account)?

Red-Five: Forseti service account credentials json file added as k8s secret and mounted to the images.
Cloud SQL proxy refs the file directly; For Forseti we set the env variable GOOGLE_APPLICATION_CREDENTIALS to point to the mounted secret.

ahoying: We should not be using exported service account keys in general. Getting hold of the key file is equivalent to being able to anonymously assume the identity of that service account from any system. There is no current support binding between the service account and the k8s cluster.

Rather the requirement at this time is to run a node pool for forseti that has the node pool default service account set to the forseti service account, and use the existing default service account code to get the credential from metadata.

When per pod identity is supported in gke, then we should evaluate switching to that.

angelsungoogle: > Forseti service account credentials json file added as k8s secret and mounted to the images.
> Cloud SQL proxy refs the file directly; For Forseti we set the env variable GOOGLE_APPLICATION_CREDENTIALS to point to the mounted secret.

yes, I'm asking because I just deployed a cluster using the GCP resources(bucket and credentials etc) from another project (didn't want to create a new set of duplicate resources). The gsutil command gave an error in the pod log saying the compute engine default service account cannot download the bucket config file. it seems that the gsutil command is called with the default service account, instead of the service account in the credentials json file

I then tried to create the bucket on this project by installing forseti, that fixed the issue.

angelsungoogle: > We should not be using exported service account keys in general. Getting hold of the key file is equivalent to being able to anonymously assume the identity of that service account from any system. There is no current support binding between the service account and the k8s cluster.
> 
> Rather the requirement at this time is to run a node pool for forseti that has the node pool default service account set to the forseti service account, and use the existing default service account code to get the credential from metadata.
> 
> When per pod identity is supported in gke, then we should evaluate switching to that.

so if we are using the default service account instead, that means we need to run the iam assignment process on that service account before starting sql proxy and the forseti server, correct?

ahoying: Right.

1) Create service account
2) Configure IAM bindings for resources the service account needs access to
3) Create a new node pool in the cluster, associate it with service account from 1
4) Ensure forseti pods target the correct node pool, https://cloud.google.com/kubernetes-engine/docs/concepts/node-pools#deploying_services_to_specific_node_pools

Red-Five: fyi my clients intention is to potentially run Forseti server on docker/k8s outside of GCP (on another cloud provider that shall not be named :open_mouth: ) They are aware of the need for secure practices around key management.

angelsungoogle: yes we are aware of that. we have a few questions regarding how your organization wanna proceed with Forseti, messaged you via slack ;)

Red-Five:
@angelsungoogle Please feel free to proceed with the review/merge process for this batch of changes to the docker-poc branch.

angelsungoogle:
Thanks for the refactoring @Red-Five ! 
regarding what this PR provides, I've marked the status of the following items, could you confirm?
[YES] GKE CronJob, short life
[YES] GKE Service, long life
[YES] Control which services run in the container
[NO] Externally trigger CronJob when running as GKE Service
[YES] Control log level
[YES] Error handling

Red-Five:
###
docker_entrypoint.sh
 
Added command line arguments to enable more flexibility in entrypoint usage
--bucket
--log_level
--run_server
--run_client
--run_cronjob
--services
--sql_host
--sql_port

Default script behavior is to run server ("scanner model inventory explain notifier")
if only --bucket is specified

Default cloud sql host and port settings to work on cos.
Automatically over-ride cloud sql host port settings on k8s.

Stop script on error.

Refactored into functions.

###
cloudbuild.yaml
Enable caching of base image in build process to reduce build time.

###
forseti.template.yaml renamed to forseti.cronjob.template.yaml
 and template modified to specify docker_entrypoint arguments

###
cos_setup_forseti.sh modified to specify docker_entrypoint --bucket argument

Red-Five:
Running all as k8s services is still a todo as is proper error handling.

angelsungoogle:
got it. thanks for clarify! all good to merge

Issue #2415, opened 2018-12-14
Forseti server service account doesn't have storage.objects.list permissions


knitterb:
Version: v2.9.0
Module: Installer
Installation Method: `$ python install/gcp_installer.py`

When installing Forseti through Cloud Shell, the Forseti server doesn't start up properly due to a missing permission on the server service account (see logs below).  This appears to be a race condition where the permissions are not set in time for the server's local installation to complete.

This issue is exacerbated where the account used for installation does not have Org level permissions, resulting in a delay for the application of the `storage.objects.list` permission on the service account.  To be clear, the permissions at the org level fail as expected for the lack of Org level permissions, but this is a supported installation use case per [link](https://forsetisecurity.org/docs/latest/configure/general/non-org-root.html#how-to-install).  The delay (and order of applied permissions) causes the required permission to not be set on time.

The `storage.objects.list` permissions should be set on the service account earlier in the installation process to account for the delay due to lack of Org level permissions.

From `/tmp/deployment.log`:
```
Finished processing dependencies for forseti-security==2.9.0
AccessDeniedException: 403 forseti-server-gcp-29969b3@knitter-forseti.iam.gserviceaccount.com does not have storage.objects.list access to forseti-server-29969b3.
AccessDeniedException: 403 forseti-server-gcp-29969b3@knitter-forseti.iam.gserviceaccount.com does not have storage.objects.list access to forseti-server-29969b3.
Cloud SQL Instance Connection string: knitter-forseti:us-central1:forseti-server-db-29969b3
SQL port: 3306
Forseti DB name: forseti_security
Could not find the configuration file: /home/ubuntu/forseti-security/configs/forseti_conf_server.yaml.
Starting services.
Failed to start cloudsqlproxy.service: Unit cloudsqlproxy.service not found.
Attempting to update database schema, if necessary.
['/home/ubuntu/forseti-security/install/gcp/upgrade_tools/db_migrator.py']
Traceback (most recent call last):
  File "/home/ubuntu/forseti-security/install/gcp/upgrade_tools/db_migrator.py", line 201, in <module>
    inventory_dao.initialize(SQL_ENGINE)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.9.0-py2.7.egg/google/cloud/forseti/services/inventory/storage.py", line 1004, in initialize
    BASE.metadata.create_all(engine)
[...]

PR #2419, opened 2018-12-17
+ Move up service account permissions to be before deploy func

hshin-g:
+ Move up service account permissions to be before deploy func

References Issue #2415


install/gcp/installer/forseti_client_installer.py
```diff
@@ -49,14 +49,15 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
             bool: Whether or not the deployment was successful
             str: Deployment name
         """
+        gcloud.grant_client_svc_acct_roles(
```
blueandgold: Moving this up will alleviate the race condition, but will not guarantee that it is solved in a deterministic way.

What we really want to do, is to verify after the roles are granted, that the permissions are indeed enabled on the platform, before doing the subsequent deploy steps.

It's not clear what we can do to deterministically verify the permissions are indeed enabled.  Use gcloud to do a list iam policy?  

Or should we set a reasonable pause of N seconds?

hshin-g:
Close in lieu of #2497.

PR #2435, opened 2018-12-20
Add resource scanner

### LINKED ISSUE #2434: Verify resource existence
Today, while we can create rules on resources there is no guarantee these rules will actually match anything. There is no link between the Forseti inventory and the scanner rules. This can lead to nasty bugs where one may assume rules will apply to resources, but the resources are never found/matched.

Proposal: add a resource scanner that can check for a resource tree or multiple sub-trees.

umairidris:
Fix #2434.


tests/scanner/audit/resource_rules_engine_test.py
```diff
@@ -0,0 +1,234 @@
+# Copyright 2017 The Forseti Security Authors. All rights reserved.
```
dufton: Add a test for:
* Resource is present but under the wrong parent (e.g. project id is under the wrong folder/org)
* Expected child resource is missing.

umairidris: Done.

umairidris:
@joecheuk done.

rules/resource_rules.yaml
```diff
@@ -0,0 +1,42 @@
...
+#       resource_id: 'proj-1:dataset2'
+#   - type: 'project'
+#     resource_id: 'proj-2'
+#     children:
+#     - type: 'bucket'
+#       resource_id: 'bucket-3'
+#     - type: 'dataset'
+#       resource_id: 'proj-2:dataset1'
+#     - type: 'instance'
+#       resource_id: '*'
```
joecheuk: can we do regex matching like 'prod-*' here?

umairidris: Currently only supports '*'. Clarified in the doc.

PR #2440, opened 2018-12-21
k8s Services

Red-Five:
Placeholder for changes to support running forseti as a long running k8s service.

Do not merge yet. This is just back of the envelope code right now.


install/scripts/docker_entrypoint.sh
```diff
@@ -13,13 +13,34 @@
...
+# --bucket <bucket>                             the Forseti GCS bucket containing configuration files etc
+# --log_level <info,debug,etc>                  the Forseti server log level
+# --run_server                                  start the Forseti server
+# --services <list of services>                 over-ride default services "scanner model inventory explain notifier"
+# --run_client                                  just provide a container to run client commands
+# --run_cronjob                                 run the cronjob immediately after starting server, for k8s CronJob
+# --sql_host <host ip>                          over-ride k8s (CLOUDSQLPROXY_SERVICE_HOST), cos (localhost) default
+# --sql_port <port>                             over-ride k8s (CLOUDSQLPOXY_SERVICE_PORT), cos (3306) default
+
+# Use cases
```
angelsungoogle: @blueandgold these are the use cases we plan to cover for the experimental version. please let me know if you have any questions.

install/scripts/k8s_setup_forseti.sh
```diff
@@ -42,43 +65,28 @@
...
+elif ${DEPLOY_SERVER}; then
     # Example, forseti server as k8s Cluster IP Service
     # Deploy Cloud SQL Proxy in its own pod
     # Create a Cluster IP Service for Cloud SQL Proxy
     # Deploy forseti server in its own pod
     # Create a Cluster IP Service for forseti server
-    # TODO As PoC deploy client in its own pod ?
-    # TODO As PoC Create a Cluster IP Service for forstei client?
-
+    # TODO As PoC deploy client in its own pod / service?
```
angelsungoogle: @blueandgold suggestions?

angelsungoogle: discussed with @blueandgold and we think it's best to deploy client as its own long-running pod. For the K8s Cron Job case where the server is not always available, we think a client deployment is not necessary in that case but are open to what works best for you, since you are the primary user of that case.

install/scripts/k8s_setup_forseti.sh
```diff
@@ -56,7 +58,29 @@
...
 	kubectl apply -f forseti.cronjob.yaml
+
+elif [[ ${DEPLOY_SERVER}=true ]]; then
+    # Example, forseti server as k8s Cluster IP Service
+    # Deploy Cloud SQL Proxy in its own pod
+    # Create a Cluster IP Service for Cloud SQL Proxy
+    # Deploy forseti server in its own pod
+    # Create a Cluster IP Service for forseti server
+    # TODO As PoC deploy client in its own pod ?
+    # TODO As PoC Create a Cluster IP Service for forstei client?
```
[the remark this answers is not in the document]
angelsungoogle: after discussion with @blueandgold , we suggest as PoC the client could be its own pod, and have a Cluster IP service to be accessible from only within the cluster

install/scripts/docker_entrypoint.sh
```diff
@@ -191,16 +233,16 @@ main(){
...
     # Run server or client; not both in same container
-    if [[ ${RUN_SERVER}="true" ]]; then
+    if ${RUN_SERVER}; then
+        download_server_configuration_files
         start_server
-    elif [[ ${RUN_CLIENT}="true" ]]; then
-        start_client
+    elif ${RUN_CLIENT}; then
+        download_client_configuration_files
+        # start_client not needed?
```
angelsungoogle: @blueandgold

PR #2451, opened 2018-12-27
Ensure tests are always consistent when dicts are translated to strings.

ahoying:
* Update all usages of json.dumps() to use sort_keys=True when the results of the call may be used in a comparison or test.

* Ensure all other string serialization functions used in comparisons and tests return consistent results.

* Fix various issues in tests that can cause them to be flaky on some platforms and fail some percentage of runs.


google/cloud/forseti/common/data_access/csv_writer.py
```diff
@@ -477,6 +478,24 @@
...
+
+    Returns:
+        dict: A row with nested dicts transformed to json string.
+    """
+    new_row = {}
+    for key, value in row.items():
+        if isinstance(value, dict):
+            new_row[key] = json.dumps(value, sort_keys=True)
+        else:
+            new_row[key] = value
```
joecheuk: I think the newly created dict might have different order when being called multiple times with the same input, does it affect the comparison? 

e.g. 
input: {a: 1, b:2}
output can either be {b:2, a:1} or {a:1, b:2}

tests/scanner/groups_scanner_test.py
```diff
@@ -40,9 +40,21 @@ def _render_ascii(self, starting_node, attr):
...
+                lines = value
+            else:
+                lines = str(value).split("\n")
+            rows.append(u"%s%s" % (pre,
+                                   json.dumps(lines[0], sort_keys=True)))
+            for line in lines[1:]:
+                rows.append(u"%s%s" % (fill,
+                                       json.dumps(line, sort_keys=True)))
+
+        return str('\n'.join(rows))
```
joecheuk: nit: join method returns a string, the str() is not needed

ahoying: Good catch, left over from testing.

PR #2475, opened 2019-01-07
Update Dev Env Setup Instruction

blueandgold:
Update the dev environment instructions with what we learned recently, and have repeated on a virtualbox OS X instance.

Also in this PR, is moving the troubleshooting section to the bottom.

### DIFF
--- a/_docs/_latest/develop/dev/setup.md
+++ b/_docs/_latest/develop/dev/setup.md
@@ -42,8 +42,12 @@ Use the following commands to install the necessary dependencies:
 
 Install python-dev:
 
+The system python that comes with OS X does not allow certain packages
+(such as six) to be modified. Install brew’s version of python so that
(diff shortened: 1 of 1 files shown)


_docs/_latest/develop/dev/setup.md
```diff
@@ -155,41 +205,45 @@ try the following:
 
 ------------------
```
osandoval011: Are you sure about this?  I was able to fix just by adding a dummy my_config.h file to the directory where #includes looks from.  But if this works then this is a better solution.  I had read that python expects mysql to need the my_config even though this is no longer the case hence why the dummy file works

blueandgold: Yes, I am sure about this, this was the only thing I had to try to fix this problem.

Issue #2467, opened 2019-01-04
forseti inventory create fails: can't reconnect until invalid transaction is rolled back


rtoma:
Hi,

We are running Forseti 2.9.0 on a GCE VM using CloudSQL mysql as DB. Server and client are running on same VM. All manually deployed.

`forseti inventory create` fails on:
```
Jan  4 07:00:18 localhost forseti-cron: + forseti inventory create --import_as 20190104T060018
Jan  4 07:05:37 localhost forseti-cron: {
Jan  4 07:05:37 localhost forseti-cron:   "errors": 0,
Jan  4 07:05:37 localhost forseti-cron:   "warnings": 0,
Jan  4 07:05:37 localhost forseti-cron:   "step": "",
Jan  4 07:05:37 localhost forseti-cron:   "finalMessage": false,
Jan  4 07:05:37 localhost forseti-cron:   "lastError": "",
Jan  4 07:05:37 localhost forseti-cron:   "id": "1546581619066946",
Jan  4 07:05:37 localhost forseti-cron:   "lastWarning": ""
Jan  4 07:05:37 localhost forseti-cron: }
Jan  4 07:05:37 localhost forseti-cron: Error occurred on the server side, message: <_Rendezvous of RPC that terminated with:
Jan  4 07:05:37 localhost forseti-cron: #011status = StatusCode.UNKNOWN
Jan  4 07:05:37 localhost forseti-cron: #011details = "Exception iterating responses: (sqlalchemy.exc.InvalidRequestError) Can't reconnect until invalid transaction is rolled back [SQL: u'SELECT cai_temporary_store.name AS cai_temporary_store_name, cai_temporary_store.parent_name AS cai_temporary_store_parent_name, cai_temporary_store.content_type AS cai_temporary_store_content_type, cai_temporary_store.asset_type AS cai_temporary_store_asset_type, cai_temporary_store.asset_data AS cai_temporary_store_asset_data \nFROM cai_temporary_store \nWHERE cai_temporary_store.content_type = %s AND cai_temporary_store.asset_type = %s AND cai_temporary_store.name = %s'] [parameters: [{}]]"
[...]

PR #2496, opened 2019-01-16
Rolling back the session on sqlalchemyerror when deleting CAI data

joecheuk:
Relates #2467


google/cloud/forseti/services/inventory/storage.py
```diff
@@ -758,6 +758,7 @@ def clear_cai_data(session):
         except SQLAlchemyError as e:
             LOGGER.exception('Attempt to delete data from CAI temporary store '
                              'failed, disabling the use of CAI: %s', e)
+            session.rollback()
```
blueandgold: Thanks for this... is this the only place where the rollback is needed?  Are there others?

PR #2497, opened 2019-01-17
Assign Project IAM roles immediately after creating server and client service accounts

hshin-g:
PR to address Issue #2415.

Assign the following Project IAM roles immediately after creating Forseti Server and Client service accounts:

storage.objectViewer
storage.objectCreator
cloudsql.client
logging.logWriter


install/gcp/installer/forseti_server_installer.py
```diff
@@ -77,6 +78,15 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
             bool: Whether or not the deployment was successful.
             str: Deployment name.
         """
+        self.has_roles_script = gcloud.grant_server_svc_acct_project_roles(
```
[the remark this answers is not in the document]
hshin-g: This will not work because the bucket name (generated during deploy) is required for some of the permissions, and will fail.

install/gcp/installer/forseti_server_installer.py
```diff
@@ -77,6 +78,15 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
...
+            self.target_id,
+            self.project_id,
+            self.gcp_service_acct_email,
+            self.user_can_grant_roles)
+
+        # Sleep for 10s to allow service acct permissions to take hold.
```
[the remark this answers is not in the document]
hshin-g: Done.

install/gcp/installer/forseti_server_installer.py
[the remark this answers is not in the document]
blueandgold: Per offline discussion, that assinment of objectAdmin role to the CAI bucket is not needed.  That can be removed, and then this should work.

install/gcp/installer/forseti_server_installer.py
```diff
@@ -77,6 +77,14 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
...
             str: Deployment name.
         """
+        self.has_roles_script = gcloud.grant_server_svc_acct_roles(
+            self.enable_write_access,
+            self.access_target,
+            self.target_id,
+            self.project_id,
+            self.gcp_service_acct_email,
+            self.user_can_grant_roles)
+
```
blueandgold: Will you still be adding the time out here?  I think we still need it to mitigate the race condition.  With the previous comment would be awesome.

install/gcp/installer/forseti_server_installer.py
```diff
@@ -77,6 +77,14 @@ def deploy(self, deployment_tpl_path, conf_file_path, bucket_name):
             bool: Whether or not the deployment was successful.
             str: Deployment name.
         """
+        self.has_roles_script = gcloud.grant_server_svc_acct_roles(
```
blueandgold: For consistency, can you please make the same change in the `forseti_client_installer.py` as well?  (include the timeout)

install/gcp/installer/forseti_server_installer.py
hshin-g: Added in.

install/gcp/installer/forseti_server_installer.py
hshin-g: Added in.

hshin-g:
@blueandgold @joecheuk Please review- @knitterb has tested this out on his end with non-org access and it has been approved on his end.

Issue #2499, opened 2019-01-18
Remove storage.objectAdmin role on CAI bucket


joecheuk:
Remove storage.objectAdmin role on CAI bucket since it's not required.

We already have storage.objectViewer and storage.objectCreator roles on the project level and are using the bucket lifecycle management to handle object deletion therefore the extra permission (delete) from the objectAdmin role is not needed.

PR #2501, opened 2019-01-18
Take out references to storage.objectAdmin in installer.

hshin-g:
References Issue #2499.

Take out references to storage.objectAdmin in installer as it is not needed.


hshin-g:
@joecheuk - Ran forseti inventory create, created model,  ran scanner, ran notifier and they are all working. Checked the permissions of the service account against the service account permissions that are generated in the normal 2.10.0 install and they sync up. Looks good to go.

Issue #2514, opened 2019-01-24
gcp_install.py fails when running with --dry-run


maurofaccenda:
Forseti version: v2.10.0
Modules: installer
Error message:
```
Creating GCP Service Account... This is a dry run, so don't actually create the service account.
	forseti-server-gcp-[redacted]
This is a dry run, so skipping this step.
Traceback (most recent call last):
  File "install/gcp_installer.py", line 23, in <module>
    gcp_installer.run()
  File "[redacted]forseti-security/install/gcp/gcp_installer.py", line 139, in run
    forseti_setup.run_setup()
  File "[redacted]forseti-security/install/gcp/installer/forseti_installer.py", line 139, in run_setup
    bucket_name)
  File "[redacted]/forseti-security/install/gcp/installer/forseti_server_installer.py", line 81, in deploy
    deployment_tpl_path, conf_file_path, bucket_name)
  File "[redacted]/forseti-security/install/gcp/installer/forseti_installer.py", line 252, in deploy
    message=loading_message)
  File "[redacted]/forseti-security/install/gcp/installer/util/utils.py", line 435, in start_loading
    if exit_condition_checker and exit_condition_checker():
  File "[redacted]/forseti-security/install/gcp/installer/forseti_installer.py", line 247, in <lambda>
    deployment_name, constants.DeploymentStatus.DONE))
  File "[redacted]/forseti-security/install/gcp/installer/util/gcloud.py", line 955, in check_deployment_status
    deployment_name, '--format=json'])
  File "[redacted]/forseti-security/install/gcp/installer/util/utils.py", line 378, in run_command
    stderr=subprocess.PIPE)
  File "/usr/lib64/python2.7/subprocess.py", line 394, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.7/subprocess.py", line 1047, in _execute_child
    raise child_exception
TypeError: execv() arg 2 must contain only strings
```
Command executed: `python install/gcp_installer.py --type=server --no-cloudshell --dry-run`

I went through the code and I saw on: https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/install/gcp/installer/util/gcloud.py#L851

That when runing as a dry run, it returns an integer when the function is meant to return an string. That seems to be the root cause of the issue.

blueandgold:
Thank you for reporting this.  @maurofaccenda Would this be something that you can contribute a fix?  If not, we can add it to our work schedule, though it would be faster if you can help.  :)

[...]

PR #2516, opened 2019-01-25
Remove untested/unsupported flags (dry-run & advanced)

joecheuk:
Flags --dry-run and --advanced were never tested, will remove them for now and add them back in once we have decided to support them in the future.

Relates #1843, #2514


joecheuk:
Still need to do some more testing to verify everything is working.

PR #2517, opened 2019-01-25
Updated rules to include comment to clarify resource_ids.

### LINKED ISSUE #2063: Folder and Project resources inconsistently use resource id or resource number
v2.4.0
I was trying to test that the log sink scanner actually triggered violations.  It turns out that if the resource is a **folder** it needs to be the **folder number**, BUT if the resource is a **project**, the resource id needs to be a **project id**.

```
rules:
  - name: 'blah'
    mode: required
    resource:
      - type: project
        applies_to: self
        resource_ids: ['**PROJECT_ID**']
    sink:
      destination: 'WRONG_DESTINATION'
      filter: ''
      # required, but doesn't mean anything for projects
      # https://cloud.google.com/sdk/gcloud/reference/beta/logging/sinks/create
      include_children: '*'
  - name: 'blah'
    mode: required
    resource:
      - type: folder
        applies_to: self
        resource_ids: ['**FOLDER_NUMBER**']
    sink:
      destination: 'WRONG_DESTINATION'
      filter: ''
      include_children: '*'
```

When I print out the `parent` object from https://github.com/GoogleCloudPlatform/forseti-security/blob/3b01742a995a73350021bfff69fa5cd7b0136ea2/google/cloud/forseti/scanner/scanners/log_sink_scanner.py#L132, I see that if the parent is a **folder**, `parent.name` is the **folder number** but if the parent is a **project**, `parent.name` is the **project id**.

My other rules use the wildcard with project so I hadn't run into this problem before, but this suggests that this is a problem in the model interface.

EDIT: It turns out the Folder ID is a number too, so when I mention Folder Number I mean Folder ID

joecheuk:
Resolves #2063

### DIFF
--- a/rules/audit_logging_rules.yaml
+++ b/rules/audit_logging_rules.yaml
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 # rules:
-#   # Note: Please use organization id, folder number and project id
-#   # when specifying resource ids.
+#   # Note: please use the id of the resource (such as organization id,
+#   # folder id, project id, etc.) when specifying the resource ids.
 #   # example rules
 #   - name: 'Require AUDIT_READ on all services, with exemptions.'
 #     resource:
--- a/rules/audit_logging_rules.yaml
+++ b/rules/audit_logging_rules.yaml
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 # rules:
+#   # Note: Please use organization id, folder number and project id
+#   # when specifying resource ids.
 #   # example rules
 #   - name: 'Require AUDIT_READ on all services, with exemptions.'
 #     resource:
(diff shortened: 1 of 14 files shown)


joecheuk:
Thanks for reviewing @hshin-g , will address the comment once @blueandgold finished reviewing since it will involve a lot of copying and pasting in many files and would like to change them once in for all.

rules/audit_logging_rules.yaml
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 # rules:
+#   # Note: Please use organization id, folder number and project id
```
blueandgold: Technically, there is no folder number.... so this will have to be `folder id`.
How about this:
```
# Note: please use the id of the resource (such as organization id, folder id,
# project id, etc.) when specifying the resource ids.
```
I think the confusion might be when people use `name`, instead of `id`.

PR #2524, opened 2019-02-05
k8s Services Cont

Red-Five:
Do not merge yet. Placeholder for next set of k8s changes. 

Fix issue with grpc communication on k8s


Red-Five:
@blueandgold feel free to merge this into docker_poc branch to fix the cli connectivity issue.

At this time docker_poc has a basic k8s cron approach and a long running server approach.

At some point, grpc load balancing needs special handling if/when start to refactor into micro services, but for now as we are only running one forseti server it doesn't matter.

Red-Five:
CLI connectivity fixed.
Main item remaining is to decide on and configure the cron job approach in the long running server deployment. We could just use the same crontab approach that the VM approach uses. Docker image needs crontab installed. Schedule can be picked up and set by docker_entrypoint.sh

blueandgold:
Thanks for figuring out the client connectivity to the server -- that's awesome!!

Crontab approach sounds good to me, since that replicates the current functionality.  If we ever have to change to say, cronjob on k8s, then we can iterate on that.

PR #2534, opened 2019-02-07
Switch to using firewalls.patch for firewall enforcer.

ahoying:
The compute.firewalls.update API does not properly handle EGRESS rules and other new firewall fields.

### DIFF
--- a/google/cloud/forseti/common/gcp_api/repository_mixins.py
+++ b/google/cloud/forseti/common/gcp_api/repository_mixins.py
@@ -403,19 +403,19 @@ class InsertResourceMixin(object):
         return self.execute_command(verb=verb, verb_arguments=arguments)
 
 
-class UpdateResourceMixin(object):
-    """Mixin that implements the Update API command for resources."""
+class _ModifyResourceBaseMixin(object):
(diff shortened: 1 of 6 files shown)


google/cloud/forseti/common/gcp_api/repository_mixins.py
```diff
@@ -447,14 +447,67 @@ def update(self, resource, data, target=None, verb='update', **kwargs):
...
             resource_link = self._build_resource_link(**arguments)
             return _create_fake_operation(resource_link, verb, target)
 
         return self.execute_command(verb=verb, verb_arguments=arguments)
 
 
+class PatchResourceMixin(_ModifyResourceBaseMixin):
+    """Mixin that implements the Patch API command for resources."""
+
+    def patch(self, resource, data, target=None, verb='patch', **kwargs):
```
joecheuk: if this is for patch only, why is verb an argument instead of a fixed value? (same for update)

ahoying: Good question, the main reason is that some APIs may use a different verb for the same meaning, so this gives flexibility when using the mixin.

PR #2543, opened 2019-02-11
Inventory's status should be PARTIAL SUCCESS if warnings are found

red2k18:


google/cloud/forseti/services/inventory/storage.py
```diff
@@ -129,7 +129,10 @@ def complete(self, status=IndexState.SUCCESS):
             status (str): Final inventory_status.
         """
         self.completed_at_datetime = date_time.get_utc_now_datetime()
-        self.inventory_status = status
+        if self.inventory_index_warnings:
+            self.inventory_status = IndexState.PARTIAL_SUCCESS
```
joecheuk: You should set the status when you are calling the complete method instead of updating the status inside the method

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -1172,11 +1172,14 @@ def rollback(self):
 
     def commit(self):
         """Commit the stored inventory."""
-
+        if InventoryIndex.inventory_index_warnings:
```
joecheuk: Why are you referencing the class instead of the existing object?

Issue #2443, opened 2018-12-22
Findings Sent to Cloud SCC Should Use the Update Method.


blueandgold:
Findings are currently sent to Cloud SCC using the [`CreateFinding` method](https://github.com/GoogleCloudPlatform/forseti-security/blob/075821ebe7ff8cb7fafa90b7b83cdf0bdd3843a3/google/cloud/forseti/common/gcp_api/securitycenter.py#L124-L148).

Cloud SCC team has suggested that we change to using the `UpdateFinding` method, which also will create findings if they don't exist. Using `update` also has the benefit of modifying the finding.

antoine-castex:
When a finding is sent to CSCC , and if this finding is fixed , when did he disapear from CSCC ?? because 2 weeks after and still here for me ...

red2k18:
Hi @antoine-castex, thanks for bringing this up! Have you tried resolving the finding from the CSCC GUI itself?

katze120:
I have the same issue with Forseti findings too, and am not able to find any way to close the findings from CSCC GUI. Only workaround I found so far is to add security marks to hide the findings by filtering. 
FYI - Findings from other sources e.g. Cloud Security Scanner will disappear from CSCC automatically if the finding is not there at subsequent scan; they also have "eventTime" updated if they persists.

red2k18:
Thanks all for reporting this issue! Below is the solution I came up with. Please let me know your thoughts on this so we can fix this asap.

Create an update method in Forseti that will compare the results of the latest notifier run with the results of the previous notifier run that is displayed on CSCC dashboard, and update the findings in CSCC accordingly. This will ensure the violations displayed in CSCC are consistent with the results of the latest notifier run.

@MartenvanWezel eager to hear your thoughts on this.

red2k18:
Thanks Marten for your kind response! I will work on creating the implementation plan then.

red2k18:
This issues is fixed in #2618 and is part of v2.13.0 that is slated for release this week.

PR #2553, opened 2019-02-13
patch instead of create cscc findings

katze120:
Use "[patch](https://cloud.google.com/security-command-center/docs/reference/rest/v1beta1/organizations.sources.findings/patch)" instead of "create" method to create or update CSCC findings.

Benefits:
1. For persistent findings, this will update "eventTime" on CSCC
2. If user marks a forseti finding as "INACTIVE" and the finding still exists during next Forseti run, "patch" method can update the same finding back to "ACTIVE" state with a new eventTime. 
This is one attempt towards resolving #2443 . While it does not yet remove out-of-date findings, it now makes finding removal outside of Forseti  (i.e. programmatically setting the finding to INACTIVE state) compatible with subsequent Forseti runs. 
If we continue to use "create" method, incorrectly inactivated findings by users cannot be reactivated by Forseti. 

---


google/cloud/forseti/common/gcp_api/repository_mixins.py
```diff
@@ -358,6 +358,27 @@ def create(self, verb='create', **kwargs):
         return self.execute_query(verb=verb, verb_arguments=arguments)
 
 
+class PatchQueryMixin(object):
```
ahoying: There is an existing PatchResourceMixin class you can use, please switch to that.

google/cloud/forseti/common/gcp_api/securitycenter.py
```diff
@@ -139,11 +140,10 @@ def create_finding(self, finding, organization_id=None, source_id=None,
             # beta api
             try:
                 LOGGER.debug('Creating finding with beta api.')
-                response = self.repository.findings.create(
+                response = self.repository.findings.patch(
```
ahoying: In line 102 above, you need to pass an additional parameter to init: key_field='name'

Then update the call to patch like this:

response = self.repository.findings.patch(
    '{}/findings/{}'.format(source_id, finding_id),
    findings)

katze120: thanks for the quick review and feedback!

google/cloud/forseti/common/gcp_api/repository_mixins.py
katze120: thanks!

PR #2573, opened 2019-02-15
KMS scanner now scans for enabled keys only

red2k18:


google/cloud/forseti/scanner/scanners/kms_scanner.py
```diff
@@ -114,18 +116,21 @@ def _retrieve(self):
...
                         'got %s, want kms_keyring' % key.parent_type_name
                     )
 
-                keys.append(crypto_key.CryptoKey.from_json(
-                    key.name,
-                    key.full_name,
-                    key.parent_type_name,
-                    key.type,
-                    key.data))
+                if state == 'ENABLED':
```
joecheuk: Can you add a comment here, saying why we want to do the comparison?

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -384,9 +383,9 @@ def find_violations(self, key):
         """
         violations = []
         creation_time = key.primary_version.get('createTime')
+        state = key.primary_version.get('state')
         scan_time = date_time.get_utc_now_datetime()
-
-        if self.rule['mode'] == BLACKLIST:
+        if self.rule['mode'] == BLACKLIST and state == 'ENABLED':
```
joecheuk: Since we might plan to support other modes in the future, can we separate the 2 checks in this if statement?

e.g. put the following under "state = key.primary_version.get('state')"

if state != 'ENABLED':
  // We don't scan for keys that are not currently enabled.
  return []

tests/scanner/scanners/kms_scanner_test.py
```diff
@@ -82,6 +82,9 @@ def setUpClass(cls):
 
             crypto_key.data = fake_kms_scanner_data.NON_ROTATED_CRYPTO_KEY_DATA
 
+            crypto_key.data = (
```
joecheuk: why do we set the crypto_key.data twice? (here and two lines above)

tests/scanner/test_data/fake_kms_scanner_data.py
```diff
@@ -51,6 +51,24 @@
                  '"algorithm":"GOOGLE_SYMMETRIC_ENCRYPTION",'
                  '"protectionLevel":"SOFTWARE"}}')
 
+NON_ROTATED_CRYPTO_KEY_STATE_DATA = ('{"createTime":"2019-01-22T23:30:18.939244464Z",'
```
joecheuk: The naming should be NON_ROTATED_CRYPTO_KEY_DESTROYED_DATA instead

PR #2584, opened 2019-02-20
Added 4 use cases to the initial KMS Scanner

red2k18:


google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
         self.rule_index = rule_index
         self.rule = rule
 
-    def is_key_rotated(self, creation_time, scan_time):
-        """Check if the key has been rotated within the time speciifed in the
-         policy.
+    def find_match_rotation_period(self, key, rotation_period, mode):
```
joecheuk: Using match in the method name is unclear in this case, when comparing rotation_period, it's looking for either greater than/equals to or less than/equals to and not looking for exact match.

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
...
-        for key_data in crypto_key:
-            key_rotation_period = key_data.get('rotation_period')
         LOGGER.debug('Formatting rotation time...')
+        creation_time = key.primary_version.get('createTime')
+        scan_time = date_time.get_utc_now_datetime()
         last_rotation_time = creation_time[:-5]
         formatted_last_rotation_time = datetime.datetime.strptime(
             last_rotation_time, string_formats.TIMESTAMP_MICROS)
         days_since_rotated = (scan_time - formatted_last_rotation_time).days
+        print('days:', days_since_rotated)
```
joecheuk: remove print statement

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
...
-            the key was created.
-            scan_time (datetime): Snapshot timestamp.
+            key (Resource): The resource to check for a match.
+            mode (string): The mode specified in the rule.
+            rotation_period (string): The cut off rotation schedule of crypto
+            key specified in rule file.
 
         Returns:
-            bool: Returns true if key was rotated within the time specified.
+            bool: Returns true if a match is found.
```
joecheuk: is a match a violation ?

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
...
+        Args:
+            key (Resource): The resource to check for a match.
+            rule_algorithms (string): The algorithms of this rule.
 
-        if days_since_rotated > key_rotation_period:
-            return False
+        Returns:
+            bool: Returns true if a match is found.
+        """
+        LOGGER.debug('Finding rule algorithm in key algorithm')
```
joecheuk: Log message is not clear about what it's doing

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
...
+        key_algorithm = key.primary_version.get('algorithm')
+        for algorithm in rule_algorithms:
+            if key_algorithm == algorithm:
+                return True
+        return False
+
+    def find_match_protection_level(self, key, rule_protection_level):
```
joecheuk: Should rule_protection_level be a list, like algorithms?

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
...
+        Returns:
+            bool: Returns true if a match is found.
+        """
+        key_protection_level = key.primary_version.get('protectionLevel')
+        if key_protection_level == rule_protection_level:
+            return True
+        return False
 
-        return True
+    def find_match_purpose(self, key, rule_purpose):
```
joecheuk: Should rule_purpose be a list, like algorithms?

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
...
+        Args:
+            key (Resource): The resource to check for a match.
+            rule_purpose (string): The purpose of this rule.
+
+        Returns:
+            bool: Returns true if a match is found.
+        """
+        key_purpose = key.purpose
+        symmetric_algorithm = 'ENCRYPT_DECRYPT'
+        if rule_purpose == 'symmetric':
```
joecheuk: Do this instead of nesting the if statements:

if rule_purpose == 'symmetric' and key_purpose == symmetric_algorithm:

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
...
+            if key_purpose == symmetric_algorithm:
+                return True
+        elif not key_purpose == symmetric_algorithm:
```
joecheuk: I dont understand this part, why arent you matching key_purpose with rule_purpose, instead you are putting in custom strings like 'ENCRYPT_DECRYPT' or 'symmetric' in here

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -346,34 +338,107 @@ def __init__(self, rule_name, rule_index, rule):
...
+
+        Args:
+            key (Resource): The resource to check for a match.
+            rule_state (string): The state of this rule.
+
+        Returns:
+            bool: Returns true if a match is found.
+        """
+        key_state = key.primary_version.get('state')
+        for state in rule_state:
```
joecheuk: is rule_state a string or a list?

google/cloud/forseti/scanner/audit/kms_rules_engine.py
```diff
@@ -382,30 +447,61 @@ def find_violations(self, key):
...
-                    rotation_period=key.rotation_period,
-                    violation_reason=violation_reason,
-                    key_creation_time=key.create_time,
-                    version_creation_time=creation_time,
-                    resource_data=key.data))
+
+        mode = self.rule['mode']
+        has_violation = False
+
+        crypto_key = self.rule['key']
```
joecheuk: crypto_key can be confusing here, crypto_key_rule might be better

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: The use case for this method is to check if a key is symmetric or asymmetric. The valid values of `purpose` in rule is `symmetric` and `asymmetric`. The only purpose allowed for symmetric key encryption is `ENCRYPT_DECRYPT`. So I'm checking against that.
https://cloud.google.com/kms/docs/algorithms

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: It is a list. Corrected.

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: Ah, ok. Renamed.

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: Done

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: Valid values for rule_purpose is `symmetric` or `asymmetric`. Hence it is not a list. Read more about this use case in comments below.

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: https://cloud.google.com/kms/docs/reference/rest/v1/ProtectionLevel

Technically there are only 2 valid values, either of which can be used at a time. Hence it is not a list.

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: Done

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: Rephrased.

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: If the mode is `blacklist` and match is found or if the mode is `whitelist` and there's no match then it is a violation.

google/cloud/forseti/scanner/audit/kms_rules_engine.py
red2k18: It's looking if match is found or not based on the mode and `days_since_rotated`. We want to throw violations only if `days_since_rotated  > rotation_period` and the method is doing exactly that.

PR #2604, opened 2019-02-28
Fix missing IAP violations

blueandgold:
The `group` field in `backend services` seems to have changed to `resourceGroup`.

I checked the documentation and it doesn't seem to be consistent with the actual data from CAI.

https://cloud.google.com/compute/docs/reference/rest/v1/backendServices

Is there some other translation needed?

@red2k18 @hshin-g

### DIFF
--- a/google/cloud/forseti/scanner/scanners/iap_scanner.py
+++ b/google/cloud/forseti/scanner/scanners/iap_scanner.py
@@ -291,8 +291,11 @@ class _RunData(object):
         alternate_services = set()
         direct_access_sources = set()
         for backend in backend_service.backends:
+            print '-----'
+            print backend
+            print '-----'
             instance_group = self.find_instance_group_by_url(
-                backend.get('group'))
+                backend.get('resourceGroup'))
             if not instance_group:
                 continue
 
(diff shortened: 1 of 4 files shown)


ahoying:
> Does this work with CAI turned off?

Right, this change needs to be made here: https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/services/inventory/base/cai_gcp_client.py#L303. If the key we get from CAI is different than the key defined in the public documentation and returned by calling the API directly.

google/cloud/forseti/scanner/scanners/iap_scanner.py
```diff
@@ -292,7 +292,7 @@ def make_iap_resource(self, backend_service, project_full_name):
         direct_access_sources = set()
         for backend in backend_service.backends:
             instance_group = self.find_instance_group_by_url(
-                backend.get('group'))
+                backend.get('resourceGroup'))
```
ahoying: Please make the change in the CAI inventory, rather than changing the scanner.

joecheuk: Updated, PTAL!

blueandgold: Thanks for helping Joe!

Issue #2613, opened 2019-03-01
Error running resource scanner with folders


dufton:
Resource scanner will fail if the scanned hierarchy contains a folder.

Relevant stack trace:
```
Error running scanner: ResourceScanner
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.11.1-py2.7.egg/google/cloud/forseti/scanner/scanner.py", line 117, in run
    scanner.run()
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.11.1-py2.7.egg/google/cloud/forseti/scanner/scanners/resource_scanner.py", line 56, in run
    resources = self._retrieve()
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.11.1-py2.7.egg/google/cloud/forseti/scanner/scanners/resource_scanner.py", line 79, in _retrieve
    resource_util.create_resource_from_db_row(resource)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.11.1-py2.7.egg/google/cloud/forseti/common/gcp_type/resource_util.py", line 127, in create_resource_from_db_row
    create_resource_from_db_row(row.parent) if row.parent else None)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.11.1-py2.7.egg/google/cloud/forseti/common/gcp_type/resource_util.py", line 129, in create_resource_from_db_row
    return create_resource_from_json(row.type, parent, row.data)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.11.1-py2.7.egg/google/cloud/forseti/common/gcp_type/resource_util.py", line 150, in create_resource_from_json
    return resource_type.get('class').from_json(parent, json_string)
AttributeError: type object 'Folder' has no attribute 'from_json'"
```

joecheuk:
Resolved #2615

PR #2615, opened 2019-03-01
Add from_json method for BillingAccount and Folder

dufton:
Add support for BillingAccount and Folder to the resource_utils.create_resource_from_json method.

This should fix the resource scanner: #2613

### DIFF
--- a/google/cloud/forseti/common/gcp_type/folder.py
+++ b/google/cloud/forseti/common/gcp_type/folder.py
@@ -73,7 +73,9 @@ class Folder(resource.Resource):
         folder_dict = json.loads(json_string)
         name = folder_dict['name']
         folder_id = name.split('/')[-1]
-        full_name = '{}folder/{}/'.format(parent.full_name, folder_id)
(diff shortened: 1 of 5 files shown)


google/cloud/forseti/common/gcp_type/folder.py
```diff
@@ -56,3 +58,29 @@ def __init__(
...
+            json_string (str): JSON string of a folder GCP resource.
+
+        Returns:
+            Folder: folder resource.
+        """
+        folder_dict = json.loads(json_string)
+        name = folder_dict['name']
+        folder_id = name.split('/')[-1]
+        full_name = '{}folder/{}/'.format(parent.full_name, folder_id)
+        lifecycle = folder_dict.get('lifecycleState',
```
ahoying: Depending on the inventory, a folder may not have a parent, please do a check like https://github.com/GoogleCloudPlatform/forseti-security/blob/dev/google/cloud/forseti/common/gcp_type/project.py#L86.

And add an appropriate test.

PR #2623, opened 2019-03-07
Updated most of the explainer methods to use stream

### LINKED ISSUE #2620: Update explainer commands to use stream to avoid return payload size > message size limit
e.g. the returned value of access_by_member is not being streamed, and will throw an error when the returned value is > 4 mb.

joecheuk:
Resolves #2620


joecheuk:
Hey @ahoying , updating the existing unit tests seem to be sufficient as the return values from the explainer client have changed from lists to generators which caused most of the unit tests to failed.

PR #2640, opened 2019-03-15
Fixed Pylint errors

red2k18:


google/cloud/forseti/common/gcp_api/securitycenter.py
```diff
@@ -195,15 +195,12 @@ def list_findings(self, source_id):
         response = self.repository.findings.list(parent=source_id)
         return response
 
-    def update_finding(self, finding, finding_id, state, event_time,
-                       source_id=None):
+    # pylint: disable=inconsistent-return-statements
```
ahoying: You should have the return statement on line 213 return an empty dict and not override the pylint.

red2k18: Fixed, thanks.

PR #2657, opened 2019-03-22
Handle CSCC list API edge case

red2k18:


google/cloud/forseti/notifier/notifiers/cscc_notifier.py
```diff
@@ -223,14 +223,32 @@ def _send_findings_to_cscc(self, violations, source_id=None):
             formated_findings_in_page = (
```
blueandgold: It seems a small check here would work?
```
for page in paged_findings_in_cscc:
  if not page:
    continue
  <rest of your code>
```

red2k18: `page` isn't empty even when findings are not retrieved. So this wouldn't work.

red2k18:
#2667 fixes this issue.

PR #2680, opened 2019-03-29
Migrate to CSCC GA API in Forseti

red2k18:


google/cloud/forseti/notifier/notifiers/cscc_notifier.py
```diff
@@ -223,10 +223,12 @@ def _send_findings_to_cscc(self, violations, source_id=None):
             for page in paged_findings_in_cscc:
                 formated_findings_in_page = (
                     ast.literal_eval(json.dumps(page)))
-                findings_in_page = formated_findings_in_page.get('findings')
-                if not findings_in_page:
+                list_findings_result_paged = (
+                    formated_findings_in_page.get('listFindingsResult'))
+                if not list_findings_result_paged:
```
blueandgold: When this happens (`list_findings_result_paged` is `None`), it means we are getting the beta api response, right?  If so, you should not continue here.  You should process the response again, based on the beta api format.

red2k18: `list_findings_result_page` is None when findings aren't retrieved from CSCC. It is not related to beta api response.

blueandgold: Thanks, that's right.  We are migrating everyone to v1, and beta will not be used anymore.  Can you please update your PR's title to be "Migrate" instead of "Integrate" to be more clear?

Issue #2714, opened 2019-04-05
Research on: cloud_sql_proxy should use private IP's


marko7460:
cloud_sql_proxy should use private IP's to connect to Cloud SQL. This is a more secure way and more compliant way of connecting to database.

red2k18:
Can this be released next week?

hshin-g:
As per comment in #2715, closing relevant PRs until more investigation is done.

kevensen:
Per the [cloud_sql_proxy](https://cloud.google.com/sql/docs/mysql/sql-proxy#private-ip) docs, the ` --ip_address_types=PRIVATE` is used when a CloudSQL instance has BOTH a public and private IP.  The reason for this is without the switch, the cloud_sql_proxy will default to using the public IP.

That said, the open PR for this, https://github.com/forseti-security/terraform-google-forseti/pull/144, configures it as an either or scenario.  In this case, the `--ip_address_types=PRIVATE` flag is unnecessary, likely throwing an error.

I think https://github.com/forseti-security/terraform-google-forseti/pull/144 is a good approach.

PR #2715, opened 2019-04-05
setting cloud_sql_proxy to use private IP's to connect to mysql

marko7460:
Fix for https://github.com/GoogleCloudPlatform/forseti-security/issues/2714


install/gcp/scripts/initialize_forseti_services.sh
```diff
@@ -49,7 +49,7 @@ CONFIG_VALIDATOR_COMMAND+=" -port=50052"
 sudo chmod ugo+x ${FORSETI_HOME}/external-dependencies/config-validator/ConfigValidatorRPCServer
 
 SQL_PROXY_COMMAND="$(which cloud_sql_proxy)"
-SQL_PROXY_COMMAND+=" -instances=${SQL_INSTANCE_CONN_STRING}=tcp:${SQL_PORT}"
+SQL_PROXY_COMMAND+=" -instances=${SQL_INSTANCE_CONN_STRING}=tcp:${SQL_PORT} -ip_address_types=PRIVATE"
```
joecheuk: Based on the documentation, the correct way of specifying this option is to use double dashes "--ip_address_types=PRIVATE", can you add another dash to the option?

https://cloud.google.com/sql/docs/mysql/sql-proxy

marko7460: I have tested with `-` and `--` dashes. They both work. From `cloud_sql_proxy --help` i got this:
```
  -ip_address_types
    Default to be 'PUBLIC,PRIVATE'. Options: a list of strings separated by ',', e.g. 'PUBLIC,PRIVATE'
```

I have reported this discrepancy as in issue on Documentation link you provided above.

TODO:
1. I will change to `--` because this is still official documentation.
2. Deployment Templates need to be updated to set Cloud SQL with private IP. I think that at the moment CloudSQL is set only with Public IP.

joecheuk: Following up on your comment, would you mind also adding the changes to the deployment template to use private IP?

dekuhn:
FYI: We have announced a code freeze starting May 16, 2019, and ending May 30, 2019. During this time we won't be merging any new pull requests. For more details you can read the full announcement details [here](https://groups.google.com/a/forsetisecurity.org/forum/#!topic/announce/iIV-sr5zfds).

dekuhn:
This issue is blocked on getting CLA signed. Please let us know if you are having trouble there?

Anything else we can do to move this forward?

dekuhn:
My update of the issue caused the CLA check to run again so that is good.

There is one unresolved comment I think blocking us from getting this submitted.

install/gcp/scripts/initialize_forseti_services.sh
marko7460: i don't think that my fix is correct. `-ip_address_types` is necessary if cloudSQL is setup with PRIVATE+PUBLIC ip. If cloudSQL is setup with only PRIVATE or only PUBLIC ip than there is no need for `-ip_address_types`

hshin-g:
@aaron-lane @marko7460 Is this PR resolved/merged into #2834 ?

marko7460:
@hshin-g It looks like two PRs are the same but i don't think this PR will work. `-ip_address_types` is necessary if cloudSQL is setup with PRIVATE+PUBLIC ip. If cloudSQL is setup with only PRIVATE or only PUBLIC ip than there is no need for -ip_address_types.

If the cloudSQL is setup with only PRIVATE ip for example then `-ip_address_types` will throw an error.

I think we should close the PR. This need further investigation.

hshin-g:
@marko7460 Thanks for the information! I'll close both PRs for now, but will leave Github issue #2714 open. Would you have an ETA on when we would be able to resolve this?

PR #2720, opened 2019-04-06
Updated upgrade instructions to include adding GSuite groups settings scope

joecheuk:


joecheuk:
Resolves #2721

_docs/_latest/setup/upgrade.md
```diff
@@ -1308,10 +1308,15 @@ Example command: `gcloud compute instances reset forseti-server-vm-70ce82f --zon
...
-  - Update [KE scanner rule file](https://github.com/GoogleCloudPlatform/forseti-security/blob/v2.14.0/rules/ke_scanner_rules.yaml)
-    under `rules/` in your Forseti server GCS bucket to include sample rules according to CIS benchmark.
-  - Add [Groups settings rule file](https://github.com/GoogleCloudPlatform/forseti-security/blob/v2.14.0/rules/groups_settings_rules.yaml)
-    under `rules/` in your Forseti server GCS bucket to include Groups Settings rules.
+   - Update [KE scanner rule file](https://github.com/GoogleCloudPlatform/forseti-security/blob/v2.14.0/rules/ke_scanner_rules.yaml)
+     under `rules/` in your Forseti server GCS bucket to include sample rules according to CIS benchmark.
+   - Add [Groups settings rule file](https://github.com/GoogleCloudPlatform/forseti-security/blob/v2.14.0/rules/groups_settings_rules.yaml)
+     under `rules/` in your Forseti server GCS bucket to include Groups Settings rules.
+1. GSuite scope updates:
+   - Add GSuite scope `https://www.googleapis.com/auth/apps.groups.settings` to your Forseti server service
```
blueandgold: Is the enabling the Groups Setting API mentioned somewhere?

joecheuk: I dont see it on the gsuite.md page, is it supposed to be required?

blueandgold: Yes, the `Groups Settings` API does need to be enabled.

joecheuk: Can't seem to find ways to enable groups settings api, instead theres an enable api access step for all GSuite APIs.
https://developers.google.com/admin-sdk/groups-settings/prerequisites

blueandgold: So, I think there's some misunderstanding of the intent here.  We should just call out **specifically and simply** here in the upgrade instruction, that the `Groups Settings` API needs to be enabled.  That's it.  :)

Linking to the perquisite seems unnecessary, since for upgrading, all the prerequisites should have already been done.  So, we should not mention the prerequisites.

So, please do this:
1: remove the line that links to prerequisites.
2: add another high level list items
```
1. API update:
  - Enable the Groups Settings API
```

joecheuk: Updated

PR #2748, opened 2019-04-16
Updated SQLAlchemy version to 1.3.0

red2k18:


red2k18:
Fixes #2749

dekuhn:
Do we need to keep this PR open or can you close it. I believe you were just testing if we could cleanly just upgrade the package which say unit tests break?

PR #2751, opened 2019-04-18
Update location scanner doc to include supported resource types

red2k18:


_docs/_latest/configure/scanner/rules.md
```diff
@@ -1068,7 +1068,8 @@ rules:
 * `applies_to`
   * `type`
     * **Description**: The type of resource to apply the rule to.
-    * **Valid values**: Currently only supports `bucket`.
+    * **Valid values**: One of `bucket`, `cloud_sql_instance`, `dataset`,
```
joecheuk: I think it's cloudsqlinstance, kubernetes_cluster instead of cloud_sql_instance and ke_cluster, can you verify them?

red2k18: Verified and updated.

PR #2762, opened 2019-04-24
removed version for pylint-quotes

kevensen:
Setting version 0.2.0 for pylint-quotes proves somewhat problematic.  Version 0.2.0 requires python >3.4 per PyPi.  The setup script fails out if major version 3 is detected.

I recommend removing the version of pylint-quotes unless it is specifically needed.  In which case, I would set the required version of Python to 3.x


kevensen:
Disregard.  I was attempting to do a ```pip install``` not ```python setup.py install```

PR #2769, opened 2019-04-24
Updated config validator violation to include violation message properly.

### LINKED ISSUE #2735: Violation message is missing in the config validator violation notification

joecheuk:
Resolves #2735


joecheuk:
> Will this cause any upgrade or release issue? (LGTM)

This will not cause any upgrade issue as the column has already been created in the violations table.

PR #2797, opened 2019-05-07
Fix KMS Scanner test

red2k18:


tests/scanner/scanners/kms_scanner_test.py
```diff
@@ -224,7 +224,7 @@ def test_run_scanner_rotation_period_whitelist_match(self,
         self.scanner.run()
         crypto_key = self.scanner._retrieve()
         violations = self.scanner._find_violations(crypto_key)
-        self.assertEquals(1, len(violations))
+        self.assertEquals(6, len(violations))
```
osandoval011: @red2k18 Are we sure its correct to only have 1 now?

PR #2826, opened 2019-05-15
Add storage.objectViewer role to GCP_READ_IAM_ROLES

marmolejo:
GCP_READ_IAM_ROLES lacks permission storage.buckets.get to retrieve
bucket ACLs when running the inventory.


marmolejo:
This fixes #2827.

marmolejo:
According to this: https://cloud.google.com/storage/docs/access-control/iam-roles roles/storage.objectViewer is the role that has the minimum set of permissions including storage.objects.get

joecheuk:
Hi @marmolejo , thanks for creating the PR!

The `roles/storage.objectViewer` role gives view permission to all the objects inside a bucket but based on the error message in the issue above, it seems like we are more interested in `storage.buckets.get` instead of `storage.objects.get`, can you double check if that's the case?

marmolejo:
Yes, sorry you are right. The standard role listed in the documentation is roles/storage.legacyBucketReader which have: 
storage.buckets.get 
storage.objects.list

dekuhn:
FYI: We have announced a code freeze starting May 16, 2019, and ending May 30, 2019. During this time we won't be merging any new pull requests. For more details you can read the full announcement details [here](https://groups.google.com/a/forsetisecurity.org/forum/#!topic/announce/iIV-sr5zfds).

hshin-g:
@marmolejo Can you submit a corresponding PR for this in the Forseti Terraform module [here](https://github.com/forseti-security/terraform-google-forseti)?

kevensen:
@marmolejo Is this still a work in progress?  If so, please let us know by commenting here.

In order to maintain consistency in behavior between the Python installer ([soon to be deprecated](https://forsetisecurity.org/docs/latest/setup/install-with-python-installer.html#install-with-python-installer)) and Terraform, do you mind adding equivalent functionality to the [terraform-google-forseti](https://github.com/forseti-security/terraform-google-forseti) module?

PR #2836, opened 2019-05-17
Ignore 404 from get SA key api response.

joecheuk:
Resolves #2836

### DIFF
--- a/google/cloud/forseti/common/gcp_api/iam.py
+++ b/google/cloud/forseti/common/gcp_api/iam.py
@@ -449,7 +449,7 @@ class IAMClient(object):
                          name, key_type, flattened_results)
             return flattened_results
         except (errors.HttpError, HttpLib2Error) as e:
-            if isinstance(e, errors.HttpError) and e.resp.status == 404:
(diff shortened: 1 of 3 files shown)


google/cloud/forseti/common/gcp_api/iam.py
```diff
@@ -449,6 +449,9 @@ def get_service_account_keys(self, name, key_type=None):
                          name, key_type, flattened_results)
             return flattened_results
         except (errors.HttpError, HttpLib2Error) as e:
+            if isinstance(e, errors.HttpError) and str(e.resp.status) == '404':
```
ahoying: Do you have  copy of the specific response returned from the API? Can you use logic like https://github.com/forseti-security/forseti-security/blob/dev/google/cloud/forseti/common/gcp_api/compute.py#L36 instead to ensure that other causes of 404 response code doesn't also get dropped by the crawler?

joecheuk: Updated with a match on the error message to make sure it's SA not found.

PR #2845, opened 2019-05-22
Added iter method for k8 resources

red2k18:


red2k18:
Opened #2861 instead.

Issue #2843, opened 2019-05-22
Broken links based on search results on forsetisecurity.org web site.


dekuhn:
When clicking on search results on the forsetisecurity.org web site e.g. https://forsetisecurity.org/search/?q=diagram I get 404 error due to the link going to a specific version of the docs that have been pruned.

Error: Not Found
The requested URL /docs/v2.2/concepts/architecture.html was not found on this server.

dekuhn:
If I know where to go I can browse and get to the content: https://forsetisecurity.org/docs/latest/concepts/architecture.html. Not sure why the search results are indexing or returning the pruned content/links vs using "latest".

dekuhn:
I have requested via webmaster tools to reindex www.forsetisecurity.org - 
https://search.google.com/search-console/inspect?resource_id=https%3A%2F%2Fwww.forsetisecurity.org

Based on https://support.google.com/webmasters/answer/9012289#url_not_on_google re-indexing may take a week or two.

dekuhn:
In digging into this further I have found the sitemap.xml generated is not valid. The sitemap URLs should be the fully qualified url not relative paths.

Sitemap can be read, but has errors
Invalid URL
4739 instances
This is not a valid URL. Please correct it and resubmit.
Examples
URL:
/docs/latest/concepts/architecture.html
Line 4

dekuhn:
Sitemap PR submitted for 2.16 release. Still waiting for Webmaster Tools to reindex the site to address the broken search links. The Sitemap fix will also help in the future with broken links.

dekuhn:
The site has been re-indexed and PR to build proper sitemap is merged for release.. Resolving!

PR #2847, opened 2019-05-23
Update _config.yml to fix sitemap generation

dekuhn:
Adding url to properly generate sitemap.xml with jekyll-sitemap plugin.

Details can be found at https://www.rubydoc.info/gems/jekyll-sitemap


dekuhn:
Fixes #2843

PR #2859, opened 2019-05-28
Fixed blacklist scanner to use decode the request content and updated model handle to use hex instead of str.

joecheuk:


dekuhn:
Is this linked to a github reported issue?

joecheuk:
@dekuhn this is related to py3 upgrades, not a real issue in py2

Issue #2825, opened 2019-05-15
key lifecycleState might not exist when crawling resources


joecheuk:
"[forseti-security][2.15.0] google.cloud.forseti.services.inventory.crawler(visit): 'key: lifecycleState, data: {}'
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.15.0-py2.7.egg/google/cloud/forseti/services/inventory/crawler.py", line 114, in visit
    resource.get_iam_policy(self.get_client())
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.15.0-py2.7.egg/google/cloud/forseti/services/inventory/base/resources.py", line 107, in wrapper
    result = f(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.15.0-py2.7.egg/google/cloud/forseti/services/inventory/base/resources.py", line 741, in get_iam_policy
    if self.enumerable():
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.15.0-py2.7.egg/google/cloud/forseti/services/inventory/base/resources.py", line 811, in enumerable
    return self['lifecycleState'] == 'ACTIVE'
  File "/usr/local/lib/python2.7/dist-packages/forseti_security-2.15.0-py2.7.egg/google/cloud/forseti/services/inventory/base/resources.py", line 201, in __getitem__
    raise KeyError('key: {}, data: {}'.format(key, self._data))
KeyError: 'key: lifecycleState, data: {}'"

dekuhn:
@joecheuk can we add a target Milestone for this Issue & PR.

PR #2879, opened 2019-05-31
Clarified in the config file to use project number / folder number / organization number when specifying the root resource id.

joecheuk:


joecheuk:
Resolves #2825

PR #2887, opened 2019-06-07
Add 'Contact Us' to top menu bar

red2k18:


_includes/site/layout/base/nav.html
```diff
@@ -28,6 +28,7 @@
         <li><a href="{{ site.baseurl }}/community/">Community</a></li>
         <li><a href="{{ site.baseurl }}/releases/">Releases</a></li>
         <li><a href="{{ site.baseurl }}/news/">News</a></li>
+        <li><a href="{{ site.baseurl }}/contact-us/">Contact Us</a></li>
```
blueandgold: We tried a few things, and we should be using the format below, to link to a specific page, whereas the above only works for directory-level.

```
<li><a href="{% link _docs/latest/use/get-help.md %}">Contact Us</a></li>
```

red2k18: Will submit a PR with this change.

PR #2896, opened 2019-06-10
Merging forsetisecurity.org-dev to forsetisecurity.org

red2k18:


contact-us/index.md
```diff
@@ -0,0 +1,7 @@
+---
+layout: default
+title: Contact Us
```
blueandgold: We should not have a page that just redirects to another page.

red2k18: Having `Contact Us` redirect to new page leads to inconsistency. Should the new page display data from `get help` page instead of having a link that redirects to `get help` page?

PR #2912, opened 2019-06-17
dont validate storage-component api when getting Bucket resources from CAI data

### LINKED ISSUE #2897: Inventory depends on certain google apis to be enabled while CAI can support those resources w/o enabling those APIs.
e.g. to inventory GCS bucket, we first check and see if storage API is enabled :-

```
def storage_api_enabled(self):
"""whether storage api is enabled.

Returns:
bool: if this API service is enabled on the project.
"""
return self.is_api_enabled('storage-component.googleapis.com')
```

But if CAI is enabled, storage bucket info is coming from CAI instead, we need to update to make sure for resources that we support w/ CAI doesn't depend only on the enablement of google api for that resource.

katze120:

Fixes #2897


google/cloud/forseti/services/inventory/base/resources.py
```diff
@@ -2279,8 +2279,7 @@ class SpannerInstanceIterator(resource_iter_class_factory(
 class StorageBucketIterator(resource_iter_class_factory(
         api_method_name='iter_storage_buckets',
         resource_name='storage_bucket',
-        api_method_arg_key='projectNumber',
-        resource_validation_method_name='storage_api_enabled')):
+        api_method_arg_key='projectNumber')):
```
joecheuk: Can we instead update the storage_api_enabled() method to something along the line of

storage_enabled() := true if storage api or CAI api enabled, false otherwise?

By removing this I am afraid users might get a lot of 403 if CAI is not enabled

katze120: I included thoughts around this in https://github.com/forseti-security/forseti-security/issues/2897#issuecomment-501501497 actually. 

In summary, my opinion is that we dont need to check any api in this StorageBucketIterator function because:
- This api validation in this function is done on the target project being scanned. (if we need CAI, we only need it on the Forseti project)
- This function eventually calls iter_cai_assets which works on data already in cai_temporary_store table and does not make any API calls

PR #2918, opened 2019-06-18
Findings updated correctly and sent to CSCC

red2k18:


red2k18:
Merged #2915 instead. Closing this.

Issue #2949, opened 2019-06-26
Update ke_rules.yaml to scan for the following ke versions


joecheuk:
Vulnerabilities to mitigate:
CVE-2019-11477
CVE-2019-11478
CVE-2019-11479

The patch will be available in Kubernetes 1.11.8-gke.10, Kubernetes 1.12.7-gke.24, Kubernetes 1.12.8-gke.10 and Kubernetes 1.13.6-gke.13.

PR #2953, opened 2019-06-28
Update ke_rules.yaml to scan for the added KE versions

hshin-g:
Resolves Issue #2949

### DIFF
--- a/rules/ke_rules.yaml
+++ b/rules/ke_rules.yaml
@@ -87,7 +87,17 @@ rules:
         minor: '5-gke.4'
         operator: '>='
       - major: '1.11'
-        operator: '>='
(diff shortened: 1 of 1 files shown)


rules/ke_rules.yaml
```diff
@@ -87,7 +87,17 @@ rules:
...
+        operator: '='
+      - major: '1.12'
+        minor: '7-gke.24'
+        operator: '='
+      - major: '1.12'
+        minor: '8-gke.10'
+        operator: '='
+      - major: '1.13'
+        minor: '6-gke.13'
+        operator: '='
```
joecheuk: You can use >= for 1.13

PR #2966, opened 2019-07-09
Generated 2.18 version and added upgrade instructions

joecheuk:


paskal:
[Site](https://github.com/forseti-security/forseti-security/tree/forsetisecurity.org) is not updated yet by the way. Also, documentation URL with **[v2.16](https://forsetisecurity.org/docs/v2.16/)** and **[latest](https://forsetisecurity.org/docs/latest/)** works, but **[v2.17](https://forsetisecurity.org/docs/v2.17/)** (and **[v2.18](https://forsetisecurity.org/docs/v2.18/)** of course) were not generated: maybe this can should be fixed before updating latest?

PR #2968, opened 2019-07-09
Merge forsetisecurity.org to forsetisecurity.org-dev

joecheuk:


paskal:
Re-posting from other (less relevant) issue:
Documentation URL with **[v2.16](https://forsetisecurity.org/docs/v2.16/)** and **[latest](https://forsetisecurity.org/docs/latest/)** works, but **[v2.17](https://forsetisecurity.org/docs/v2.17/)** (and **[v2.18](https://forsetisecurity.org/docs/v2.18/)** of course) were not generated: maybe this should be fixed before updating latest?

joecheuk:
Hi @paskal , thanks for updating the PR! We noticed the same problem and in the previous PR, not only 2.18 is generated but the missing 2.17 is also being generated so once the new version is released, we will have everything in place.

PR #2970, opened 2019-07-10
Merge forsetisecurity.org-dev into forsetisecurity.org

joecheuk:


joecheuk:
Manually updating the CLA due to old merge commits being squashed and the history is diff between prod and dev again.

PR #2985, opened 2019-07-19
Add Deprecation Message to Python Installer

blueandgold:
The message is just a placeholder, and welcome suggestion on what it should be.  

The format and prompt is below.

image

### DIFF
--- a/install/gcp/gcp_installer.py
+++ b/install/gcp/gcp_installer.py
@@ -99,8 +99,6 @@ def run():
         '\n\n'
     )
     print('\033[91m', deprecation_message, '\033[0m')
-    input("Press Enter to continue...")
-
 
     if not args.get('type'):
         # If the user didn't specify a type, install both server and client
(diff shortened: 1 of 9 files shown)


install/gcp/gcp_installer.py
```diff
@@ -87,6 +87,21 @@ def run():
...
     server_config = ServerConfig(**args)
 
+    deprecation_message = (
+        '\n\n'
+        '##################################################################\n'
+        '#  ATTENTION: This python installer will be deprecated           #\n'
+        '#  on 9/30/2019, and will be replaced by the easier and simpler  #\n'
+        '#  to use terraform installer.  Try out this short, interactive  #\n'
+        '#  tutorial to learn how to use the terraform installer.         #\n'
+        '#  http://foobar.com                                             #\n'
```
morgante: Can we just link to https://forsetisecurity.org/docs/latest/setup/install.html ?

blueandgold: Done

Issue #2737, opened 2019-04-11
Silence non-fatal ERROR messages from CSCC for findings that already exists.


blueandgold:
These are non-fatal, but we should silence it (i.e. not shown as error), just handle it as debug level.
```
HttpError: <HttpError 409 when requesting https://securitycenter.googleapis.com/v1/organizations/302681460499/sources/7460974056269853532/findings/a6f9f27a3a5f25c5ebad67271c39cc78?alt=json returned "Requested entity already exists">
```

ocervell:
+1 for this, those are just making users anxious :)

blueandgold:
This issue should not be closed, until the PR is actually merged.  Which will be done automatically for you when you use the keyword `fixes`.
https://help.github.com/en/articles/closing-issues-using-keywords

PR #2987, opened 2019-07-20
CSCC API exceptions are now muted

red2k18:


red2k18:
Fixes #2737

google/cloud/forseti/common/gcp_api/securitycenter.py
```diff
@@ -138,15 +137,12 @@ def create_finding(self, finding, source_id=None, finding_id=None):
...
                          response)
             return response
-        except (errors.HttpError, HttpLib2Error) as e:
-            LOGGER.exception(
-                'Unable to create CSCC finding: Resource: %s', finding)
-            violation_data = (
-                finding.get('source_properties').get('violation_data'))
-            raise api_errors.ApiExecutionError(violation_data, e)
+        except (errors.HttpError, HttpLib2Error):
+            LOGGER.debug('CSCC API exception encountered while creating '
```
blueandgold: Per issue #2737, we only want to silence the specific errors for `409 "Requested entity already exists"`.  Other errors would still be handled as exceptions, and should not be silenced.

red2k18: Updated code accordingly!

google/cloud/forseti/notifier/notifiers/cscc_notifier.py
```diff
@@ -257,8 +255,6 @@ def _send_findings_to_cscc(self, violations, source_id=None):
                     client.update_finding(finding,
                                           finding_id,
                                           source_id=source_id)
-                    LOGGER.debug('Successfully updated finding in CSCC:\n%s',
-                                 finding)
```
blueandgold: If you are going to remove this here, you should also update the `update_finding()` to make sure that the `LOGGER.debug`  there will capture the same findings details.  i.e. we don't lose any details with this change.

google/cloud/forseti/common/gcp_api/securitycenter.py
```diff
@@ -138,15 +139,27 @@ def create_finding(self, finding, source_id=None, finding_id=None):
...
             return response
         except (errors.HttpError, HttpLib2Error) as e:
-            LOGGER.exception(
-                'Unable to create CSCC finding: Resource: %s', finding)
-            violation_data = (
-                finding.get('source_properties').get('violation_data'))
-            raise api_errors.ApiExecutionError(violation_data, e)
+            arguments = e.args
+            all_args = []
+            for argument in arguments:
```
blueandgold: Why we be looping these arguments? Why not just go:
```
raw_error = e.args[1]
```

google/cloud/forseti/common/gcp_api/securitycenter.py
```diff
@@ -138,15 +139,27 @@ def create_finding(self, finding, source_id=None, finding_id=None):
...
+                all_args.append(argument)
+            raw_error = all_args[1]
+            error = raw_error.decode('utf-8')
+            formatted_error = json.loads(error)
+            error_msg = formatted_error['error']['message']
+            if error_msg == 'Requested entity already exists':
+                LOGGER.debug('CSCC API exception encountered while creating '
```
blueandgold: This debug message should be very specific, since you are handling a specific case, and can log it as such.  Something like this:
```
LOGGER.debug('Unable to create finding.  Finding already exists in CSCC. %s', finding)
```

google/cloud/forseti/notifier/notifiers/cscc_notifier.py
red2k18: These log statement had to be removed as it says created/updated finding even if there is an exception.

google/cloud/forseti/common/gcp_api/securitycenter.py
red2k18: Done

google/cloud/forseti/common/gcp_api/securitycenter.py
red2k18: Done

google/cloud/forseti/notifier/notifiers/cscc_notifier.py
blueandgold: What I'm suggesting is to just add the `finding` detail to the debug message here:

https://github.com/forseti-security/forseti-security/blob/a9a820feb400dca968eef941fc4d0fc1d8f97ba7/google/cloud/forseti/common/gcp_api/securitycenter.py#L183

google/cloud/forseti/common/gcp_api/securitycenter.py
```diff
@@ -138,15 +139,23 @@ def create_finding(self, finding, source_id=None, finding_id=None):
...
-            LOGGER.exception(
-                'Unable to create CSCC finding: Resource: %s', finding)
-            violation_data = (
-                finding.get('source_properties').get('violation_data'))
-            raise api_errors.ApiExecutionError(violation_data, e)
+            raw_error = e.args[1]
+            error = raw_error.decode('utf-8')
+            formatted_error = json.loads(error)
+            error_msg = formatted_error['error']['message']
+            if error_msg == 'Requested entity already exists':
```
blueandgold: Checking on a text output seems to be fragile, as this can stop working if the text changes.  What do you think about checking the error code `409`?  This way, it will be more robust to text changes. Is the error code in the exception?

red2k18: Yes, error code is also part of the raw error. Updated code to check for error code instead.

PR #2999, opened 2019-07-26
Add ability to specify server endpoint on CLI.

### LINKED ISSUE #2998: forseti-orchestrator endpoint by variable is not scalable
As it stands, the forseti-orchestrator sets its endpoint via an environment variable injected into the pod by Kubernetes.

https://github.com/forseti-security/forseti-security/blob/fa0ace0ad33af23e5b3a652e8695bbcc1e9e30b6/install/scripts/docker_entrypoint.sh#L116-L120

This only works if the Helm release is named **Forset**, hence the **FORSETI**_SERVER_SERVICE_HOST.

This is not scaleable and should be configurable on the CLI.

kevensen:
With this PR, the server endpoint is configurable at the command line.  This improves on the existing functionality where the server endpoint is derived from a hard coded environment variable.

This closes #2998

### DIFF
--- a/install/scripts/docker_entrypoint.sh
+++ b/install/scripts/docker_entrypoint.sh
@@ -46,6 +46,9 @@ RUN_TEST=false
 SQL_HOST=127.0.0.1
 SQL_PORT=3306
 
+SERVER_HOST=127.0.0.1
+SERVER_PORT=50051
+
 # Check if the k8s cloud sql proxy environment variables have been set, if so
 # overwrite our cloud sql variables with their contents
 if [[ !(-z ${CLOUDSQLPROXY_SERVICE_HOST}) ]]; then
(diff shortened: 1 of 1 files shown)


install/scripts/docker_entrypoint.sh
```diff
@@ -46,6 +46,9 @@ RUN_TEST=false
 SQL_HOST=127.0.0.1
 SQL_PORT=3306
 
+SERVER_HOST=127.0.0.1
```
blueandgold: I recall that Duncan before had issues with `127.0.0.1`, and he can only get things working by using `0.0.0.0`. Is there any concerns here?

kevensen: That's a fair question.  I can change it but I don't think it's an issue.  In the Helm chart, I pass in the right address in the case of the server

https://github.com/kevensen/helm-charts/blob/866ef8f53af8d234a001575c8ae07d4c613afe4a/charts/forseti-security/templates/server/deployment.yaml#L46-L49

And the right hostname in the case of the orchestrator:

https://github.com/kevensen/helm-charts/blob/866ef8f53af8d234a001575c8ae07d4c613afe4a/charts/forseti-security/templates/orchestrator/forseti-cron-job.yaml#L37-L40

PR #3002, opened 2019-07-30
Fix Missing Group Members in Inventory

### LINKED ISSUE #2969: G Suite group membership is not expanded for certain users
- Forseti version: 2.17.0
 - Module: explainer
 - No relevant errors are presented in either the client or server logs.
 - Deployed via Deployment Manager following https://forsetisecurity.org/docs/latest/setup/install.html and https://forsetisecurity.org/docs/latest/configure/inventory/gsuite.html.
 - Reporting this bug as requested in the Slack channel by @joecheuk

We're seeing an issue where Forseti apparently is unaware that some G Suite users are members of groups. Note that other users' group memberships _are_ properly expanded, so G Suite integration is working at least to some level.

For example:

- user/rich@example.org and user/tom@example.org are both members of group/ops@example.org
- group/ops@example.org has a large number of effective permissions:
`rich@forseti-client-vm-8fe5e30:~$ forseti explainer access_by_member group/ops@example.org | grep role | wc -l
243`
- user/rich@example.org is shown to inherit all these permissions:
`rich@forseti-client-vm-8fe5e30:~$ forseti explainer access_by_member user/rich@example.org | grep role | wc -l
243`
- user/tom@example.org is *not* shown to inherit them:
`rich@forseti-client-vm-8fe5e30:~$ forseti explainer access_by_member user/tom@example.org | grep role | wc -l
8`
- We have verified that, in GCP, user/tom@example.org _can_ perform actions using permissions given by his membership of group/ops@example.org, including actions that Forseti reports he should _not_ be able to, for example he can in fact power off this instance:
`rich@forseti-client-vm-8fe5e30:~$ forseti explainer check_policy instance/xxx compute.instances.stop user/tom@example.org
{
  "result": false
}`
- There are several users in our organisation affected by this issue. A cursory check seems to imply that users of long standing are affected; new joiners are not. An exhaustive test of this theory has not been completed as yet.
- For an affected user, the roles listed by `forseti explainer access_by_member` for that user are _only_ those granted directly to the user, not to a group.
- For unaffected users, _both_ direct user permissions _and_ group permissions are listed by `forseti explainer access_by_member`

So **it seems that we have a subset of users for which Forseti is unable to evaluate group memberships**.

Can someone please assist with this issue?

blueandgold:
This fixes the problem where GSuite group membership is not expanded for certain users (issue #2969).  As it turns out, it's because the members who are in multiple groups are not fully inventoried.  Please see if this is a viable solution, before I write the unit test.

image

Fixes #2969

### DIFF
--- a/google/cloud/forseti/services/inventory/storage.py
+++ b/google/cloud/forseti/services/inventory/storage.py
@@ -1350,11 +1350,16 @@ class Storage(BaseStorage):
         if self.readonly:
             raise Exception('Opened storage readonly')
 
-        previous_id = self._get_resource_id(resource)
-        if previous_id:
-            resource.set_inventory_key(previous_id)
-            self.update(resource)
-            return
+        resource_data = resource.data()
+        # Group members do not need to be checked if it already exists
+        # in Inventory, as a user can be members in multiple groups.
+        # IOW: group members must always be inserted.
+        if resource_data.get('kind') != 'admin#directory#member':
+            previous_id = self._get_resource_id(resource)
+            if previous_id:
+                resource.set_inventory_key(previous_id)
+                self.update(resource)
+                return
 
         rows = Inventory.from_resource(self.inventory_index, resource)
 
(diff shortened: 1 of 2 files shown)


blueandgold:
Yes, I have tested this manually.  Now that you approved, I will add the
unit test.  Thanks!

PR #3006, opened 2019-07-31
Fix api_key Parameter in SendGrid API Client

blueandgold:
image

https://github.com/sendgrid/sendgrid-python/blob/master/sendgrid/sendgrid.py#L59


blueandgold:
Closing, because this does not affect sendgrid version 5.6.

PR #3008, opened 2019-07-31
Add Error Handling When Root Resource Is Not Configured Properly

blueandgold:
Otherwise, the server will crash. And it's not obvious that explain must be disabled when starting the server.

### DIFF
--- a/google/cloud/forseti/services/explain/service.py
+++ b/google/cloud/forseti/services/explain/service.py
@@ -59,10 +59,14 @@ class GrpcExplainer(explain_pb2_grpc.ExplainServicer):
         Returns:
             bool: True if Explainer module can be used.
         """
-        root_resource_id = (
-            self.explainer.config.inventory_config.root_resource_id)
-        if 'organizations' in root_resource_id:
-            return True
+        try:
+            root_resource_id = (
+                self.explainer.config.inventory_config.root_resource_id)
+            if 'organizations' in root_resource_id:
+                return True
+        except (AttributeError, TypeError):
+            LOGGER.exception('Unable to check forseti server config. '
+                             'Explain will not be supported.')
         return False
 
     @staticmethod
--- a/google/cloud/forseti/services/explain/service.py
+++ b/google/cloud/forseti/services/explain/service.py
@@ -65,8 +65,8 @@ class GrpcExplainer(explain_pb2_grpc.ExplainServicer):
             if 'organizations' in root_resource_id:
                 return True
         except (AttributeError, TypeError):
-            LOGGER.exception('Unable to check forseti server config. '
-                             'Explain will not be supported.')
+            LOGGER.exception('Unable to check the root resource in the server '
+                             'config file. Explain will not be supported.')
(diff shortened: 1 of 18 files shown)


joecheuk:
What do you mean by disabling explain? We don't have any direct way of disabling it other than updating the forseti.service which we don't want users to update directly. I don't think the server would crash with composite root and explain together, when running the explain command there will be an error message returned to the user saying that explain is not supported when using composite root.

blueandgold:
@joecheuk this is the error: with using composite root in my dev environment.  Disabling meaning starting up the server without the `explain` param in the `--services` flag 
image

joecheuk:
Looks like inventory_config object is none instead of the root_resource_id is not in place, more like a bug to me - can you double check and see if the root_resource_id is commented out? It should be there with an empty string if it's not being used. e.g.
```
inventory: 
    root_resource_id: ""
    composite_root_resources: 
       - "projects/11111111111"
```

The following configuration works with explain service enabled but when using explain commands, it will returned an error message saying that it's not supported which i believe is the intended behavior.

blueandgold:
Yup, same error if i just have empty string:
```
    # Root resource to start crawling from, formatted as
    # <resource_type>/<resource_id>, (e.g. "organizations/12345677890")
    # Note: For root resource id at a project level, you will need to use
    # project number instead of project id, e.g. "projects/<project number>".
    root_resource_id: 
```

joecheuk:
`root_resource_id:` is not the same as `root_resource_id: ""`, can you try specifying an empty string there and rerunning it again?

blueandgold:
Using empty string does work.  Now, that is weird.  It could use documenation to clarify.  Either way, I'll fix the code to make this more robust, and better user experience.

blueandgold:
@joecheuk , updated the PR to add error handling.  The behaviors are now:
- Server will not start, if the root resource and composite root are not configured properly.  This is because the `config` file will not be not be set correctly, and I see little value to force the server to start.  Or else, there will be other areas where an unset config can cause the server to crash.
- Explain service will properly start, when composite root is configured and the root resource is configured as Null value, which is the most natural thing for people to do, rather than using of empty string `' '`

google/cloud/forseti/services/explain/service.py
```diff
@@ -59,10 +59,14 @@ def _determine_is_supported(self):
         Returns:
             bool: True if Explainer module can be used.
         """
-        root_resource_id = (
-            self.explainer.config.inventory_config.root_resource_id)
-        if 'organizations' in root_resource_id:
-            return True
+        try:
+            root_resource_id = (
+                self.explainer.config.inventory_config.root_resource_id)
```
joecheuk: Do you think its better to check for the none case beforehand instead of raising the exception?
Something along the line of 
```
if not (self.explainer.config.inventory_config && self.explainer.config.inventory_config.root_resource_id):
  return False
```

blueandgold: Good idea, done.

PR #3009, opened 2019-07-31
Turbinia terraform

berggren:
This is the initial PR to add Terraform deployment support for Turbinia (https://github.com/google/turbinia. Turbinia is an open-source framework for deploying, managing, and running distributed forensic workloads. It is intended to automate running of common forensic processing tools.


contrib/incident-response/infrastructure/modules/turbinia/templates/scripts/install-turbinia-worker.sh.tpl
```diff
@@ -0,0 +1,71 @@
...
+# Turbinia needs a recent version of urllib3
+pip install urllib3 --upgrade
+pip install cryptography --upgrade
+
+# Install Plaso
+add-apt-repository -y ppa:gift/stable
+apt-get -y install python-plaso plaso-tools
+
+# Install Hindsight
+pip install pyhindsight
```
aarontp: I'm a little conflicted about installing these manually instead of using the `extras_require` target for `[worker]`.  I think it's probably better to use Plaso from the Gift PPA, but now we need to manually keep the worker dependencies in sync.  Maybe we should just add a comment in the `turbinia/setup.py` to reference updating this file when any new dependencies are added.

I think this is good for now, but let's chat about how to handle this.

BTW, this will only work under python2 for now because there isn't a python3 version of hindsight yet.  See https://github.com/google/turbinia/issues/413.

contrib/incident-response/infrastructure/modules/turbinia/templates/scripts/install-turbinia-worker.sh.tpl
```diff
@@ -0,0 +1,71 @@
...
+
+# Create system user
+useradd -r -s /bin/nologin turbinia
```
aarontp: We'll need to add turbinia to the `disk` group and also make sure it's in sudoers with `NOPASSWD`.

contrib/incident-response/infrastructure/modules/turbinia/templates/scripts/install-turbinia-worker.sh.tpl
```diff
@@ -0,0 +1,71 @@
...
+until [[ -f "$${BOOT_FINISHED_FILE}" ]]; do
+  sleep 1
+done
+
+# --- MAIN ---
+apt-get update
+apt-get -y install python-pip
+
+# Install Turbinia
+pip install https://github.com/google/turbinia/archive/master.zip
```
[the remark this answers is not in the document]
aarontp: Actually, can we parameterize this?  I'd like to be able to point terraform at a specific branch or tag if possible.

berggren:
Hi @hshin-g, this PR is ready for review.

contrib/incident-response/infrastructure/modules/turbinia/main.tf
```diff
@@ -0,0 +1,259 @@
...
+ */
+
+# Random ID for creating unique resource names.
+# TODO: Allow overide by config
+resource "random_id" "infrastructure-random-id" {
+  byte_length = 8
+}
+
+# Enable GCP services individually.
+resource "google_project_service" "cloudfunctions" {
```
hshin-g: Is there a specific reason to enable the services individually (besides dependency for the resources created below)? It might be cleaner to add all the services into a list and then use the count of that list to enable the services within one block, like:

```
locals = {
  services_list = [
    "admin.googleapis.com",
    "appengine.googleapis.com",]
} 

resource "google_project_service" "main" {
  count              = length(local.services_list)
  project            = var.project_id
  service            = local.services_list[count.index]
  disable_on_destroy = "false"
}
```

contrib/incident-response/infrastructure/modules/turbinia/main.tf
```diff
@@ -0,0 +1,259 @@
...
+}
+
+resource "google_storage_bucket_object" "cloudfunction-archive" {
+  name   = "function.zip"
+  bucket = "${google_storage_bucket.output-bucket.name}"
+  source = "${path.module}/data/function.zip"
+  depends_on = ["data.archive_file.cloudfunction-archive"]
+}
+
+resource "google_cloudfunctions_function" "gettasks" {
```
hshin-g: Would it work to consolidate each of the cloudfunctions into a list in `locals` and use a count of the list to create each individual func? The params look like they are shared cross all three- if these can differ in the future/in another use case, ok to just leave as is.

contrib/incident-response/infrastructure/modules/turbinia/main.tf
berggren: That's nice! I didn't know about that pattern. Changed now.

contrib/incident-response/infrastructure/modules/turbinia/main.tf
berggren: Done.

berggren:
@hshin-g PTAL

PR #3013, opened 2019-07-31
Update export_assets method to take support different outputs.

ahoying:
* Create helper functions to build correct Output Config message for each supported destination.
* Update mixin to expect an OutputConfig dict instead of a destination object.
* Add stub for bigquery destination pending support in v1 API.

- [x] I have submitted a corresponding PR in the [Forseti Terraform module](https://github.com/forseti-security/terraform-google-forseti) (if applicable).

### DIFF
--- a/google/cloud/forseti/common/gcp_api/cloudasset.py
+++ b/google/cloud/forseti/common/gcp_api/cloudasset.py
@@ -169,6 +169,9 @@ class CloudAssetClient(object):
                 not exist.
             force (bool): If true, force write rows even on error.
         """
+        # TODO: Uncomment the below code and remove the exception when the
+        # bigquery support is live in the v1 API.
+        #
         # dataset_full_name = 'projects/{}/datasets/{}'.format(project_id,
         #                                                      dataset)
         # return {'bigqueryDestination': {'dataset': dataset_full_name,
--- a/google/cloud/forseti/common/gcp_api/cloudasset.py
+++ b/google/cloud/forseti/common/gcp_api/cloudasset.py
@@ -130,8 +130,57 @@ class CloudAssetClient(object):
             quota_period=quota_period,
             use_rate_limiter=kwargs.get('use_rate_limiter', True))
 
-    def export_assets(self, parent, destination_object, content_type=None,
-                      asset_types=None, blocking=False, timeout=0):
+    @staticmethod
+    def build_gcs_object_output(destination_object):
+        """Creates an OutputConfig message for output to GCS object.
+
+        Args:
+            destination_object (str): The GCS path and file name to store the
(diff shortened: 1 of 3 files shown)


google/cloud/forseti/common/gcp_api/cloudasset.py
```diff
@@ -165,10 +217,25 @@ def export_assets(self, parent, destination_object, content_type=None,
             raise ValueError('parent must start with folders/, projects/, or '
                              'organizations/')
 
+        if not destination_object and not output_config:
+            raise ValueError('Either destination_object or output_config must '
```
blueandgold: qq: If we are raising these `ValueError`, are they being handled correctly?

For example:
#1: is the log message appropriate when it references `root_resource_id`?
https://github.com/forseti-security/forseti-security/blob/1e60285d3fbb8b1ae1b2b8333496d24a31ac520e/google/cloud/forseti/services/inventory/base/cloudasset.py#L199-L201

#2: not seeing `ValueError` being handled
https://github.com/forseti-security/forseti-security/blob/a9a820feb400dca968eef941fc4d0fc1d8f97ba7/google/cloud/forseti/common/gcp_api/cloudasset.py#L176-L182

ahoying: If a value error is raised, it is due to an unexpected bad usage of the function and the resulting code should blow up. This should be always be caught in tests. If it happens live then something is fundamentally wrong with the way the method is called, and that should also result in an exception bubbling up and killing the process.

google/cloud/forseti/common/gcp_api/cloudasset.py
```diff
@@ -130,8 +130,57 @@ def __init__(self, global_configs, **kwargs):
...
+        """Creates an OutputConfig message for output to a BigQuery table.
+
+        Args:
+            project_id (str): The project that hosts the dataset.
+            dataset (str): The dataset name to output into.
+            table (str): The table name to create for the export, the table must
+                not exist.
+            force (bool): If true, force write rows even on error.
+        """
+        # dataset_full_name = 'projects/{}/datasets/{}'.format(project_id,
```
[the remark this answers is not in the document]
ahoying: Done.

google/cloud/forseti/common/gcp_api/cloudasset.py
ahoying: In other words, in cloudasset.py the ValueError may be better re-raised and cause the process to die rather then to silently handle the exception and log, because in that case the server is misconfigured and falling back to direct API querying is likely not what the user intended.

I'm going to be refactoring the cloudasset.py code in a future PR and will address this then.

PR #3014, opened 2019-08-01
Mute 501 Not Implemented for Listing AppEngine Instances

### LINKED ISSUE #2971: Appengine instances list api returning 501
Example:
<HttpError 501 when requesting https://appengine.googleapis.com/v1/apps/policyscanner-henry/services/default/versions/1/instances?alt=json returned \"Managed VMs is not supported. Please contact support.\">\n{\n  \"error\": {\n    \"code\": 501,\n    \"message\": \"Managed VMs is not supported. Please contact support.\",\n    \"status\": \"UNIMPLEMENTED\"\n

Should investigate and mute the message if that's the expected behavior of the API.

blueandgold:
Fixes #2971  Please see issue #2971 for screenshot of the error in cloud console UI and gcloud.  So, it's okay to mute.

```
2019-07-31 23:49:59,719 DEBUG google.cloud.forseti.common.gcp_api.appengine(list_instances): <HttpError 501 when requesting https://appengine.googleapis.com/v1/apps/policyscanner-henry/services/default/versions/1/instances?alt=json returned "Managed VMs is not supported. Please contact support.". Details: "[{'@type': 'type.googleapis.com/google.rpc.DebugInfo', 'detail': '[ORIGINAL ERROR] generic::unimplemented: Managed VMs is not supported. Please contact support. [google.rpc.error_details_ext] { code: 12 message: "Managed VMs is not supported. Please contact support." }'}]">
```

### DIFF
--- a/google/cloud/forseti/common/gcp_api/appengine.py
+++ b/google/cloud/forseti/common/gcp_api/appengine.py
@@ -430,7 +430,7 @@ class AppEngineClient(object):
                          project_id, service_id, version_id, flattened_results)
             return flattened_results
         except (errors.HttpError, HttpLib2Error) as e:
-            if e.args[0].status == 501:  # pylint: disable=no-member
+            if e.resp.status == 501:  # pylint: disable=no-member
                 LOGGER.debug(e)
                 return []
             if _is_status_not_found(e):
--- a/google/cloud/forseti/common/gcp_api/appengine.py
+++ b/google/cloud/forseti/common/gcp_api/appengine.py
@@ -430,7 +430,7 @@ class AppEngineClient(object):
                          project_id, service_id, version_id, flattened_results)
             return flattened_results
         except (errors.HttpError, HttpLib2Error) as e:
-            if e.args[0].status == 501:
+            if e.args[0].status == 501:  # pylint: disable=no-member
                 LOGGER.debug(e)
                 return []
             if _is_status_not_found(e):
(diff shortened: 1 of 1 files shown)


google/cloud/forseti/common/gcp_api/appengine.py
```diff
@@ -430,6 +430,9 @@ def list_instances(self, project_id, service_id, version_id):
                          project_id, service_id, version_id, flattened_results)
             return flattened_results
         except (errors.HttpError, HttpLib2Error) as e:
+            if e.args[0].status == 501:  # pylint: disable=no-member
```
joecheuk: Is this the right way of checking status code? I thought you can access the status by doing this `e.resp.status == 501`. Can we also add a test case for this

joecheuk: Here is an example on how we did checks similar to this before:
https://github.com/forseti-security/forseti-security/blob/3e489059bc5b3049e6671a8e45cb32b9bd0730ea/google/cloud/forseti/common/gcp_api/compute.py#L47

blueandgold: Done on the `e.resp.status`, good tip

blueandgold: Hmmm, in that example, `403` can be very broad, and is giving the ability to check for specific errors and handle it differently.  But in this case of `501`, I don't see how many different ways this needs to be handled.

PR #3023, opened 2019-08-05
Cloud Asset temporary table migration to local sqlite database off CloudSQL

ahoying:
* Update CAI temporary table read methods to use SQLAlchemy Core.
  * Remove per thread sessions from CAI crawler implementation
  * Remove ORM overhead when reading assets from CAI temporary table

* Move the CAI temporary table to a local sqlite database.
  * This reduces network load by 50% by reducing round trips to the cloudsql server, removing 700MB of network traffic when 350MB of test assets is imported.
  * Total time to import data into the sqlite database is further reduced from 150 seconds to 30 seconds for 350MB of test assets.
  * Temporary file is cleaned up at the end of inventory, reducing storage requirements on server.

* Further optimize sqlite database interface.
  * Set sqlite pragmas on connect to speed up writes and reads. Can write 350MB in 11 seconds on test system.
  * Optimize per thread reading, can complete over 100 queries per second on 10 threads with an 8 CPU virtual machine. Queried 5000 project cloudsql instances in 45 seconds on test VM.

This is part of a series of changes to reduce the run time of the
forseti inventory crawler for large organizations.

- [ ] All of the [tests](https://forsetisecurity.org/docs/latest/develop/dev/testing.html) pass


tests/services/inventory/cai_temporary_storage_test.py
```diff
@@ -0,0 +1,149 @@
+"""Tests for google3.experimental.users.ahoying.forseti-security.tests.services.inventory.cai_temporary_storage."""
```
joecheuk: We might want to remove the ref to google3 here

google/cloud/forseti/services/inventory/cai_temporary_storage.py
```diff
@@ -0,0 +1,421 @@
...
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.pool import SingletonThreadPool
+
+from google.cloud.forseti.common.util import logger
+from google.cloud.forseti.services.dao import create_engine
+from google.cloud.forseti.services.inventory.base.gcp import AssetMetadata
+
+LOGGER = logger.get_logger(__name__)
+BASE = declarative_base()
+MAX_ALLOWED_PACKET = 32 * 1024 * 1024  # 32 Mb default mysql max packet size
```
joecheuk: This needs to be updated for sqlite, the default setting is 1mb and can be maximized to 1gb if you customize the value of the flag.
The maximum number of bytes in the text of an SQL statement is limited to SQLITE_MAX_SQL_LENGTH which defaults to 1000000. You can redefine this limit to be as large as the smaller of SQLITE_MAX_LENGTH and 1073741824.

google/cloud/forseti/services/inventory/cai_temporary_storage.py
```diff
@@ -0,0 +1,421 @@
...
+        """Creates a database row object from the json data in a dump file.
+
+        Args:
+            asset_json (str): The json representation of an Asset.
+
+        Returns:
+            dict: database row dictionary or None if there is no data.
+        """
+        asset = json.loads(asset_json)
+        if len(asset['name']) > 512:
```
joecheuk: Is this something we can resolve w/ the new approach, to increase the length of the asset name?

google/cloud/forseti/services/inventory/base/cai_gcp_client.py
```diff
@@ -76,39 +76,23 @@ def _fixup_resource_keys(resource, key_map, only_fixup_lists=False):
...
-            # SQLite doesn't support per thread sessions cleanly, so use global
-            # session.
-            return self.cai_session
-
-        if hasattr(self._local, 'cai_session'):
-            return self._local.cai_session
-
-        self._local.cai_session = db.create_readonly_session(engine=self.engine)
-        return self._local.cai_session
+    def __del__(self):
```
joecheuk: Can't seem to find where the file is being cleaned up, where do we plan to call this method?

tests/services/inventory/cai_temporary_storage_test.py
ahoying: Oops!

google/cloud/forseti/services/inventory/cai_temporary_storage.py
ahoying: sqlite doesn't really use packets as such, so this is more of the limit of the number of inserts in a single execute call, I've updated the constant name and documentation.

the MAX_SQL_LENGTH applies to the statement portion, not the bound values, and as such we are far under the 1MB size.

google/cloud/forseti/services/inventory/base/cai_gcp_client.py
ahoying: This is called as soon as the class goes out of scope in crawler.run_crawler. It's called by the garbage collector.

google/cloud/forseti/services/inventory/cai_temporary_storage.py
ahoying: Good call! Updated to 2048.

ahoying:
> Awesome, thanks for this change! Again, the only thing I am missing is how / where the Core style is used. Can you please point that out? Thanks!

The full implementation of google/cloud/forseti/services/inventory/cai_temporary_storage.py  uses Core instead of ORM.

PR #3033, opened 2019-08-06
Removed slash to follow style guide

red2k18:


google/cloud/forseti/notifier/notifier.py
```diff
@@ -49,9 +49,8 @@ def find_notifiers(notifier_name):
         for filename in dir(module):
             obj = getattr(module, filename)
 
-            if inspect.isclass(obj) \
-                    and issubclass(obj, BaseNotification) \
-                    and obj is not BaseNotification:
+            if inspect.isclass(obj) and issubclass(obj, BaseNotification) and (
```
blueandgold: If we really aim to make this cleaner, can you instead do this?
```
if (inspect.isclass(obj) and
    issubclass(obj, BaseNotification) and
    obj is not BaseNotification):
```

Issue #3038, opened 2019-08-07
 Improve SQL Server write bottlenecks


dekuhn:
- [ ] Improve Cloud SQL Server single write bottlenecks to improve Forseti Inventory performance
- [ ] Break apart the storage class into storage writer (use sqlalchemy core, optimize for fast bulk writes) and the storage reader class, which can remain using ORM as it does today.
- [ ] Update inventory and crawler initialization to use the new writer class with a a sqlalchemy engine that maintains a pool of connections shared between threads, and not use a global session object.
- [ ] Update the ParallelCrawler class to to not use a global lock on writes and updates.	
- [ ]  Optimize the writer class to have fewer SQL calls, optimize for bulk writes and few or no updates. Updates should be an exception, and strongly discouraged in code or removed completely (also needed for BigQuery streaming writer eventually).

ahoying:
Additional work completed in PR #3068. Closing issue.

PR #3050, opened 2019-08-08
Refactored inventory to remove global write lock on database.

ahoying:
Refactored inventory to remove global write lock on database.

* Move methods to read from the inventory table into DataAccess class.
* Use SQLAlchemy Core for writing inventory rows to Cloud SQL.
* Removed all updates to inventory data, each row is only written once.
* Added full_name to the inventory table for each resource.
* Created a memory cache for written resources, detects and skips duplicate full names. This introduces a small global lock on updating the cache, but is less than a millisecond on average and removes a round trip query to sql.
* Ensured warning messages written to the inventory index table always contain the full context.
* Fix external project scanner to use new DAO for inventory data.

- [ ] All of the [tests](https://forsetisecurity.org/docs/latest/develop/dev/testing.html) pass


ahoying:
Fixes #3038

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -937,217 +906,50 @@ def write(self, resource):
...
-        if resource_data.get('kind') != 'admin#directory#member':
-            previous_id = self._get_resource_id(resource)
-            if previous_id:
-                resource.set_inventory_key(previous_id)
-                self.update(resource)
+        # Use a lock to quickly check if this is a duplicate resource before
+        # updating the cache and proceeding.
+        with self._storage_lock:
+            if resource.get_full_resource_name() in self._wrote_resources:
+                LOGGER.warning('Duplicate Resource in inventory, skipping %s',
```
blueandgold: Hmmm, are group members handled properly here? i.e. can group members in multiple groups can be added?

ahoying: Yes, groups members full names are like this (they are children of the group they are in): 
organization/111222333/gsuite_group/1013/gsuite_group_member/1025/
organization/111222333/gsuite_group/1014/gsuite_group_member/1025/

This is explicitly tested for in storage_test.py test_whether_resource_should_be_inserted_or_updated.

ahoying:
> Look great overall, one comment on group membership below. Also, it seems that ORM is still being used for reads, is there any benefit to also use sqlalchemy core?

At this point it is useful to use ORM model for reads because it is a nicer data interface for querying and working with results. That can be evaluated / addressed in the future if needed.

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -731,6 +665,97 @@ def get_inventory_indexes_older_than_cutoff(  # pylint: disable=invalid-name
...
+            base_query = base_query.filter(qry_filter)
+
+        base_query = base_query.order_by(Inventory.id.asc())
+
+        for row in base_query.yield_per(PER_YIELD):
+            yield row
+
+    @classmethod
+    def get_root(cls, session, inventory_index_id):
+        """get the resource root from the inventory.
```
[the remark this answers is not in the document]
ahoying: Done.

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -331,7 +333,9 @@ def from_resource(cls, index, resource):
             resource (Resource): Crawled resource.
 
         Returns:
-            object: database row object.
+            Tuple[list, list]: A tuple containing a single row for the main
```
joecheuk: resource row is of type dict instead of list

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -731,6 +665,97 @@ def get_inventory_indexes_older_than_cutoff(  # pylint: disable=invalid-name
...
+            session (object): Database session.
+            inventory_index_id (str): the id of the inventory to open.
+            type_list (list): List of types to iterate over, or [] for all.
+            fetch_category (Categories): The category of data to fetch.
+            with_parent (bool): Join parent with results, yield tuples.
+
+        Yields:
+            object: Single row object or child/parent if 'with_parent' is set.
+        """
+        filters = [Inventory.inventory_index_id == inventory_index_id,
```
joecheuk: This looks like an expression to me, will it preserve the query for you to use later on? Looks like it will get evaluated to a list of boolean i.e. [True, True]

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -937,217 +906,50 @@ def write(self, resource):
...
-        # in Inventory, as a user can be members in multiple groups.
-        if resource_data.get('kind') != 'admin#directory#member':
-            previous_id = self._get_resource_id(resource)
-            if previous_id:
-                resource.set_inventory_key(previous_id)
-                self.update(resource)
+        # Use a lock to quickly check if this is a duplicate resource before
+        # updating the cache and proceeding.
+        with self._storage_lock:
+            if resource.get_full_resource_name() in self._wrote_resources:
```
joecheuk: I see why the full name has to be diff here for policies, would keeping that data in the full name be the best approach or should we append the type at the end when checking instead?

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -937,217 +906,50 @@ def write(self, resource):
...
-
-        Raises:
-            Exception: If storage was opened readonly.
-        """
+        (resource_row, policy_rows) = Inventory.from_resource(
+            self.inventory_index, resource)
 
-        if self.readonly:
-            raise Exception('Opened storage readonly')
+        # Insert first row to get the primary key for the resource
```
joecheuk: if we leverage the relationship pattern in sqlalchemy we can do bulk insert without worrying about the primary key / foreign key reference, what do you think about that approach?

Note: I do not know about the performance about this approach

https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -287,6 +287,7 @@ class Inventory(BASE):
 
     id = Column(Integer, primary_key=True, autoincrement=True)
     inventory_index_id = Column(BigInteger)
+    full_name = Column(String(2048), nullable=False)
```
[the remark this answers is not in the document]
ahoying: It is the same size as the column in the existing data model table. At this point the idea is to move this usage out of the data model tables back to the inventory table for scanners. Larger architecture changes can come later and may involve moving away from the full_name concept.

google/cloud/forseti/services/inventory/storage.py
```diff
@@ -350,111 +353,83 @@ def from_resource(cls, index, resource):
...
-                    category=Categories.iam_policy,
-                    resource_id=resource.key(),
-                    resource_type=resource.type(),
-                    resource_data=json.dumps(iam_policy, sort_keys=True),
-                    other=other,
-                    inventory_errors=None))
+            policy_rows.append(dict(
+                base_row,
+                category=Categories.iam_policy,
+                full_name=cls._get_policy_full_name(resource, 'iam_policy'),
```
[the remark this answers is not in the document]
ahoying: Yes, full name is different for policies (as commented below.)

google/cloud/forseti/services/inventory/storage.py
ahoying: Fixed.

google/cloud/forseti/services/inventory/storage.py
ahoying: The == operator is overloaded for sqlAlchemy columns, so this produces an object, not a boolean: https://docs.sqlalchemy.org/en/13/core/tutorial.html#operators.

google/cloud/forseti/services/inventory/storage.py
ahoying: Having a single field that is unique for each row is useful. I'd rather not have to combine multiple fields to check uniqueness. This may change if we move closer to CAI's model in the future, but that can be addressed separately.

google/cloud/forseti/services/inventory/storage.py
ahoying: I expect behind the scenes the performance is the same or worse, as it has to do the same set of operations in order to create the row, get the id, and the use the id to create the other rows. This way we are explicit. And I'd like to keep from having to maintain a session in each thread in order to do inserts.

google/cloud/forseti/services/inventory/storage.py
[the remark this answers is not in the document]
ahoying: Let me know if you feel strongly that it should be longer than what was used for the model.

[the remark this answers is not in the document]
joecheuk: You are right on the scope of this PR. That should be a separate issue to investigate and address, opened an issue to track this: https://github.com/forseti-security/forseti-security/issues/3052

PR #3064, opened 2019-08-09
Add try except block for CAI export

hshin-g:
+ Add try except block for CAI export

### DIFF
--- a/google/cloud/forseti/services/inventory/base/cloudasset.py
+++ b/google/cloud/forseti/services/inventory/base/cloudasset.py
@@ -201,7 +201,8 @@ def _export_assets(cloudasset_client, config, root_id, content_type):
                     blocking=True,
                     timeout=timeout)
             else:
-                raise e
+                LOGGER.warning('API Error getting cloud asset data: %s', e)
+                return None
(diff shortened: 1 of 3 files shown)


google/cloud/forseti/services/inventory/base/cloudasset.py
```diff
@@ -180,13 +180,20 @@ def _export_assets(cloudasset_client, config, root_id, content_type):
...
-                                                  blocking=True,
-                                                  timeout=timeout)
+        try:
+            results = cloudasset_client.export_assets(root_id,
+                                                      export_path,
+                                                      content_type=content_type,
+                                                      asset_types=asset_types,
+                                                      blocking=True,
+                                                      timeout=timeout)
+        except api_errors.HTTPError as e:
```
blueandgold: I think the lower level `export_assets` method would raise the error here as `api_errors.ApiExecutionError`. So, that should be what you catch, and then check on the `400` resp.code

hshin-g: Addressed.

PR #3068, opened 2019-08-11
Move inventory warning messages to a child table of inventory_index.

ahoying:
This change allows warnings to be written in parallel from multiple
threads, and removes the need to update the existing inventory_index row
for each new warning.

Warnings are loaded along side the inventory index through the DAO.

- [x] I have submitted a corresponding PR in the [Forseti Terraform module](https://github.com/forseti-security/terraform-google-forseti) (if applicable).


ahoying:
> The only problem I can think of is the previous warnings / errors might not be shown correctly but if we are ok with that then this looks good!

I've added an update to ensure previous warnings are included if they exist.

PR #3080, opened 2019-08-13
Added locking before incrementing object count.

joecheuk:
+= is not threadsafe


blueandgold:

Issue #3034, opened 2019-08-06
TypeError: a bytes-like object is required, not 'str' in file_uploader


abhishekupadhyaya:
Hi,

We are using `release-2.18.0`. We found a following error in `upload_json` method of [`file_uploader.py`](https://github.com/forseti-security/forseti-security/blob/release-2.18.0/google/cloud/forseti/common/util/file_uploader.py):
```
TypeError: a bytes-like object is required, not 'str'
```
Line 38 in file_uploader.py should instead be:
```
tmp_data.write(parser.json_stringify(data).encode('utf-8'))
```
This can also be modified in `json_stringify` method in `parser.py`

Thank you,
Abhishek

blueandgold:
@abhishekupadhyaya Can you give us the entire stacktrace so that we can see what is the caller the caused this error?

marciosian:
I'm facing the same issue. v2.18  

2019-08-13 10:05:34,283 ERROR [forseti-security][2.18.0] google.cloud.forseti.common.util.file_uploader(upload_json): Unable to upload json document to bucket gs://forseti-server-***/scanner_violations/violations.ke_version_violations.1565690410195284.20190813T100534Z.json:
[{'id': 859, 'created_at_datetime': '2019-08-13T10:05:32Z', ***
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/common/util/file_uploader.py", line 38, in upload_json
    tmp_data.write(parser.json_stringify(data))
  File "/usr/lib/python3.6/tempfile.py", line 624, in func_wrapper
    return func(*args, **kwargs)
TypeError: a bytes-like object is required, not 'str'

If i switch to csv it works well.

kevensen:
@marciosian Thank you for providing the stack trace.  I'll investigate.

kevensen:
I have reproduced the issue.  Continuing to investigate.

kevensen:
@abhishekupadhyaya @marciosian The culprit is indeed line 38 of file uploader.

https://github.com/forseti-security/forseti-security/blob/b54faf68d869842e8a43472ff980e28e2ce8d3c6/google/cloud/forseti/common/util/file_uploader.py#L36-L39

In file_uploader.py, could you please change line 38 to be:

```python
tmp_data.write(parser.json_stringify(data).encode('utf-8'))
```

Then run the below to apply the changes:
```bash
[...]

PR #3084, opened 2019-08-14
Fix string to bytes issue in notifier

kevensen:
Addresses #3034


google/cloud/forseti/common/util/file_uploader.py
```diff
@@ -35,7 +35,7 @@ def upload_json(data, gcs_upload_path):
     """
     try:
         with tempfile.NamedTemporaryFile() as tmp_data:
-            tmp_data.write(parser.json_stringify(data))
+            tmp_data.write(parser.json_stringify(data).encode('utf-8'))
```
hshin-g: Should change to `.encode()` since I think by default `.encode()` uses `utf-8` and this fits with how we have done other `.encode()` in the code.

PR #3086, opened 2019-08-15
Update .travis.yml

dekuhn:
Adding [travis_wait](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) command to not timeout after 10 mins if no console output is received.

This should address Travis timeout issues when external contributors submit website changes.

Here's a handy checklist to ensure your PR goes smoothly.

- [X] I signed Google's [Contributor License Agreement](https://opensource.google.com/docs/cla/)
- [X] My code conforms to Google's [python style](https://google.github.io/styleguide/pyguide.html).
- [X] My PR at a minimum doesn't decrease unit-test coverage (if applicable).
- [X] My PR has been functionally tested.
- [X] All of the [tests](https://forsetisecurity.org/docs/latest/develop/dev/testing.html) pass
- [N/A] I have submitted a corresponding PR in the [Forseti Terraform module](https://github.com/forseti-security/terraform-google-forseti) (if applicable).

These guidelines and more can be found in our [contributing guidelines](https://github.com/forseti-security/forseti-security/blob/dev/.github/CONTRIBUTING.md).

PR #3090, opened 2019-08-16
Fix auth in the jekyll website build

mirons-google:
By specifying the correct organization when the forseti-security-travis-bot credentials are used.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -46,7 +46,7 @@ markdown: kramdown
 destination: _www/www/
 
 # Needed for local builds and using {{ site.github }}.
-repository: GoogleCloudPlatform/forseti-security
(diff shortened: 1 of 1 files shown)


kevensen:
@mirons-google @dekuhn I think the build is still failing due to the build timing out.  I've merged in @dekuhn's PR #3086 which should increase the timeout.  That build is now failing due to invalid credentials

kevensen:
Unfortunately this build is still failing.  My guess is that it is still timing out.

mirons-google:
@kevensen @dekuhn PTAL as a failing build should be un-releated to this change, because jekyll build will still work just not crawl the repo as an authenticated user.

PR #3092, opened 2019-08-17
Target a newer ruby, travis host, and jekyll versions.

mirons-google:


kevensen:
@mirons-google Do you mind updating the branch?  Other than that, this LGTM

mirons-google:
@dekuhn PTAL

mirons-google:
@dekuhn PTAL

Gemfile
```diff
@@ -1,15 +1,19 @@
 source 'https://rubygems.org'
 
+gem 'listen'
+gem 'rouge'
+gem 'execjs'
+gem 'therubyracer'
+gem 'html-proofer'
+gem 'liquid-c'
```
gkowalski-google: This fails to install for me on gLinux. Did you do anything special to get this installed?

```
cc1: error: unrecognized command line option ‘-Wno-self-assign’ [-Werror]
cc1: error: unrecognized command line option ‘-Wno-constant-logical-operand’ [-Werror]
cc1: error: unrecognized command line option ‘-Wno-parentheses-equality’ [-Werror]
cc1: all warnings being treated as errors
make: *** [Makefile:242: block.o] Error 1

make failed, exit code 2
```

.travis.yml
```diff
@@ -1,10 +1,11 @@
+dist: bionic
 sudo: required
 language: ruby
 services:
   - docker
 cache: bundler
 rvm:
-- 2.5.3
+- 2.6.3
```
gkowalski-google: Is this the Ruby version or the RVM version? Should we delete the `.ruby-version` file from the repo since this will override that file?

mirons-google: This is the ruby version [AFAIK](https://docs.travis-ci.com/user/languages/ruby/#specifying-ruby-versions-and-implementations). I didn't realize we had a .ruby-version file, we should use that because Travis will read it and we can also support developer clients that way too.

Gemfile
mirons-google: I installed liquid-c through the normal GEM/bundler installation process using RVM to manage jekyll versions.

mirons-google:
@dekuhn and @gkowalski-google Thanks, PTAL.

mirons-google:
I should add that jekyll 4.0.0 has been released and was rumored to have *significant* speed improvements -- it might be worth trying this branch with that version set and comparing --profile results.

Gemfile
gkowalski-google: Is RVM easy to setup? Can you provide any steps/resources for how to set this up locally? Our [website](https://forsetisecurity.org/docs/latest/develop/website.html) and internal docs do not mention RVM. I can update these docs, just need to know what to do.

red2k18:
#3397 to update to Jekyll 4.0

@mirons-google Can this PR be merged in?

Gemfile
red2k18: @gkowalski-google Looks like RVM is not being used. @mirons-google updated ruby version in `.ruby-version` file.

red2k18: @gkowalski-google Were you able to install `liquid-c` on your local? I was able to do so by pulling the latest `Gemfile` and by running `bundle update`. Looks like there are no other breaking changes.

gkowalski-google: @red2k18 I tried again, and I still get the same error. I tried with the latest gemfile and `bundle update` along with `gem install liquid-c -v '4.0.0' --source 'https://rubygems.org/'`. This might be dependent on OS. I believe [Travis is using RVM](https://docs.travis-ci.com/user/languages/ruby/#specifying-ruby-versions-and-implementations) and switching to this locally would probably resolve my issue. I don't want to hold up this PR if it's ready, my only concern is if this will cause issues for the next release engineer if they run into the same issue as me. @kevensen Can you try running `gem install liquid-c -v '4.0.0' --source 'https://rubygems.org/'` and see if you get this error?

kevensen: @gkowalski-google I tried @red2k18 instructions and it worked.  I think you have to run `bundle install`

kevensen:
FWIW, I was able to install the updated gem and build the website.

Gemfile
gkowalski-google: I have tried that, several times, no dice. Will have to fix my issue later.

PR #3096, opened 2019-08-19
Add Spans for CAI and API Operations

blueandgold:
Add spans for CAI and API operations so that we can get a more complete & detailed picture of what is happening in the Inventory trace.  The screenshot below shows some missing gaps being filled, but I am not unable to see the following spans despite them being called. Would appreciate any suggestions.

#1: span for `_export_asset()` is not shown

#2: span for `BigQueryClient.get_tables()` is not shown. 

#3: Spans for all the methods in `ServiceManagementClient` class are not shown.

For the API methods, I am not adding spans on the actual request call, but rather the higher level methods.  I would expect span for this would work.

image

### DIFF
--- a/google/cloud/forseti/common/gcp_api/bigquery.py
+++ b/google/cloud/forseti/common/gcp_api/bigquery.py
@@ -238,7 +238,7 @@ class BigQueryClient(object):
         except (errors.HttpError, HttpLib2Error) as e:
             raise api_errors.ApiExecutionError(project_id, e)
 
-    @tracing.traced()
+    @tracing.trace()
     def get_tables(self, project_id, dataset_id):
         """Return BigQuery tables stored in the requested project_id.
 
--- a/google/cloud/forseti/common/gcp_api/bigquery.py
+++ b/google/cloud/forseti/common/gcp_api/bigquery.py
@@ -21,6 +21,7 @@ from google.cloud.forseti.common.gcp_api import _base_repository
 from google.cloud.forseti.common.gcp_api import api_helpers
 from google.cloud.forseti.common.gcp_api import errors as api_errors
 from google.cloud.forseti.common.gcp_api import repository_mixins
+from google.cloud.forseti.common.opencensus import tracing
 from google.cloud.forseti.common.util import logger
 
 LOGGER = logger.get_logger(__name__)
@@ -237,6 +238,7 @@ class BigQueryClient(object):
         except (errors.HttpError, HttpLib2Error) as e:
             raise api_errors.ApiExecutionError(project_id, e)
 
+    @tracing.traced()
     def get_tables(self, project_id, dataset_id):
         """Return BigQuery tables stored in the requested project_id.
 
--- a/google/cloud/forseti/common/gcp_api/servicemanagement.py
+++ b/google/cloud/forseti/common/gcp_api/servicemanagement.py
@@ -21,6 +21,7 @@ from google.cloud.forseti.common.gcp_api import _base_repository
 from google.cloud.forseti.common.gcp_api import api_helpers
 from google.cloud.forseti.common.gcp_api import errors as api_errors
 from google.cloud.forseti.common.gcp_api import repository_mixins
+from google.cloud.forseti.common.opencensus import tracing
 from google.cloud.forseti.common.util import logger
 
 LOGGER = logger.get_logger(__name__)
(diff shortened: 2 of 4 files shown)


google/cloud/forseti/common/gcp_api/bigquery.py
```diff
@@ -237,6 +238,7 @@ def get_dataset_access(self, project_id, dataset_id):
         except (errors.HttpError, HttpLib2Error) as e:
             raise api_errors.ApiExecutionError(project_id, e)
 
+    @tracing.traced()
```
ocervell: This decorator is used for classes, use the `@tracing.trace()` instead for class methods or functions

google/cloud/forseti/services/inventory/crawler.py
```diff
@@ -420,7 +422,7 @@ def run_crawler(storage,
         LOGGER.info('SQLite used, disabling parallel threads.')
         parallel = False
 
-    client = _api_client_factory(storage, config, parallel)
+    client = _api_client_factory(storage, config, parallel, tracer)
```
ocervell: No need to pass `tracer` explicitly here, since this method call is in the same context as the calling method. The tracer will be inferred from context and set to the `tracer` argument of the method automatically.

google/cloud/forseti/services/inventory/crawler.py
```diff
@@ -333,7 +335,7 @@ def _api_client_factory(storage, config, parallel):
     if config.get_cai_enabled():
         # TODO: When CAI supports resource exclusion, update the following
         #       method to handle resource exclusion during export time.
-        asset_count = cloudasset.load_cloudasset_data(storage.session, config)
+        asset_count = cloudasset.load_cloudasset_data(storage.session, config, tracer)
```
ocervell: No need to pass tracer explicitly here, since this method call is in the same context as the calling method. The tracer will be inferred from context and set to the `tracer` argument of the method automatically.

google/cloud/forseti/common/gcp_api/servicemanagement.py
```diff
@@ -121,6 +122,7 @@ def get_formatted_service_name(service_name):
         return service_name
 
 
+@tracing.traced()
```
ocervell: This would create a separate span since it won't be aware of the parent `trace_id`. The tracer should be passed when instanciating this class in `forseti.services.inventory.base.gcp.py` (line 1186).

ocervell:
Regarding missing spans, make sure they don't show up under different trace ids in Stackdriver Trace. It is very common to lose context when using OpenCensus, and if you do you create new traces.

google/cloud/forseti/common/gcp_api/bigquery.py
blueandgold: Done.

google/cloud/forseti/common/gcp_api/servicemanagement.py
blueandgold: Hmmm, how do I get the tracer in `gcp.py` so that I can pass into this class?

google/cloud/forseti/services/inventory/crawler.py
blueandgold: The reason why I passed this tracer in here is because I need to pass in the tracer into the lower level `_export_assets()` method that is inside `load_cloudasset_data()`.  So, inside `load_cloudasset_data()`, how do I get a hold of the tracer?

google/cloud/forseti/services/inventory/crawler.py
blueandgold: Done.

google/cloud/forseti/services/inventory/crawler.py
ocervell: Since `load_cloudasset_data()` has the `tracing.trace()` decorator and the `tracer=None` arg in its definition, the current context tracer will be injected by the decorator when calling the function, and you can just access it within the function using the `tracer` arg.

blueandgold: Hmmm, although not passing in the tracer works here, I still can not see the lower level `_export_assets()` in the tracing output (see below).  What would be the way to make that show up?  Note: I am passing the tracer into `export_assets()`, inside the `load_cloudasset_data()`.  @ocervell 
  
image

blueandgold:
We will be using Cloud Profiler instead of Stackdriver Trace.

PR #3101, opened 2019-08-20
Do not inactivate findings that are already inactive

katze120:
- [] I have submitted a corresponding PR in the [Forseti Terraform module](https://github.com/forseti-security/terraform-google-forseti) (if applicable).

Currently the cscc notifiers try to inactivate outdated findings that are already INACTIVE again.

This PR resolves this issue and has 2 benefits:
1. Reduce time taken for CSCC Notifier run - it saves 40 mins in my testing org with 5k inactive findings)
2. Reduce number of Cloud SCC API calls - Before the change there was "Resource has been exhausted (e.g. check quota)." error from Cloud SCC API calls

### DIFF
--- a/google/cloud/forseti/notifier/notifiers/cscc_notifier.py
+++ b/google/cloud/forseti/notifier/notifiers/cscc_notifier.py
@@ -163,7 +163,7 @@ class CsccNotifier(object):
     def find_inactive_findings(new_findings, findings_in_cscc):
         """Finds the currently ACTIVE findings
            that do not correspond to the latest scanner run
-           and updates their statesto inactive.
+           and updates their states to inactive.
 
         Args:
             new_findings (list): Latest violations that are transformed to
@@ -188,8 +188,10 @@ class CsccNotifier(object):
             finding_id = finding_list[0]
             to_be_updated_finding = finding_list[1]
 
-            if (finding_id not in new_findings_map and
-                    to_be_updated_finding['state'] == 'ACTIVE'):
+            if to_be_updated_finding['state'] == 'INACTIVE':
+                continue
+
+            if finding_id not in new_findings_map:
                 to_be_updated_finding['state'] = 'INACTIVE'
                 current_time = date_time.get_utc_now_datetime()
                 actual_time = current_time.strftime(
--- a/google/cloud/forseti/notifier/notifiers/cscc_notifier.py
+++ b/google/cloud/forseti/notifier/notifiers/cscc_notifier.py
@@ -161,8 +161,9 @@ class CsccNotifier(object):
 
     @staticmethod
     def find_inactive_findings(new_findings, findings_in_cscc):
-        """Finds the findings that does not correspond to the latest scanner run
-           and updates it's state to inactive.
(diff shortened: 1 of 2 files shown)


katze120:
Looks like the ci errors come from test_cud_firewall_rule_read_only
Expected target links are https://www.googleapis.com/compute/v1/projects/project1/global/firewalls/fake-firewall
while actual target links are https://compute.googleapis.com/compute/v1/projects/project1/global/firewalls/fake-firewall

Not able to reproduce the issue when running ./install/scripts/docker_unittest_forseti.sh on my dev machine. Yet to figure out there the actual target link comes from.

google/cloud/forseti/notifier/notifiers/cscc_notifier.py
```diff
@@ -187,7 +188,8 @@ def find_inactive_findings(new_findings, findings_in_cscc):
             finding_id = finding_list[0]
             to_be_updated_finding = finding_list[1]
 
-            if finding_id not in new_findings_map:
+            if (finding_id not in new_findings_map and
+                    to_be_updated_finding['state'] == 'ACTIVE'):
```
blueandgold: Instead of using the `and` operator, can we make this check clearer and simpler by separating it?  Such as below?

```
if to_be_updated_finding['state'] == 'INACTIVE':
    continue

if (finding_id not in new_findings_map:
    <rest of the code>

```

blueandgold: Can you also please update the unit test here?  It should be very straightforward, just add a new data with `INACTIVE` state, and current assertion should still pass without any modification.

https://github.com/forseti-security/forseti-security/blob/dev/tests/notifier/notifiers/cscc_notifier_test.py#L125-L186

google/cloud/forseti/notifier/notifiers/cscc_notifier.py
```diff
@@ -161,8 +161,9 @@ def _transform_for_api(self, violations, source_id=None):
 
     @staticmethod
     def find_inactive_findings(new_findings, findings_in_cscc):
-        """Finds the findings that does not correspond to the latest scanner run
-           and updates it's state to inactive.
+        """Finds the currently ACTIVE findings
+           that do not correspond to the latest scanner run
+           and updates their statesto inactive.
```
blueandgold: nit: `statesto` should be `states to`

google/cloud/forseti/notifier/notifiers/cscc_notifier.py
katze120: updated as suggested

google/cloud/forseti/notifier/notifiers/cscc_notifier.py
katze120: updated!

google/cloud/forseti/notifier/notifiers/cscc_notifier.py
katze120: updated!

PR #3105, opened 2019-08-20
Updated to use the engine execute method on rollback / commit.

joecheuk:


google/cloud/forseti/services/inventory/storage.py
```diff
@@ -900,23 +900,34 @@ def rollback(self):
...
     def commit(self):
         """Commit the stored inventory."""
         if self.inventory_index.inventory_index_warnings:
             status = IndexState.PARTIAL_SUCCESS
+        elif self.inventory_index.inventory_index_errors:
+            status = IndexState.FAILURE
         else:
             status = IndexState.SUCCESS
         try:
             self.inventory_index.complete(status=status)
```
ahoying: Remove this line, it's redundant with the update.

Also I think you should expunge self.inventory_index from the session in open() before returning, since we don't want the session to manage inventory_index after that.

And really the class doesn't need a session at all, creating and updating the inventory index and getting the id can all be done via the core interface. Then all the class needs is the inventory_index_id, which is used to write resources and warnings, it no longer needs a pointer to the inventory_index row at all.

joecheuk: Added expunge during inventory index creation, created an issue to remove session from storage (since that might require more works to refactor and higher risk to break stuffs).

joecheuk: #3107

joecheuk: #3107

Issue #3104, opened 2019-08-20
Add Cloud Profiler to Forseti Server


blueandgold:
The sample code here worked for me:
https://cloud.google.com/profiler/docs/profiling-python#using-profiler

And below is what I got from inventory create.  Should add a flag `enable_profiler` to only run the profiler when needed.

![image](https://user-images.githubusercontent.com/6977544/63362403-9bb4c600-c326-11e9-8dd7-e45193b48421.png)

PR #3113, opened 2019-08-20
Add Cloud Profiler to Forseti server

hshin-g:
Addresses #3104.

Below is the output on Cloud Profiler:
Screen Shot 2019-08-20 at 4 06 17 PM

### DIFF
--- a/google/cloud/forseti/services/server.py
+++ b/google/cloud/forseti/services/server.py
@@ -19,10 +19,10 @@ import argparse
 import os
 import sys
 import time
-import googlecloudprofiler
 
 from concurrent import futures
 import grpc
+import googlecloudprofiler
 
 from google.cloud.forseti.common.util import logger
 
@@ -44,6 +44,8 @@ SERVICE_MAP = {
     'model': GrpcModellerFactory,
     'server': GrpcServerConfigFactory
 }
+
+# pylint: disable=too-many-locals
 
 
 def serve(endpoint,
@@ -82,10 +84,10 @@ def serve(endpoint,
         try:
             googlecloudprofiler.start(
                 service='forseti-server',
-                verbose=3
+                verbose=2,
             )
         except (ValueError, NotImplementedError) as exc:
-            LOGGER.warning('Unable to enable Cloud Profiler.', exc)
+            LOGGER.warning('Unable to enable Cloud Profiler: %s', exc)
 
     factories = []
     for service in services:
--- a/google/cloud/forseti/services/server.py
+++ b/google/cloud/forseti/services/server.py
@@ -21,7 +21,6 @@ import sys
 import time
 
 from concurrent import futures
-import googlecloudprofiler
 import grpc
 
 from google.cloud.forseti.common.util import logger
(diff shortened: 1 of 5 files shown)


google/cloud/forseti/services/server.py
```diff
@@ -22,6 +22,7 @@
 
 from concurrent import futures
 import grpc
+import googlecloudprofiler
```
blueandgold: nit: should go before `grpc`?

google/cloud/forseti/services/server.py
```diff
@@ -75,6 +80,15 @@ def serve(endpoint,
     if enable_console_log:
         logger.enable_console_log()
 
+    if enable_profiler:
+        try:
+            googlecloudprofiler.start(
+                service='forseti-server',
+                verbose=2,
```
blueandgold: I think that it would be a good idea, to keep the placeholder for the `project_id`, with a comment that for dev workflow, it can be uncommented and used in the dev environment.
```
# project_id must be set if not running on GCP, such as your local dev environment.
# project_id='my-project-id',
```

google/cloud/forseti/services/server.py
```diff
@@ -75,6 +80,15 @@ def serve(endpoint,
     if enable_console_log:
         logger.enable_console_log()
 
+    if enable_profiler:
```
blueandgold: Hmmm, I think that it would be nicer to start the profiler in the `main()` function.  #1, it keeps the serve() cleaner here and one less parameter to pass & specify, and #2 we can trace other things in `main()` in case they get added in the future. I recommend moving this to right before `serve()` is called in the `main()`

setup.py
```diff
@@ -51,6 +51,8 @@
     'sendgrid==5.6.0',
     'simple-crypt==4.1.7',
     'unicodecsv==0.14.1',
+    # Profiler related.
+    'google-cloud-profiler==1.0.8',
```
blueandgold: Apologies, I just recalled that @ahoying mentioned that instrumentation packages to be optional, so that it's easier for him to port into Google's environment.  @ahoying, please confirm?

If so, the optional dependency can be specified like this in `setup.py`:
image

Then, where you use the import in `server.py`, you can do this try/except:
image

Then when we want to use it, we can do this:
`pip3 install .[profiler]`

I took the screenshots from this PR #2317, where this optionality is implemented, so we should do the same here with the profiler.

ocervell:
Note that the `google-cloud-profiler` package should also be added to `requirements.txt` if you want it to be installed in the VM when the startup script runs. Otherwise, you'll have to add `pip3 install .[profiler]` to all the server install scripts (Terraform, Deployment Manager, Docker, GCP Install Script). 

The `.` notation for relative install also does not work with the current Docker install (pip hangs, see [this](https://github.com/pypa/pip/issues/3532) issue), which is why it's preferable to have it in `requirements.txt` along with build and test libraries.

You should also update the GCP install scripts to enable the Profiler API, and grant the correct role `cloudprofiler.agent` to the server / client service accounts.

Once this is ready, please let the CFT team know so that we can also update the Terraform accordingly.

blueandgold:
LGTM, thanks for making the optional package.

blueandgold:
@hshin-g Can you please respond to @ocervell above?  Let him know what our plan is for this build/release.  Thanks.

hshin-g:
@ocervell Thanks for your comments! In terms of packages being optional, if we add them to the `requirements.txt` file, then that would be installed with the other packages in that file on install (would not be optional at that point), so I have taken out the `google-cloud-profiler` package from the `requirements.txt` file. 

After taking out the package, you will need to run the following commands on the server VM in order to install the package and get Forseti to use it:
```
$ sudo su ubuntu
$ pip3 install google-cloud-profiler
$ sudo systemctl restart forseti
```

For this PR, we have decided to manually enable roles/permissions, so any changes made to Terraform will not be needed until a later time.

ocervell:
@hshin-g LGTM

PR #3114, opened 2019-08-21
Revert "Add storage.objectViewer role to GCP_READ_IAM_ROLES"

kevensen:
Reverts forseti-security/forseti-security#2826


kevensen:
Unfortunately the 'roles/storage.legacyBucketReader' can only be assigned to a bucket, not a service account.

PR #3129, opened 2019-08-22
Added command to remove tmp files prefixed with forseti-cai.

joecheuk:
Not sure if specifying the path here has any impact on the GKE module, @kevensen can you help verify?


install/gcp/scripts/run_forseti.sh
```diff
@@ -21,6 +21,10 @@ source /home/ubuntu/forseti_env.sh
 # they will be logged.
 set -x
 
+# Remove any CAI temp files generated by the forseti process.
+# Failed inventory due to various reasons could cause the temp files not being deleted.
+sudo rm -rf /tmp/forseti-cai-*
```
hshin-g: Are there other /tmp files that could be in here that we would want to remove?

joecheuk: I dont think there are any, these are the newly generated sqlite files only for the new branch.

PR #3147, opened 2019-08-28
Update deployment-prerequisites.md

dekuhn:
Updating gcloud sdk download link to more focused download page.


dekuhn:
Closing I think we need to do a thorough look at the dev setup and determine how we update our instructions.

Issue #3155, opened 2019-08-29
Update Forseti to support Policy Library Sync


gkowalski-google:
To support [Terraform Google Forseti #238](https://github.com/forseti-security/terraform-google-forseti/issues/238), the policy library directory that is used by the Forseti server will be changed. If Policy Library Sync is enabled (which will be controlled by an environment variable set from Terraform), then run the git-sync image as a linux service with docker.

PR #3159, opened 2019-08-30
Add policy library sync

gkowalski-google:
Addresses #3155. These changes help to support the Terraform changes (https://github.com/forseti-security/terraform-google-forseti/issues/238) to allow users to sync the policy library from a public/private Git repo as an alternative to manually copying the files to GCS.

### DIFF
--- a/deployment-templates/compute-engine/server/forseti-instance-server.py
+++ b/deployment-templates/compute-engine/server/forseti-instance-server.py
@@ -246,8 +246,8 @@ gsutil cp gs://{scanner_bucket}/configs/forseti_conf_server.yaml {forseti_server
 gsutil cp -r gs://{scanner_bucket}/rules {forseti_home}/
 
 # Download the Newest Config Validator constraints from GCS
-rm -rf {forseti_home}/policy-library
-gsutil cp -r gs://{scanner_bucket}/policy-library {forseti_home}/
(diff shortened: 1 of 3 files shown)


deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -243,8 +246,8 @@ def GenerateConfig(context):
 gsutil cp -r gs://{scanner_bucket}/rules {forseti_home}/
 
 # Download the Newest Config Validator constraints from GCS
-rm -rf {forseti_home}/policy-library
-gsutil cp -r gs://{scanner_bucket}/policy-library {forseti_home}/
+rm -rf {policy_library_home}
+gsutil cp -r gs://{scanner_bucket}/policy-library {policy_library_home}/
```
kevensen: Can you please help me understand why we are downloading the constraints from GCS?

gkowalski-google: It was my understanding that this file only gets used by the python installer. I did not want to break the existing behavior for users today or add the policy library sync feature to the soon to be deprecated installer. Is that correct or will all this be used by terraform instances as well?

kevensen: Fair enough

PR #3164, opened 2019-09-03
Integration tests

red2k18:


red2k18:
Opened #3202 after restructuring the integration tests.

PR #3168, opened 2019-09-04
Feature/fix init services script

gkowalski-google:
The Policy Library Sync changes were causing an issue with the initialize Forseti services script. The absence of the POLICY_LIBRARY_SYNC_ENABLED environment variable was causing the script to terminate early, and the necessary services were not being configured.


deployment-templates/compute-engine/server/forseti-instance-server.py
```diff
@@ -44,6 +44,7 @@ def GenerateConfig(context):
 
     FORSETI_HOME = '$USER_HOME/forseti-security'
     POLICY_LIBRARY_HOME = '$USER_HOME/policy-library'
+    POLICY_LIBRARY_SYNC_ENABLED = 'false'
```
gkowalski-google: The new sync feature is only being supported from within Terraform. Installations using the deprecated method will have this feature disabled.

Issue #3149, opened 2019-08-28
Update all GitHub references from old repo GoogleCloudPlatform/forseti-security to new org repo.


dekuhn:
We still have a lot of out dated references to https://github.com/GoogleCloudPlatform/forseti-security repo. 

GitHub redirects automatically, but we should replace references in our code to the new org path and repo name.

Action: Update all GitHub references from https://github.com/GoogleCloudPlatform/forseti-security to https://github.com/forseti-security/forseti-security

e.g. https://github.com/forseti-security/forseti-security/search?q=GoogleCloudPlatform%2Fforseti-security&unscoped_q=GoogleCloudPlatform%2Fforseti-security

https://github.com/forseti-security/terraform-google-forseti/search?q=GoogleCloudPlatform%2Fforseti-security&unscoped_q=GoogleCloudPlatform%2Fforseti-security

red2k18:
https://github.com/forseti-security/forseti-security/pull/3182 addresses this.

dekuhn:
@red2k18 Is there a PR to address the Terra-form repo as well? Maybe I missed it?

red2k18:
#3182 updates all the references in forseti-security repo. Will submit another PR to update the references in terraform-google-forseti repo.

red2k18:
Opened https://github.com/forseti-security/terraform-google-forseti/pull/248 to update references in terraform-google-forseti module.

PR #3182, opened 2019-09-09
Updated GitHub references from GoogleCloudPlatform/forseti-security to forseti-security/forseti-security

red2k18:


red2k18:
Fixes #3149

travis/jekyll_build.sh
```diff
@@ -29,7 +29,7 @@ kill %1
 bundle exec htmlproofer --check-img-http --check-html \
 --internal-domains 'forsetisecurity.org' \
 --check-favicon --report-missing-names --report-script-embeds \
---url-ignore '/GoogleCloudPlatform/forseti-security/edit/,/forseti-security/forseti-security/edit/,/maxcdn.bootstrapcdn.com/,/d3js.org/' \
+--url-ignore '/forseti-security/forseti-security/edit/,/forseti-security/forseti-security/edit/,/maxcdn.bootstrapcdn.com/,/d3js.org/' \
```
dekuhn: I don't think this reference needed to be updated. We can either should keep it or if everything is renamed we can just remove it.

gkowalski-google: Agree, it can just be removed since the next URL to be ignored is the forseti-security URL.

red2k18: Removed as everything else is updated.

gkowalski-google: Do we will still need the other url ignores? I thought @dekuhn was just referencing the first url in that list.

Issue #3152, opened 2019-08-28
Error running scanner: KeScanner, KeVersionScanner, FirewallPolicyScanner, and IapScanner


abhishekupadhyaya:
- [x] Fix FirewallPolicyScanner
- [x] Fix IapScanner
- [x] Fix KeScanner (v2.21)
- [x] Fix KeVersionScanner (v2.21)
---

Hi,

We are using **release-2.18.0** in our production system.

We are seeing following errors as part of notification process:

- "serverMessage": "Error running scanner: KeScanner: 'Traceback (most recent call last):\n File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanner.py\", line 119, in run\n scanner.run()\n File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanners/ke_base_scanner.py\", line 216, in run\n ke_clusters = self._retrieve()\n File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanners/ke_base_scanner.py\", line 208, in _retrieve\n parent_type_name=ke_cluster_type_name))[0]\nIndexError: list index out of range\n'"

- "serverMessage": "Error running scanner: KeVersionScanner: 'Traceback (most recent call last):\n File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanner.py\", line 119, in run\n scanner.run()\n File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanners/ke_base_scanner.py\", line 216, in run\n ke_clusters = self._retrieve()\n File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanners/ke_base_scanner.py\", line 208, in _retrieve\n parent_type_name=ke_cluster_type_name))[0]\nIndexError: list index out of range\n'"

[...]

Issue #3106, opened 2019-08-20
Scanner errors when running: "forseti scanner run" command


erikwalker:
- [ ] Fix IapScanner
- [x] Fix KeScanner (v2.21)
- [x] Fix KeVersionScanner (v2.21)
---

Installed forseti with terraform "Terraform v0.11.13"
The version of forseti security that is installed is:
__version__ = '2.18.0'
__package_name__ = 'forseti-security'

Having problem running forseti scanner:

when running "forseti scanner run":

The error output from the client is:

{
  "serverMessage": "Error running scanner: IapScanner: 'Traceback (most recent call last):\n  File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanner.py\", line 119, in run\n    scanner.run()\n  File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanners/iap_scanner.py\", line 682, in run\n    all_violations, _ = self._find_violations(iap_data)\n  File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanners/iap_scanner.py\", line 667, in _find_violations\n    for (iap_resources, project_resource_counts) in iap_data:\n  File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanners/iap_scanner.py\", line 645, in _retrieve\n    instance_templates=instance_templates)\n  File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/scanner/scanners/iap_scanner.py\", line 95, in __init__\n    instance_template_url)\n  File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/common/gcp_type/instance_template.py\", line 158, in from_url\n    url)\n  File \"/usr/local/lib/python3.6/dist-packages/forseti_security-2.18.0-py3.6.egg/google/cloud/forseti/common/gcp_type/key.py\", line 109, in _from_url\n    path_components = urllib.parse.urlparse(url)[2].split('/')[3:]\nTypeError: a bytes-like object is required, not 'str'\n'"
}
{
[...]

PR #3187, opened 2019-09-10
Change hash function for Inventory Resources

gkowalski-google:
Changing the hash function used by size_t_hash() to be idempotent between different runs. The builtin hash() method is salted with a random value determined at the start of each process. This was causing certain resource IDs to change between runs of the inventory.

This was causing issues for some scanners that was run separately from inventory (for example, by running `forseti scanner run` from the CLI), including: KeScanner and KeVersionScanner. This will address the issues found in #3106 and #3152.

I looked into setting the [PYTHONHASHSEED env variable to 0](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHASHSEED), so that the built-in hash function would not include a random salt. However, this has wider security concerns and to be on the safe side, I chose the [xxhash](http://cyan4973.github.io/xxHash/) method which was designed to be a fast non-cryptographic hash algorithm.

### DIFF
--- a/google/cloud/forseti/services/inventory/base/resources.py
+++ b/google/cloud/forseti/services/inventory/base/resources.py
@@ -21,6 +21,7 @@ import ctypes
 from functools import partial
 import json
 import os
+import xxhash
 
 from google.cloud.forseti.common.gcp_api import errors as api_errors
 from google.cloud.forseti.common.util import date_time
@@ -42,7 +43,7 @@ def size_t_hash(key):
     Returns:
         str: The hashed key.
     """
-    return '%u' % ctypes.c_size_t(hash(key)).value
+    return '%u' % ctypes.c_size_t(xxhash.xxh64_intdigest(key)).value
 
 
 def from_root_id(client, root_id, root=True):
--- a/google/cloud/forseti/services/inventory/base/resources.py
+++ b/google/cloud/forseti/services/inventory/base/resources.py
@@ -43,8 +43,7 @@ def size_t_hash(key):
     Returns:
         str: The hashed key.
     """
-    # pylint: disable=no-member
-    hash_digest = hashlib.blake2b(key.encode()).hexdigest()
+    hash_digest = hashlib.blake2b(key.encode()).hexdigest()  # pylint: disable=no-member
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -51,6 +51,7 @@
     'sendgrid==5.6.0',
     'simple-crypt==4.1.7',
     'unicodecsv==0.14.1',
+    'xxhash==1.4.1',
```
joecheuk: Have you looked into the hash functions in hashlib for python? They hash to the same value by default. Installing a new library just for hashing might not be ideal.

```
import hashlib
hashlib.algorithms_available
{'sha3_256', 'shake256', 'sha3_512', 'ripemd160', 'sha512', 'sha384', 'md5', 'shake_256', 'md5-sha1', 'md4', 'whirlpool', 'sha3-512', 'blake2b512', 'sha224', 'shake_128', 'sha1', 'sha3-384', 'blake2b', 'shake128', 'sha3-256', 'sha512-256', 'sha3-224', 'sha256', 'sm3', 'blake2s256', 'sha3_384', 'blake2s', 'sha512-224', 'sha3_224'}
```

gkowalski-google: Updated to use blake2b instead.

google/cloud/forseti/services/inventory/base/resources.py
```diff
@@ -42,7 +43,9 @@ def size_t_hash(key):
     Returns:
         str: The hashed key.
     """
-    return '%u' % ctypes.c_size_t(hash(key)).value
+    # pylint: disable=no-member
+    hash_digest = hashlib.blake2b(key.encode()).hexdigest()
```
joecheuk: Can you put the pylint disable comment beside the line if the length is allowed:

`hash_digest = hashlib.blake2b(key.encode()).hexdigest()  # pylint: disable=no-member`

if not, would be good if you can enable it afterward

```
    # pylint: disable=no-member
    hash_digest = hashlib.blake2b(key.encode()).hexdigest()
    # pylint: enable=no-member
```

blueandgold:
@gkowalski-google great fix here!!

Issue #2922, opened 2019-06-19
Consider Moving to mysql-connector


kevensen:
The current dependency "mysqlclient==1.4.2" relies on C libraries to compile the Python byte code.  The mysql-connector-python library doesn't have an C library dependencies: [https://pypi.org/project/mysql-connector-python/](https://pypi.org/project/mysql-connector-python/)

dekuhn:
@kevensen Can you add some context how this benefits us in a GKE deployment? IIRC relying on the C lib means you had to make some considerations for this dependacy.

kevensen:
I think this would benefit both the GCE and GKE deployments.  The fewer dependencies you install in a host or container, the smaller the surface of attack.

In the case of GKE, it's one less package installed resulting in a smaller container image (every MB makes a difference :-) ).  It's shorter time building the images as well.

kevensen:
If opting for BigQuery as a data store, this is moot.

kevensen:
If I understand correctly, Forseti will continue to use MySQL as a data store.  I think we should continue to  consider this.

kevensen:
This will enable us to use distroless containers.

kevensen:
We implemented PyMySQL.  This can be closed.

PR #3190, opened 2019-09-11
PyMySQL

kevensen:
addresses #2922 

Opted for PyMySQL instead of mysql-connector.  PyMySQL is regularly tested by SQLAlchemy.

They following are initial performance numbers when executed against a resource rich environment.  This suggests comparable execution behavior between the two libraries.

|                  | PyMySQL (seconds) | mysql-client (seconds) |
|------------------|-------------------|------------------------|
| Inventory Create | 192.749           | 224.005                |
| Scanner Run      | 24.468            | 18.467                 |
| Notifier Run     | 2.359             | 2.032                  |


blueandgold:
Nice change.  Like that it simplifies the deps.  I also read the evaluation below, and like that it is popular and well-supported.

https://wiki.openstack.org/wiki/PyMySQL_evaluation

image

kevensen:
Thanks for the review @blueandgold!

Issue #3011, opened 2019-07-31
Implement Q2'19 CAI Resources in Forseti Inventory


dekuhn:
https://cloud.google.com/resource-manager/docs/cloud-asset-inventory/overview

| Resource | Release Date |
| ------------- | ------------- |
compute.googleapis.com/RegionDisk | (2019-07-24) [Already implemented]
compute.googleapis.com/Autoscaler | (2019-07-16) [Already implemented]
compute.googleapis.com/InstanceGroup | (2019-07-16)  [Already implemented]
compute.googleapis.com/InstanceGroupManager | (2019-07-16)  [Already implemented]
compute.interconnect | 2019-06-26 [Already implemented]
compute.interconnect_attachment | 2019-06-26  [Already implemented]
compute.googleapis.com/Address | 2019-06-10 [Already implemented]
compute.googleapis.com/GlobalAddress | 2019-06-10  [Already implemented]
container.googleapis.com/NodePool | 2019-05-01 [WIP Issue ##3010]
compute.globalForwardingRule | 4/22/2019 [Already implemented]
compute.regionBackendService | 4/22/2019 [Already implemented]
bigtable.Cluster | (2019-02-28) [Already implemented]
bigtable.Instance | (2019-02-28) [Already implemented]
bigtable.Table | (2019-02-28) [Already implemented]

gkowalski-google:
@dekuhn OK to close this one out and leave https://github.com/forseti-security/forseti-security/issues/3010 open? I was leaving it open since this is the Epic.

dekuhn:
@gkowalski-google If all other items here are complete we can close the epic since we have the specific issue like you linked to address NodePool once it comes out of beta.

gkowalski-google:
Closing out this issue as the only resource not added, NodePool, is tracked by the [Q3'19 ticket](https://github.com/forseti-security/forseti-security/issues/3012).

PR #3193, opened 2019-09-12
Adding Bigtable Cluster, Instance, and Table resources from CAI.

gkowalski-google:
This supports #3011 to add Q2'19 CAI resources to the Forseti inventory. The only one not implemented yet is the NodePool, which we are waiting for CAI to release as GA.


google/cloud/forseti/services/inventory/base/resources.py
```diff
@@ -1044,6 +1044,31 @@ class BigqueryTable(resource_class_factory('bigquery_table', 'id')):
     """The Resource implementation for bigquery table."""
 
 
+# Bigtable resource classes
+class BigtableCluster(resource_class_factory('bigtable_cluster', 'name',
+                                             hash_key=True)):
```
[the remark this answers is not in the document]
gkowalski-google: The name for a Bigtable Cluster is a long string path like `//bigtable.googleapis.com/projects/project2/instances/test-bigtable/clusters/test-bigtable-c1` so I think it is better to hash this as the resource id. I think for resources where we have an actual globally unique ID (not necessarily a GUID) then we can use those. But for others we should hash, especially for long strings like this. If hash_key is False, then this string would become the resource id.

tests/services/inventory/test_data/mock_cai_resources.dump
```diff
@@ -79,6 +79,9 @@
...
 {"name":"//bigquery.googleapis.com/projects/1042/datasets/bq_test_ds","asset_type":"bigquery.googleapis.com/Dataset", "resource":{"version":"v2","discovery_document_uri":"https://bigquery.googleapis.com/$discovery/rest","discovery_name":"Dataset","parent":"//cloudresourcemanager.googleapis.com/projects/1042","data":{"creationTime":"1540220016171","datasetReference":{"datasetId":"bq_test_ds","projectId":"project2"},"location":"us","id":"project2:bq_test_ds","kind":"bigquery#dataset","lastModifiedTime":"1540220016171"}}}
+{"name":"//bigtable.googleapis.com/projects/project2/instances/test-bigtable/clusters/test-bigtable-c1","asset_type":"bigtableadmin.googleapis.com/Cluster","resource":{"version":"v2","discovery_document_uri":"https://bigtableadmin.googleapis.com/$discovery/rest","discovery_name":"Cluster","parent":"//bigtable.googleapis.com/projects/project2/instances/test-bigtable","data":{"defaultStorageType":"SSD","location":"projects/project2/locations/us-west1-a","name":"projects/project2/instances/test-bigtable/clusters/test-bigtable-c1","state":"READY"}}}
```
red2k18: Please update the project_id based on https://github.com/forseti-security/forseti-security/blob/dev/tests/services/inventory/test_data/mock_gcp_results.py#L5979 for mock data.

gkowalski-google: Chatted with @red2k18 offline about this. The inventory tests for CAI data are using different project Ids then the API tests and this would require making more changes.

PR #3202, opened 2019-09-13
[PoC] Continuous Integration and Release Automation

red2k18:


.travis.yml
```diff
@@ -1,38 +1,32 @@
-# Copyright 2017 The Forseti Security Authors. All rights reserved.
```
dekuhn: Where did the license / disclaimer go?

dekuhn:
@red2k18 will we need to keep in sync the terraform deployment files in sync with those in https://github.com/forseti-security/terraform-google-forseti?

.travis.yml
red2k18: Was removed in error. Will add it back.

red2k18:
@dekuhn the terraform configuration related to automating test cases need not be in `terraform-google-forseti` module. So we do not need to keep the terraform configuration in both the repositories in sync.

integration/fixtures/bastion/main.tf
```diff
@@ -0,0 +1,82 @@
+/**
```
blueandgold: `integration` as a directory name is not clear what kind of integration is being referred to.  It would be awesome and much more clear, if this directory is actually named `integration_tests` as in the doc when we considered the options.

.kitchen.yml
```diff
@@ -0,0 +1,46 @@
...
+verifier:
+  name: terraform
+  color: false
+  fail_fast: false
+
+suites:
+  - name: forseti_integration_testing
+    driver:
+      name: terraform
+      root_module_directory: integration/fixtures/forseti_integration_testing
```
blueandgold: Seems like we don't need to have a specific `forseti_integration_testing` reference here, as there is only one fixture here.

blueandgold: So, the root module should be `integration_tests/fixtures`.

.kitchen.yml
```diff
@@ -0,0 +1,46 @@
...
+      command_timeout: 1800
+    lifecycle:
+      pre_verify:
+        - chmod go-rwx integration/fixtures/bastion/tls_private_key
```
blueandgold: What is the `bastion` used for?

.kitchen.yml
```diff
@@ -0,0 +1,46 @@
...
+    verifier:
+      systems:
+        - name: inventory-module
+          backend: ssh
+          controls:
+            - inventory
+          hosts_output: forseti-server-vm-ip
```
blueandgold: Why does this point to the server vm?  How about the client vm?

.travis.yml
```diff
@@ -13,26 +12,34 @@
...
+    - docker
+    before_install:
+    - "./install/scripts/docker_setup_forseti.sh"
+    install:
+    -
+    script:
+    - "./install/scripts/docker_run_forseti.sh"
+    - "./install/scripts/docker_unittest_forseti.sh"
+    - "./install/scripts/docker_pystyle_forseti.sh"
+    after_success:
```
blueandgold: nit: add the comment as before

.travis.yml
```diff
@@ -1,4 +1,3 @@
-# Copyright 2017 The Forseti Security Authors. All rights reserved.
 #
```
dekuhn: Was there a reason you removed the Copyright line??
e.g.
# Copyright 2019 The Forseti Security Authors. All rights reserved.

.travis.yml
```diff
@@ -13,26 +12,34 @@
...
+  - stage: Integration test
+    name: Integration test - docker
+    services:
+    - docker
+    before_install:
+    - openssl aes-256-cbc -K $encrypted_5c0ae2aa862c_key -iv $encrypted_5c0ae2aa862c_iv
+      -in secret-key.tar.enc -out secret-key.tar -d
+    - tar xvf secret-key.tar
+    install:
+    - docker image pull gcr.io/cloud-foundation-cicd/cft/developer-tools:0.2.0
```
blueandgold: Why static reference to a specific version?  What happens the images is updated?  Should we just use the latest?

.travis.yml
```diff
@@ -13,26 +12,34 @@
...
+    script:
+    - docker container run -it -e KITCHEN_TEST_BASE_PATH="integration/tests" -e SERVICE_ACCOUNT_JSON="$(< release-silver.json)"
+      --env-file travis-integration.env -v $(pwd):/workspace gcr.io/cloud-foundation-cicd/cft/developer-tools:0.2.0
```
blueandgold: ditto on version

integration/fixtures/forseti_integration_testing/main.tf
```diff
@@ -0,0 +1,110 @@
...
+      private_key         = "${tls_private_key.main.private_key_pem}"
+      bastion_host        = "${module.bastion.host}"
+      bastion_port        = "${module.bastion.port}"
+      bastion_private_key = "${module.bastion.private_key}"
+      bastion_user        = "${module.bastion.user}"
+    }
+  }
+}
+
+module "integration-test-project-1" {
```
blueandgold: Can we drop the `1` here?  Even if there are more projects, we should use descriptive naming, rather than a number.

.travis.yml
```diff
@@ -13,26 +12,34 @@
...
-  - ./install/scripts/docker_run_forseti.sh
-  - ./install/scripts/docker_unittest_forseti.sh
-  - ./install/scripts/docker_pystyle_forseti.sh
-
-after_success:
-  # Collect codecov results.
-  - docker -l error exec -it build /bin/bash -c "bash <(curl -s https://codecov.io/bash)"
+jobs:
+  include:
+  - stage: Integration test
```
blueandgold: Can you please explain having different jobs here?  Will the Integration test run when a PR is created or updated?

integration/fixtures/forseti_integration_testing/scripts/.gitignore
```diff
@@ -0,0 +1,2 @@
+terraform.tfvars
+sshkey
```
blueandgold: nit: add newline

integration/fixtures/forseti_integration_testing/main.tf
```diff
@@ -0,0 +1,110 @@
...
+  source  = "terraform-google-modules/project-factory/google"
+  version = "~> 3.0"
+
+  billing_account = "${var.billing_account}"
+  name = "integration-1"
```
blueandgold: ditto: let's have a more descriptive name, rather than using a number.  How about `integration-test`?

integration/fixtures/forseti_integration_testing/scripts/wait-for-forseti.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "Forseti is ready."
+    echo "Waiting for startup script to complete"
+    sleep 60
+    exit 0
+  else
+    sleep 1
+  fi
+done
+
+echo "Forseti was not ready after 360 seconds!"
```
blueandgold: Since `360` is used several times, can you put it into a variable, so that it's easy to change in the future?

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+            command("forseti inventory purge 0").result
+        end
+    end
+
+    describe "Create an inventory and check database" do
+        subject do
+            command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT resource_id) from gcp_inventory where category='resource' and resource_type = 'project';\"")
+        end
+
+        before(:context) do
```
blueandgold: Why the use of `(:context)` here?  It's not used elsewhere.

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+            command("forseti inventory create").result
+            command("sudo apt-get -y install mysql-client").result
```
blueandgold: This setup seems like it should not be part of a test.  Can it be moved elsewhere?

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+        end
+
+        # its("stdout") { should match "2" }
```
blueandgold: Why is this commented out?  It's the check that we need.  Also, can it not be hardcoded to a number.  Instead can the expected count can be determined from the source config used to populate the environment?

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+        its("stderr") { should eq ""}
```
blueandgold: Why checking on `stderr` here?

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+    describe "Purge inventory" do
+        subject do
+            command("forseti inventory purge 0")
+        end
+
+        before do
+            command("forseti inventory create").result
+        end
+
+        its("stdout") { should match /purged/ }
```
blueandgold: The database needs to be checked that the inventory data is actually purged.

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+    describe "List inventory" do
+        subject do
+            command("forseti inventory list")
+        end
+
+        before do
+            command("forseti inventory create").result
+        end
+
+        its("stderr") { should eq ""}
```
blueandgold: ditto on checking the database, rather than just there is no error

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+        subject do
+            command("forseti inventory get #{inventory_id}")
+        end
+
+        let :inventory_id do
+            command("forseti inventory create").result
+            JSON.parse(command("forseti inventory list").stdout).fetch("id")
+        end
+
+        its("stderr") { should eq ""}
```
blueandgold: ditto on better checking

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+        subject do
+            command("forseti inventory delete #{inventory_id}")
+        end
+
+        let :inventory_id do
+            command("forseti inventory create").result
+            JSON.parse(command("forseti inventory list").stdout).fetch("id")
+        end
+
+        its("stderr") { should eq ""}
```
blueandgold: ditto on better checking, please address this everywhere

integration/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,426 @@
...
+        subject do
+            command("forseti model create --inventory_index_id #{inventory_id} test_model")
+        end
+
+        let :inventory_id do
+            command("forseti inventory create").result
+            JSON.parse(command("forseti inventory list").stdout).fetch("id")
+        end
+
+        its("stdout") { should match /SUCCESS/ }
```
blueandgold: ditto on checking the database that the data model has been created successfully: we have the right tables, and we have the right data in the tables

.kitchen.yml
red2k18: Consider it done

.kitchen.yml
red2k18: `bastion` is used to SSH into the server VM.

integration/fixtures/bastion/main.tf
red2k18: Renamed.

integration/fixtures/forseti_integration_testing/scripts/.gitignore
red2k18: Added

.travis.yml
red2k18: Done

.travis.yml
red2k18: Nope, I intended to modify only the contents. Added it back.

integration/fixtures/forseti_integration_testing/main.tf
red2k18: Yes, updated.

integration/fixtures/forseti_integration_testing/main.tf
red2k18: Sure, updated.

integration/fixtures/forseti_integration_testing/scripts/wait-for-forseti.sh
red2k18: Done.

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: `:context` is used in all the tests where there's more than one assertion.

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: This is required to SSH into the database to check the count of the resources in the org.

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: This is commented out at the moment so that test doesn't fail as new resources are added for testing. This will change the count. Not sure if the count can be determined from the source config, but can check on this.

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: This is a way to have least number of assertions while still ensuring the test case is properly checked. If there are no errors that means all the assertions are correct.

dekuhn:
@red2k18 PR is failing deployment due to Forseti timeout [see travis log](https://travis-ci.org/forseti-security/forseti-security/jobs/585783983#L868)

red2k18:
@dekuhn I noticed that. Forseti server is taking variable amount of time to start up and that's why build is failing.

.kitchen.yml
red2k18: `hosts_output` is set to server VM as the existing tests are executed in server VM. Client VM can be added instead of/in addition to the server VM.

.travis.yml
red2k18: I think we should point to the docker image that works for us and not use the latest one as there might be breaking changes. Also, I don't think there is a `latest` tag.

.travis.yml
red2k18: I think we should point to the docker image that works for us and not use the latest one as there might be breaking changes. Also, I don't think there is a latest tag

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Added a test to check the database to see if inventory data is purged.

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Added a test to check the database to see if inventory data is created.

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Added a test to check the database to see if inventory is gotten.

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Added a test to check the database to see if inventory is deleted.

.travis.yml
red2k18: Updated `.travis.yml`. PTAL. Integration tests will not run for every PR. Newly added check was not included before for testing of this PR on Travis.

red2k18:
@blueandgold Addressed all the comments except database check for model.

.travis.yml
```diff
@@ -13,26 +12,34 @@
...
-
-script:
-  - ./install/scripts/docker_run_forseti.sh
-  - ./install/scripts/docker_unittest_forseti.sh
-  - ./install/scripts/docker_pystyle_forseti.sh
-
-after_success:
-  # Collect codecov results.
-  - docker -l error exec -it build /bin/bash -c "bash <(curl -s https://codecov.io/bash)"
+jobs:
```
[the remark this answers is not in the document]
red2k18: Stages are already used and is defining the order of execution. Currently integration tests are executed before unit tests.

.kitchen.yml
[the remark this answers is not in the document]
red2k18: tls_private_key and ssh key are generated dynamically and are used to SSH into the server VM.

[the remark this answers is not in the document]
red2k18: `tls_private_key` is located at `integration/fixtures/bastion/tls_private_key`. You will be able to see it when you run it locally in the dev environment. For security reasons, this key is not saved and is generated dynamically upon running `kitchen converge`.

integration_tests/fixtures/forseti_integration_testing/main.tf
```diff
@@ -0,0 +1,110 @@
+/**
```
blueandgold: Please shorten the path here to `integration_tests/fixtures/testing/main.tf`.  Because this is a testing fixture.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,493 @@
...
+        subject do
+            command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT resource_id) from gcp_inventory where category='resource' and resource_type = 'project';\"")
+        end
+
+        before(:context) do
+            command("forseti inventory create").result
+            command("sudo apt-get -y install mysql-client").result
+        end
+
+        # its("stdout") { should match "2" }
```
blueandgold: The count here need to be determined from the source config.  Where is the source config where all the resources are defined?

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,493 @@
...
+    end
+
+    describe "Create an inventory and check database" do
+        subject do
+            command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT resource_id) from gcp_inventory where category='resource' and resource_type = 'project';\"")
+        end
+
+        before(:context) do
+            command("forseti inventory create").result
+            command("sudo apt-get -y install mysql-client").result
```
blueandgold: Move the installation of `mysql-client` to the setup part.  Let's not have this installed in the test, which is being done multiple times.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,493 @@
+require 'json'
+control 'inventory' do
+
+     describe "Create an inventory" do
```
blueandgold: Can this this test should be combined with `Create an inventory and check database` below?  Can we have multiple `do` commands and assert on them within one test?  Otherwise, the tests here are very repetitive.

For example why have a separate purge test where inventory is created again, if we can test the purge command here?

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,493 @@
...
+            command("forseti inventory create").result
+        end
+
+        its("stdout") { should match /purged/ }
+
+    end
+
+     describe "Purge inventory database check" do
+        subject do
+            command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select * from inventory_index;\"")
```
blueandgold: We also need to check the gcp_inventory table is also empty.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,493 @@
...
+        end
+
+        its("stdout") { should match "" }
+
+        after do
+            command("forseti inventory purge 0").result
+        end
+    end
+
+    describe "Create inventory and model" do
```
blueandgold: It feels the model tests should be moved to a separate file.  Thoughts?  Would make the current file shorter and more focused on inventory tests.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
```diff
@@ -0,0 +1,493 @@
...
+            command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select * from inventory_index;\"")
+        end
+
+        let :inventory_id do
+            command("forseti inventory create").result
+            JSON.parse(command("forseti inventory list").stdout).fetch("id")
+            command("forseti inventory delete #{inventory_id}").result
+        end
+
+        its("stdout") { should match "" }
```
blueandgold: For this to be a valid test, at least a couple of inventories needs to be created.  So, when you delete one, you can assert that one is still remaining in the index and gcp_inventory tables.  This way, we can test that the delete is not simply purging everything.

integration_tests/fixtures/forseti_integration_testing/scripts/.gitignore
```diff
@@ -0,0 +1,2 @@
+terraform.tfvars
+sshkey
```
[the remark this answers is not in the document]
red2k18: Done

integration_tests/fixtures/forseti_integration_testing/scripts/wait-for-forseti.sh
```diff
@@ -0,0 +1,35 @@
...
+    echo "Waiting for startup script to complete"
+    sleep 60
+    exit 0
+  else
+    sleep 1
+  fi
+done
+
+echo "Forseti was not ready after $MAX seconds!"
+exit 1
```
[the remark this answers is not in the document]
red2k18: Done

integration_tests/fixtures/forseti_integration_testing/scripts/wait-for-forseti.sh
```diff
@@ -0,0 +1,35 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Block until the Forseti startup script has finished running.
+
+MAX=900
+
+echo "Waiting for up to $MAX seconds for Forseti to be ready."
```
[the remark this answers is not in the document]
red2k18: Renamed variable.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Updated SQL query to check both the tables.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Tests have been rewritten to create inventory / model once at the beginning and make assertions on that to test inventory and model Forseti commands.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Done.

integration_tests/fixtures/forseti_integration_testing/scripts/wait-for-forseti.sh
```diff
@@ -0,0 +1,35 @@
...
+
+for _ in $(seq 1 $MAX); do
+  if [[ -f /etc/profile.d/forseti_environment.sh ]]; then
+    echo "Forseti is ready."
+    echo "Waiting for startup script to complete"
```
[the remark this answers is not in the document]
red2k18: Updated `echo` commands.

.kitchen.yml
```diff
@@ -0,0 +1,46 @@
...
+verifier:
+  name: terraform
+  color: false
+  fail_fast: false
+
+suites:
+  - name: forseti_integration_testing
+    driver:
+      name: terraform
+      root_module_directory: integration_tests/fixtures/forseti_integration_testing
```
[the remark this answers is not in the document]
red2k18: Done

integration_tests/fixtures/forseti_integration_testing/main.tf
```diff
@@ -0,0 +1,110 @@
...
+    always_run = "${uuid()}"
+  }
+
+  provisioner "remote-exec" {
+    script = "${path.module}/scripts/wait-for-forseti.sh"
+
+    connection {
+      type                = "ssh"
+      user                = "ubuntu"
+      host                = "${module.forseti-install-simple.forseti-server-vm-ip}"
```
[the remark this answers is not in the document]
red2k18: `forseti-install-simple` module is in the same file (`main.tf`) on line 43.

integration_tests/fixtures/forseti_integration_testing/main.tf
```diff
@@ -0,0 +1,110 @@
...
+    always_run = "${uuid()}"
+  }
+
+  provisioner "remote-exec" {
+    script = "${path.module}/scripts/wait-for-forseti.sh"
+
+    connection {
+      type                = "ssh"
+      user                = "ubuntu"
+      host                = "${module.forseti-install-simple.forseti-client-vm-ip}"
```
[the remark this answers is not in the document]
red2k18: forseti-client-vm-ip is gotten from the Outputs of `terraform-google-forseti` module that is described as the source of `forseti-install-simple` module on line 44.

integration_tests/fixtures/forseti_integration_testing/main.tf
```diff
@@ -0,0 +1,110 @@
...
+      private_key         = "${tls_private_key.main.private_key_pem}"
+      bastion_host        = "${module.bastion.host}"
+      bastion_port        = "${module.bastion.port}"
+      bastion_private_key = "${module.bastion.private_key}"
+      bastion_user        = "${module.bastion.user}"
+    }
+  }
+}
+
+resource "null_resource" "wait_for_client" {
```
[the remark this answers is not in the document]
red2k18: `wait_for_server` and `wait_for_client` cannot be consolidated as these are two different resources.

integration_tests/fixtures/forseti_integration_testing/scripts/wait-for-forseti.sh
```diff
@@ -0,0 +1,35 @@
...
+    sleep 60
+    exit 0
+  else
+    sleep 1
```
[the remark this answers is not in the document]
red2k18: Updated to sleep 10 seconds.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Sounds good to me. Isolated tests per module.

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Done

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Resources are defined in the`main.tf`.

integration_tests/fixtures/forseti_integration_testing/main.tf
red2k18: Renamed to `integration_tests/fixtures/forseti/main.tf` as it needs to match with `integration_tests/fixtures/forseti`.

.travis.yml
```diff
@@ -13,26 +13,36 @@
...
+  include:
+  - stage: Integration test
+    name: Integration test - docker
+    if: env(TRAVIS_EVENT_TYPE) = cron
```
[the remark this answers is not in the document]
red2k18: There is a `cron` job that runs unit tests daily. We are leveraging the same cron job to fulfill the requirement to run integration tests daily and so the check on `cron` type.

.kitchen.yml
red2k18: Sure thing

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Updated test to check the database.

red2k18:
> > @dekuhn the terraform configuration related to automating test cases need not be in `terraform-google-forseti` module. So we do not need to keep the terraform configuration in both the repositories in sync.
> 
> How will we test terra-form changes with the automated testing?

Terraform changes are tested in `terraform-google-forseti` module as a new Terraform module needs to be released with every new version of Forseti. For integration testing, Terraform configuration is only used to create new resources and for basic wiring. For infrastructure purpose (to SSH into the database) and to deploy Forseti, we're using the Terraform configuration that already exists and is tested.

.travis.yml
```diff
@@ -13,26 +12,34 @@
...
+    before_install:
+    - openssl aes-256-cbc -K $encrypted_5c0ae2aa862c_key -iv $encrypted_5c0ae2aa862c_iv
+      -in secret-key.tar.enc -out secret-key.tar -d
+    - tar xvf secret-key.tar
+    install:
+    - docker image pull gcr.io/cloud-foundation-cicd/cft/developer-tools:0.2.0
+    script:
+    - docker container run -it -e KITCHEN_TEST_BASE_PATH="integration/tests" -e SERVICE_ACCOUNT_JSON="$(< release-silver.json)"
+      --env-file travis-integration.env -v $(pwd):/workspace gcr.io/cloud-foundation-cicd/cft/developer-tools:0.2.0
+      /usr/local/bin/test_integration.sh
```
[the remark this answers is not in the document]
red2k18: Yes, it lives in the container. Comment can be added. 

It is doing all the 3 tasks - deploys Forseti, runs tests and deconstructs the environment.

red2k18:
@dekuhn CFT doesn't publish `latest` tag, However, I'm using docker image version 0.2.0. We do not need to update with every new version that is released as there might be breaking changes and we might not need the new features and the 0.2.0 version meets our requirements.

When we do need to update the version, changes to update the version needs to be made in `.travis.yml` only.

red2k18:
@blueandgold @dekuhn I have addressed all the comments.

dekuhn:
> @dekuhn the terraform configuration related to automating test cases need not be in `terraform-google-forseti` module. So we do not need to keep the terraform configuration in both the repositories in sync.

@red2k18 My question here is about how we test Terraform related changes. I am not clear how Terraform changes will or can be tested with the current way the integration tests are being executed. Can you please outline how these types of changes are covered with the integration tests or not?

integration_tests/tests/forseti_integration_testing/controls/inventory.rb
dekuhn: @red2k18 with Forseti on GKE we are using the Python 3 implementation of MySQL; can we use that here as well? e.g. https://github.com/forseti-security/forseti-security/pull/3190

red2k18:
> > @dekuhn the terraform configuration related to automating test cases need not be in `terraform-google-forseti` module. So we do not need to keep the terraform configuration in both the repositories in sync.
> 
> @red2k18 My question here is about how we test Terraform related changes. I am not clear how Terraform changes will or can be tested with the current way the integration tests are being executed. Can you please outline how these types of changes are covered with the integration tests or not? For example will we need to setup another implementation of the test framework to test Terraform changes in the `terraform-google-forseti repo`?

Terraform related changes are tested only in `terraform-google-forseti` module and not in this PR that aims to automate release and implement continuous integration. No, we do not need to setup another implementation of the test framework.

Infrastructure code used in this PR is already tested in the `terraform-google-forseti` repo.

.kitchen.yml
```diff
@@ -0,0 +1,46 @@
...
+    verifier:
+      systems:
+        - name: inventory-module
+          backend: ssh
+          controls:
+            - inventory
+          hosts_output: forseti-server-vm-ip
+          user: ubuntu
+          key_files:
+            - integration/fixtures/forseti_integration_testing/sshkey
```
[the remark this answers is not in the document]
dekuhn: @red2k18 @aaron-lane I am trying to audit and understand how the SSH is setup and where this is configured. Can you point me to the configuration that defines that this is created dynamically? Should I see this in any output log? I did initial research but did not find anything definitive. Thanks!

dekuhn:
> @dekuhn CFT doesn't publish `latest` tag, However, I'm using docker image version 0.2.0. We do not need to update with every new version that is released as there might be breaking changes and we might not need the new features and the 0.2.0 version meets our requirements.
> 
> When we do need to update the version, changes to update the version needs to be made in `.travis.yml` only.

I am not sure why you are telling me this? I added this in response to @blueandgold comment 6 days ago that the CFT team does not use a latest tag and this is the version explicitly defined in the travis.yml. I have no concerns here.

.kitchen.yml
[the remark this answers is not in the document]
red2k18: https://github.com/forseti-security/forseti-security/blob/continuous-integration/integration_tests/fixtures/bastion/outputs.tf#L27

https://github.com/forseti-security/forseti-security/blob/continuous-integration/integration_tests/fixtures/forseti/main.tf

PTAL

.travis.yml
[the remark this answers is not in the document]
dekuhn: What is the rationale for running the integration tests before the unit tests? Would you not want it the other way around?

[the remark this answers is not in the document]
red2k18: Integration tests were temporarily executed before the unit tests to speed up the development and testing process. Now that I have completed work on this, I have switched the order of execution of tests.

https://travis-ci.org/forseti-security/forseti-security/jobs/589459833

.kitchen.yml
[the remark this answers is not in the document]
dekuhn: Thanks I was able to confirm how and the key size. Thanks for the TF file links etc.

.kitchen.yml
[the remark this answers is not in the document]
dekuhn: Thanks I was able to confirm how and the key size. Thanks for the TF file links etc.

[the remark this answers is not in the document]
dekuhn: Thanks I was able to confirm how and the key size. Thanks for the TF file links etc.

.kitchen.yml
red2k18: Done 9 days ago

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Moved `mysql-client` installation to the top so it can be used by all the tests.

integration/tests/forseti_integration_testing/controls/inventory.rb
red2k18: Elaborating on that - `stderr` ensure there are no errors in the standard output. Database check confirms this.

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,95 @@
+require 'json'
+control 'inventory' do
+
+    describe "Inventory" do
+        before :context do
+            command("forseti inventory purge 0").result
+            command("forseti inventory create").result
+            command("sudo apt-get -y install mysql-client").result
```
blueandgold: I have asked this before, please move this install of mysql-client into the vm setup, and outside of the test.

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,95 @@
...
+                expect(command("forseti inventory list").stdout).to match /SUCCESS/
+            end
+
+            it "should be visible in the database" do
+                expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select COUNT(DISTINCT gcp_inventory.inventory_index_id) from gcp_inventory join inventory_index ON inventory_index.id = gcp_inventory.inventory_index_id;\"").stdout).to match /1/
+            end
+
+             # Below assertion should be uncommented after populating test data.
+             # it "should be visible in the database" do
+             #    expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT resource_id) from gcp_inventory where category='resource' and resource_type = 'project';\"").stdout).to match /1/
```
blueandgold: Before we merge, I like to see that we can assert the actual result in the database, against the test resources that were defined to be populated.  i.e. if we define N projects to be created in the test environment, I want to ensure the N projects are in the inventory database.

integration_tests/tests/forseti/controls/model.rb
```diff
@@ -0,0 +1,67 @@
...
+    describe "Model" do
+        before :context do
+            command("forseti inventory purge 0").result
+            command("forseti inventory create").result
+
+            # This variable cannot be used after all the inventories have been purged..
+            @inventory_id = JSON.parse(command("forseti inventory list").stdout).fetch("id")
+            command("forseti model create --inventory_index_id #{@inventory_id} model_new").result
+            command("forseti model use model_new").result
+            command("sudo apt-get -y install mysql-client").result
```
blueandgold: ditto: this should not be in the test, but should be rather in the setup of the VM.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: I have moved `mysql-client` installation at the top so it is executed only once for all the tests. I will look into moving this outside of the test file.

integration_tests/tests/forseti/controls/model.rb
red2k18: I have moved mysql-client installation at the top so it is executed only once for all the tests. I will look into moving this outside of the test file.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: I have uncommented this check. Current count of the projects in the database is high at this time as project was created for test purpose every time; I have removed that logic. A lot of projects are in the pending delete state at this moment and so they are reflected when the query in question is executed. It is likely that the test will fail when the projects are deleted. The count needs to be updated at that time.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: Done

integration_tests/tests/forseti/controls/model.rb
red2k18: Done

red2k18: You can check out the build logs here -  https://travis-ci.org/forseti-security/forseti-security/jobs/590150397

integration_tests/tests/forseti/controls/inventory.rb
red2k18: You can check out the build logs here -  https://travis-ci.org/forseti-security/forseti-security/jobs/590150397

integration_tests/tests/forseti/controls/inventory.rb
blueandgold: Deleted projects should be able to be filtered out by the lifecycle state.

Even beyond projects, this would be very useful other resources.  e.g. if i defined 10 buckets to be created in my test inv, is there really 10 buckets in the database.

red2k18: Done

integration_tests/fixtures/bastion/main.tf
```diff
@@ -0,0 +1,83 @@
...
+  network = "${var.network}"
+
+  allow {
+    protocol = "tcp"
+    ports    = ["22"]
+  }
+
+  direction     = "INGRESS"
+  priority      = "100"
+  source_ranges = ["0.0.0.0/0"]
```
blueandgold: Should access here be narrowed to travis's IP range?  https://docs.travis-ci.com/user/ip-addresses/

integration_tests/fixtures/forseti/.gitignore
```diff
@@ -0,0 +1,3 @@
+terraform.tfvars
```
blueandgold: Why have a local `.gitignore` file here?  Is there any consideration why this should not be in the top-level `.gitignore` file?

integration_tests/fixtures/forseti/outputs.tf
```diff
@@ -0,0 +1,84 @@
...
+output "suffix" {
+  description = "The random suffix appended to Forseti resources"
+  value       = "${module.forseti-install-simple.suffix}"
+}
+output "forseti-cai-storage-bucket" {
+  description = "Forseti CAI storage bucket"
+  value       = "${module.forseti-install-simple.forseti-cai-storage-bucket}"
+}
+
+output "resources_names" {
```
blueandgold: It might be better to call this `kms_resources_names`, this matches your description.  Think it's also good to output grouping of resources by their types, so it's more usable in the test.

integration_tests/fixtures/forseti/scripts/wait-for-forseti.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# Block until the Forseti startup script has finished running.
+
+MAX_WAIT=900
+
+echo "Waiting for up to $MAX_WAIT seconds for Forseti to be ready."
+
+for _ in $(seq 1 $MAX_WAIT); do
+  if [[ -f /etc/profile.d/forseti_environment.sh ]]; then
+    echo "Forseti is not ready yet"
```
blueandgold: Just to double check, what is the state here?  Forseti is not ready?  Then why `exit 0`?

main.tf
```diff
@@ -0,0 +1,245 @@
+/**
```
blueandgold: Why is this `main.tf` checked into the top-level directory?  Why not in the `integration_tests` directory?

.kitchen.yml
```diff
@@ -0,0 +1,45 @@
+# Copyright 2018 Google LLC
```
blueandgold: Why is this `.kitchen.yml` checked into the top-level directory?  Why not in the `integration_tests` directory?

main.tf
blueandgold: Why do we need a separate `main.tf` file here? How does this reconcile with the `main.tf` in the `terraform-google-forseti` repo?  What if there are changes in `terraform-google-forseti` repo, then do we have to manually keep this up to date?

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,375 @@
...
+control 'explain' do
+
+    describe "Explain" do
+        before :context do
+            command("forseti inventory purge 0").result
+            command("forseti inventory create").result
+            inventory_id = JSON.parse(command("forseti inventory list").stdout).fetch("id")
+            command("forseti model create --inventory_index_id #{inventory_id} model_new").result
+            command("forseti model use model_new").result
+            command("sudo apt-get -y install mysql-client").result
```
blueandgold: Installing `mysql-client` should not be here.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,375 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+require 'json'
+control 'explain' do
```
blueandgold: Can you show me what is invoking these explain tests?

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,375 @@
...
+        end
+
+        describe "List IAM roles" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_roles --prefix roles/iam").stdout).to match /roles\/iam.organizationRoleAdmin/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_roles --prefix roles/iam").stdout).to match /roles\/iam.organizationRoleViewer/
```
blueandgold: Can all the `list_role`  matching be consolidated?  Wouldn't the output be all the roles instead of just one?

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,375 @@
...
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_permissions --roles roles/iam.roleAdmin").stderr).to eq ""
+            end
+        end
+
+        describe "List IAM storage.Admin permissions" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_permissions --roles roles/storage.Admin").stdout).to match /firebase.projects.get/
```
blueandgold: Can all the `list_permissions`  matching be consolidated?  Wouldn't the output be all the permissions instead of just one?

integration_tests/fixtures/forseti/outputs.tf
blueandgold: Make sure you update where this is used in the test.

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,115 @@
...
+
+             it "should be visible in the database" do
+                 expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"SELECT count(DISTINCT resource_type) from gcp_inventory where resource_type in ('kms_cryptokey', 'kms_keyring');\"").stdout).to match /#{resources_names.count}/
+             end
+        end
+
+        describe "List an inventory" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti inventory list").stderr).to eq ""
```
blueandgold: Why check on `stderr` vs checking on `stdout`?  We expect the list command to produce an output here, right?  So should check on the output.

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,115 @@
...
+
+            it "should be visible in the database" do
+                expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT gcp_inventory.inventory_index_id) from gcp_inventory join inventory_index on inventory_index.id = gcp_inventory.inventory_index_id where inventory_index.id = #{@inventory_id};\"").stdout).to match /1/
+            end
+        end
+
+        describe "Get an inventory" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti inventory get #{@inventory_id}").stderr).to eq ""
```
blueandgold: Why check on `stderr` vs checking on `stdout`?  We expect the `get` command to produce an output here, right?  So should check on the output.

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,115 @@
...
+                command("forseti inventory create").result
+            end
+
+            # Displays number of inventories in the database before purging inventories.
+            it "should be visible in the database" do
+                expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT gcp_inventory.inventory_index_id) from gcp_inventory join inventory_index ON inventory_index.id = gcp_inventory.inventory_index_id;\"").stdout).to match /2/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti inventory delete #{@inventory_id}").stderr).to match ""
```
blueandgold: ditto, should check on `stdout`

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,115 @@
...
+                expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT gcp_inventory.inventory_index_id) from gcp_inventory join inventory_index ON inventory_index.id = gcp_inventory.inventory_index_id;\"").stdout).to match /2/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti inventory purge 0").stdout).to match /purged/
+            end
+
+            # Displays number of inventories in the database after purging inventories.
+            it "should be visible in the database" do
+                expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT gcp_inventory.inventory_index_id) from gcp_inventory join inventory_index ON inventory_index.id = gcp_inventory.inventory_index_id;\"").stdout).to match /0/
```
blueandgold: We should also add an expect for the inventory index table, that the rows there are gone.

integration_tests/tests/forseti/controls/model.rb
```diff
@@ -0,0 +1,82 @@
...
+            # This variable cannot be used after all the inventories have been purged..
+            @inventory_id = JSON.parse(command("forseti inventory list").stdout).fetch("id")
+            command("forseti model create --inventory_index_id #{@inventory_id} model_new").result
+            command("forseti model use model_new").result
+        end
+
+        describe "Create and get a model" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti model get model_new").stderr).to eq ""
```
blueandgold: Check on `stdout`

integration_tests/tests/forseti/controls/model.rb
```diff
@@ -0,0 +1,82 @@
...
+
+            it "should be visible in the database" do
+                expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select state from model where name = 'model_new';\"").stdout).to match /PARTIAL_SUCCESS/
+            end
+        end
+
+        describe "List a model" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti model list model_new").stderr).to eq ""
```
blueandgold: ditto on `stdout`

integration_tests/tests/forseti/controls/model.rb
```diff
@@ -0,0 +1,82 @@
...
+            end
+        end
+
+        describe "Use a model" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti model use model_new").stderr).to eq ""
```
blueandgold: ditto on `stdout`

.travis.yml
```diff
@@ -13,26 +13,37 @@
...
+    script:
+    - "./install/scripts/docker_run_forseti.sh"
+    - "./install/scripts/docker_unittest_forseti.sh"
+    - "./install/scripts/docker_pystyle_forseti.sh"
+    after_success:
+    - # Collect codecov results.
+    - docker -l error exec -it build /bin/bash -c "bash <(curl -s https://codecov.io/bash)"
+  - stage: Integration test
+    name: Integration test - docker
+    # if: env(TRAVIS_EVENT_TYPE) = cron
```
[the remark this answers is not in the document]
red2k18: Will uncomment the condition after testing it on Travis. Integration tests will not run on every PR when that condition is uncommented.

integration_tests/fixtures/forseti/.gitignore
red2k18: It can certainly be consolidated, and it has been consolidated.

integration_tests/fixtures/forseti/outputs.tf
red2k18: Renamed.

main.tf
red2k18: `main.tf` in the top directory has the resources that needs to be created to deploy Forseti. `main.tf` in `integration_tests/fixtures/forseti` has the resources that needs to be created to run tests.

integration_tests/tests/forseti/controls/explain.rb
red2k18: Removed.

integration_tests/tests/forseti/controls/model.rb
red2k18: There is no stdout for `use` command.

integration_tests/tests/forseti/controls/model.rb
red2k18: Added it back.

integration_tests/tests/forseti/controls/model.rb
red2k18: Added it back.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: Done

integration_tests/tests/forseti/controls/inventory.rb
red2k18: If there are no errors then the database check confirms that the command ran successfully. Hence, I had removed the checks on `stdout` to avoid redundancy. Added it back.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: Done

integration_tests/tests/forseti/controls/inventory.rb
red2k18: If there are no errors then the database check confirms that the command ran successfully. Hence, I had removed the checks on stdout to avoid redundancy. Added it back.

integration_tests/tests/forseti/controls/explain.rb
red2k18: All the controls are invoked by kitchen in `kitchen.yml`. Since there is only one suite and controls are not specified, all the controls are invoked.

red2k18: This makes the code cleaner.

main.tf
red2k18: main.tf in the top directory has the resources that needs to be created to deploy Forseti. main.tf in integration_tests/fixtures/forseti has the resources that needs to be created to run tests. 

The code in `forseti-security/main.tf` needs to be updated only if there are changes in the way Forseti is deployed

integration_tests/fixtures/forseti/scripts/wait-for-forseti.sh
red2k18: I have removed that `echo` command that said Forseti is not ready yet. Essentially, Forseti is ready 60 seconds after the condition in the `if` block is met. Updated the script.

integration_tests/tests/forseti/controls/explain.rb
red2k18: The only other way to consolidate the tests is to use a regular expression. The complex regex impacts the readability and is harder to maintain when new assertions needs to be added.

integration_tests/tests/forseti/controls/explain.rb
red2k18: The only other way to consolidate the tests is to use a regular expression. The complex regex impacts the readability and is harder to maintain when new assertions needs to be added.

integration_tests/fixtures/bastion/main.tf
red2k18: It is challenging to come up with the IP addresses that needs to be allowed. Narrowing down IP address makes it difficult for developers to develop and test on their local environment. SSH key will also be needed to be able to SSH in. Not to forget, entire environment will be deconstructed in few minutes. So, it is fine to leave it as is. Also, CFT team is working on identity aware proxy module that is a great way to properly secure the environment and that can be leveraged at a later time.

.kitchen.yml
red2k18: `kitchen.yml` needs to be in the root to do top-level checks and to configure what hardware you need to run the tests on; how to provision the machine; and which combinations of test suites to exercise on the hardware.

integration_tests/fixtures/forseti/outputs.tf
red2k18: The output list is only needed to count the unique items that are defined.

main.tf
red2k18: `main.tf` in the root directory has been removed.

red2k18: `main.tf` in the root directory has been removed.

.travis.yml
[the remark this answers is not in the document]
blueandgold: Can you please add a comment to explain why we are checking on `cron`?  It might not be obvious to future reader of this code.

integration_tests/tests/forseti/controls/inventory.rb
blueandgold: The check on `stdout` is to verify that the grpc layer is working.  So, please check on the `stdout` here.

integration_tests/tests/forseti/controls/inventory.rb
blueandgold: ditto: The check on `stdout` is to verify that the grpc layer is working.  So, please check on the `stdout` here.

integration_tests/tests/forseti/controls/inventory.rb
blueandgold: This still checking on `stderr`.

integration_tests/tests/forseti/controls/model.rb
blueandgold: Seems it's still checking on `stderr`.

integration_tests/tests/forseti/controls/model.rb
blueandgold: The `stdout` for the `use` command is `""`.  Better than `stderr` here.
Also, you want to check that the model is set correctly in the CLI config, after the `use` command: `forseti config show`

red2k18:
@blueandgold Thanks for the review! Addressed two new comments.

integration_tests/tests/forseti/controls/inventory.rb
blueandgold: Just to make sure I understand, how the inspect expect works here.

We have our response as below, how can we add another check, say on the `status` field?
```
{
  "id": "1569821282926292",
  "startTimestamp": "2019-09-30T05:28:03Z",
  "completeTimestamp": "2019-09-30T05:40:29Z",
  "schemaVersion": 1,
  "countObjects": 12319,
  "status": "SUCCESS",
  "warnings": "",
  "errors": ""
}
```

.travis.yml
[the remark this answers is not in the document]
blueandgold: Can you uncomment this line now?
```
# if: env(TRAVIS_EVENT_TYPE) = cron
```

integration_tests/tests/forseti/controls/inventory.rb
blueandgold: Why is this still checking on `stderr`?

integration_tests/tests/forseti/controls/model.rb
blueandgold: If you have the `stdout`, then you really don't need the `stderr` check here.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: That would be another assertion. Added check on status: SUCCESS as well.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: I thought it would be nice to check on both `stdout` and `stderr`. Removed `stderr` checks as requested.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: I thought it would be nice to check on both `stdout` and `stderr`. Removed `stderr` checks as requested.

integration_tests/tests/forseti/controls/model.rb
red2k18: I thought it would be nice to check on both `stdout` and `stderr`. Removed `stderr` checks as requested.

.travis.yml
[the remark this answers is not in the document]
red2k18: Yes, the build is passing so I uncommented this. 
https://travis-ci.org/forseti-security/forseti-security/jobs/593108911

integration_tests/fixtures/bastion/main.tf
blueandgold: Please add a comment to document why `0.0.0.0/0` is okay from all the discussions we had on chat: this host is empheral, dev need access, etc.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_roles --prefix roles/iam").stdout).to match /roles\/iam.serviceAccountUser/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_roles --prefix roles/iam").stdout).to match /roles\/iam.workloadIdentityUser/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_roles --prefix roles/iam").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Dropped `stderr`.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_roles --prefix roles/storage").stdout).to match /roles\/storagetransfer.user/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_roles --prefix roles/storage").stdout).to match /roles\/storagetransfer.viewer/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_roles --prefix roles/storage").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Dropped `stderr`.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_permissions --roles roles/iam.roleAdmin").stdout).to match /resourcemanager.projects.get/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_permissions --roles roles/iam.roleAdmin").stdout).to match /resourcemanager.projects.getIamPolicy/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_permissions --roles roles/iam.roleAdmin").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Dropped `stderr`.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_permissions --roles roles/storage.Admin").stdout).to match /storage.objects.setIamPolicy/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_permissions --roles roles/storage.Admin").stdout).to match /storage.objects.update/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_permissions --roles roles/storage.Admin").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Dropped `stderr`.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --role roles/storage.admin --expand_groups").stdout).to match /project\/release-automate-silver/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --role roles/storage.admin --expand_groups").stdout).to match /serviceaccount\/project-factory-22907@release-automate-silver.iam.gserviceaccount.com/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --role roles/storage.admin --expand_groups").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Dropped `stderr`.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --role roles/storage.admin").stdout).to match /project\/release-automate-silver/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --role roles/storage.admin").stdout).to match /serviceaccount\/project-factory-22907@release-automate-silver.iam.gserviceaccount.com/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --role roles/storage.admin").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Dropped `stderr`.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --permission storage.buckets.delete").stdout).to match /serviceaccount\/84605163300@cloudservices.gserviceaccount.com/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --permission storage.buckets.delete").stdout).to match /serviceaccount\/project-factory-22907@release-automate-silver.iam.gserviceaccount.com/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --permission storage.buckets.delete").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Dropped `stderr`.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --permission storage.buckets.delete --expand_groups").stdout).to match /serviceaccount\/84605163300@cloudservices.gserviceaccount.com/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --permission storage.buckets.delete --expand_groups").stdout).to match /serviceaccount\/project-factory-22907@release-automate-silver.iam.gserviceaccount.com/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer access_by_authz --permission storage.buckets.delete --expand_groups").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Dropped `stderr`.

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -0,0 +1,373 @@
...
+            command("forseti inventory create").result
+            inventory_id = JSON.parse(command("forseti inventory list").stdout).fetch("id")
+            command("forseti model create --inventory_index_id #{inventory_id} model_new").result
+            command("forseti model use model_new").result
+        end
+
+        describe "List members" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_members --prefix rdevani").stderr).to eq ""
```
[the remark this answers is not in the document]
red2k18: Updated assertion.

integration_tests/fixtures/bastion/main.tf
red2k18: Done

.travis.yml
```diff
@@ -13,26 +13,39 @@
...
+    install:
+    -
+    script:
+    - "./install/scripts/docker_run_forseti.sh"
+    - "./install/scripts/docker_unittest_forseti.sh"
+    - "./install/scripts/docker_pystyle_forseti.sh"
+    after_success:
+    - # Collect codecov results.
+    - docker -l error exec -it build /bin/bash -c "bash <(curl -s https://codecov.io/bash)"
+  - stage: Integration test
```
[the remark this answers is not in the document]
red2k18: It's happening in `/usr/local/bin/test_integration.sh`. Documented the same in `travis.yml`.

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,122 @@
...
+
+            it "should be visible in the database" do
+                expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(DISTINCT gcp_inventory.inventory_index_id) from gcp_inventory join inventory_index on inventory_index.id = gcp_inventory.inventory_index_id where inventory_index.id = #{@inventory_id};\"").stdout).to match /1/
+            end
+        end
+
+        describe "Get an inventory" do
+
+            it "should be visible from the command-line" do
+                expect(command("forseti inventory list").stdout).to match /#{@inventory_id}/
```
blueandgold: This command should be `get`.

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,122 @@
...
+    describe "Inventory" do
+        before :context do
+            command("forseti inventory purge 0").result
+            command("forseti inventory create").result
+
+             # This variable cannot be used after all the inventories have been purged..
+            @inventory_id = JSON.parse(command("forseti inventory list").stdout).fetch("id")
+        end
+
+        describe "Create an inventory" do
```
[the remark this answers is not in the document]
red2k18: As discussed, inventory needs to be created once for all the tests instead of creating for each and every test. Hence, `inventory create` command needs to be in the before section. The description is aptly named `Create an inventory` as successful creation of inventory is checked in this test by asserting against `SUCCESS` and by performing the database checks.

integration_tests/tests/forseti/controls/inventory.rb
```diff
@@ -0,0 +1,122 @@
...
+
+            it "should be visible from the command-line" do
+                expect(command("forseti inventory list").stdout).to match /SUCCESS/
```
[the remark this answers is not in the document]
red2k18: Context is different as the list command is checking for the successful creation of inventory on line 32 and checking if the inventory id is retrieved on line 51. Even though the command is same, the usage and the assertion based on the test they are in are different.

integration_tests/tests/forseti/controls/inventory.rb
red2k18: Updated.

.travis.yml
```diff
@@ -13,26 +13,41 @@
 # limitations under the License.
 
 sudo: required
+jobs:
+  include:
+  - stage: Unit test
+    name: Unit Tests with Python 3.6
+    language: python
```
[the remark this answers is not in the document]
red2k18: It can be added. Added it in.

integration_tests/tests/forseti/controls/inventory.rb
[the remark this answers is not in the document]
blueandgold: I don't think you should repeat this command.

image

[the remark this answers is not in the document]
red2k18: Removed duplicate command

Issue #2938, opened 2019-06-22
2.17 - Terraform install -missing variable db_migrator.py 


jf-marquis-Adeo:
- [x] Update Forseti db_migrator.py to support custom database name for GCE
- [ ] Update Terraform and Helm Charts to support custom database name on GKE
---

Hello,
after installing 2.17 through terraform with specified db_name
 ` cloudsql_db_name                                    = "forseti-security"`
my forseti refuse to start...
after investigation i see that in /home/ubuntu/forseti-security/install/gcp/upgrade_tools/db_migrator.py database name is hardcoded :
`DEFAULT_DB_CONN_STR = 'mysql://root@127.0.0.1:3306/forseti_security'`
so db_migrator failed
`['/home/ubuntu/forseti-security/install/gcp/upgrade_tools/db_migrator.py']
Traceback (most recent call last):
  File "/home/ubuntu/forseti-security/.eggs/SQLAlchemy-1.2.18-py3.6-linux-x86_64.egg/sqlalchemy/engine/base.py", line 2228, in _wrap_pool_connect
    return fn()
  File "/home/ubuntu/forseti-security/.eggs/SQLAlchemy-1.2.18-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 425, in connect
    return _ConnectionFairy._checkout(self)
  File "/home/ubuntu/forseti-security/.eggs/SQLAlchemy-1.2.18-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 822, in _checkout
    fairy = _ConnectionRecord.checkout(pool)
  File "/home/ubuntu/forseti-security/.eggs/SQLAlchemy-1.2.18-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 554, in checkout
    rec = pool._do_get()
  File "/home/ubuntu/forseti-security/.eggs/SQLAlchemy-1.2.18-py3.6-linux-x86_64.egg/sqlalchemy/pool.py", line 1250, in _do_get
    self._dec_overflow()
  File "/home/ubuntu/forseti-security/.eggs/SQLAlchemy-1.2.18-py3.6-linux-x86_64.egg/sqlalchemy/util/langhelpers.py", line 67, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
[...]

PR #3206, opened 2019-09-16
Fix db_migrator.py script for customer Forseti DB name

gkowalski-google:
Updated the db_migrator.py script by using the FORSETI_DB_NAME environment variable for the Forseti database name. This is set via Terraform and the startup script. Will have some additional PRs for the GKE changes.

Addresses #2938 for GCE installations.


blueandgold:
This has been release in Forseti v2.21 release.

Issue #2880, opened 2019-06-03
Group settings scanner gives violation even though all the settings match.


red2k18:
Below is the rule file config and group settings provided by the user.

rules:
  - name: All groups with iam policies should have all of these settings
    mode: whitelist
    only_iam_groups: True
    groups_emails:
      - '*'
    settings:
      #allowExternalMembers: True
      whoCanJoin: "INVITED_CAN_JOIN"
      whoCanInvite: "ALL_MANAGERS_CAN_INVITE"
      whoCanAdd: "ALL_MANAGERS_CAN_ADD"
      allowExternalMembers: False
      #whoCanLeaveGroup: "ALL_MANAGERS_CAN_LEAVE"\

red2k18:
<?xml version="1.0" encoding="UTF-8"?>
<entry xmlns="http://www.w3.org/2005/Atom" xmlns:apps="http://schemas.google.com/apps/2006" xmlns:gd="http://schemas.google.com/g/2005">
 <id>tag:googleapis.com,2010:apps:groupssettings:GROUP:it@domain.com</id>
 <title>Groups Resource Entry</title>
 <content type="text">IT</content>
 <author>
  <name>Google</name>
 </author>
 <apps:email>it@domain.com</apps:email>
 <apps:name>IT</apps:name>
 <apps:description/>
 <apps:whoCanJoin>INVITED_CAN_JOIN</apps:whoCanJoin>
 <apps:whoCanViewMembership>ALL_MEMBERS_CAN_VIEW</apps:whoCanViewMembership>
 <apps:whoCanViewGroup>ALL_MEMBERS_CAN_VIEW</apps:whoCanViewGroup>
 <apps:whoCanInvite>ALL_MANAGERS_CAN_INVITE</apps:whoCanInvite>
 <apps:whoCanAdd>ALL_MANAGERS_CAN_ADD</apps:whoCanAdd>
 <apps:allowExternalMembers>false</apps:allowExternalMembers>
 <apps:whoCanPostMessage>ANYONE_CAN_POST</apps:whoCanPostMessage>
 <apps:allowWebPosting>true</apps:allowWebPosting>
 <apps:maxMessageBytes>26214400</apps:maxMessageBytes>
 <apps:isArchived>true</apps:isArchived>
 <apps:archiveOnly>false</apps:archiveOnly>
 <apps:messageModerationLevel>MODERATE_NONE</apps:messageModerationLevel>
[...]

PR #3237, opened 2019-09-26
Fix the Group Settings Type

gkowalski-google:
The GSuite API is returning the `allowExternalMembers` as a string: either `false` or `true`. Since this is not returned as a json boolean, the json.loads() method is not converting this to a Python boolean value. This means that all non-empty strings will be converted to True by the bool() method. This fix correctly converts these strings to a Python bool.

Addresses #2880

### DIFF
--- a/google/cloud/forseti/common/gcp_type/groups_settings.py
+++ b/google/cloud/forseti/common/gcp_type/groups_settings.py
@@ -44,7 +44,7 @@ class GroupsSettings(resource.Resource):
             whoCanViewMembership (str): Setting for who can view membership.
             whoCanViewGroup (str): Setting for who can view group.
             whoCanInvite (str): Setting for who can invite to group.
-            allowExternalMembers (str): Setting for are external members
+            allowExternalMembers (bool): Setting for are external members
             allowed.
             whoCanLeaveGroup (str): Setting for who can leave group.
         """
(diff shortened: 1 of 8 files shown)


google/cloud/forseti/common/gcp_type/groups_settings.py
```diff
@@ -79,5 +79,6 @@ def from_json(cls, email, settings):
             whoCanViewMembership=settings.get('whoCanViewMembership'),
             whoCanViewGroup=settings.get('whoCanViewGroup'),
             whoCanInvite=settings.get('whoCanInvite'),
-            allowExternalMembers=settings.get('allowExternalMembers'),
+            allowExternalMembers=str(
+                settings.get('allowExternalMembers', '')).lower() == 'true',
```
gkowalski-google: Main change of the PR.

hshin-g: Is there a reason this is converted to a string and compared to a string vs. left as bool and compared to bool?

gkowalski-google: `settings.get('allowExternalMembers', '')` is the response from the GSuite API, which returns this value as a string instead of a bool. For example `allowExternalMembers="true"`. I am converting the string to a string to future proof, in case of one day this is correctly returned as a bool.

hshin-g: Ok thanks for the explanation!

PR #3261, opened 2019-10-03
Fixing MySQL connection string for dev

kevensen:
By specifying mysql+pymysql://, SQLAlchemy will use the _PyMySQL_ driver.  By default SQLAlchemy uses _mysqlclient_.  Forseti uses PyMySQL with PR #3190.


kevensen:
This can go into 2.23

Issue #3217, opened 2019-09-19
Set password on MySQL root account


gkowalski-google:
- [ ] Investigate what parts of Forseti rely on the root account
  - [ ] rename the account from root to something that is not a default to brute force
- [ ] Update Terraform module to set root password
- [ ] Update Forseti to use root password (if needed)
---

### Story
The Forseti Terraform module [creates a root account with no password](https://github.com/forseti-security/terraform-google-forseti/blob/master/modules/cloudsql/main.tf#L98) (after [Terraform deletes the default root account](https://www.terraform.io/docs/providers/google/r/sql_database_instance.html)). This does not follow best practices and leaves the database in a potentially vulnerable state.

### Proposed Solution
Investigate if/what parts of Forseti rely on this root account. Update Terraform module and Forseti to support CloudSQL root account with a password.

### Acceptance Criteria
- [ ] GIVEN: a new project
WHEN: Forseti GCEis installed
THEN: the MySQL root user has a password
- [ ] GIVEN: an existing Forseti installation on GCE
WHEN: Forseti is upgraded via Terraform
THEN: the MySQL root user has a password

kevensen:
I think we need to solve https://github.com/forseti-security/terraform-google-forseti/issues/310 in the process of resolving this.

gkowalski-google:
Notes about why these changes got reverted. Please take the following into consideration.

**Issues**
1) User installs Forseti 2.24.0 with Terraform < 5.1.0: Forseti 2.24.0 will expect Terraform to pass in the Cloud SQL user/pass via environment variable. If this is not passed in (for example someone uses Terraform < 5.1.0), then Forseti will not be able to connect to Cloud SQL
2) User installs Forseti < 2.24.0 with Terraform >= 5.1.0: Terraform 5.1.0 sets a different Cloud SQL user and random password. Forseti < 2.24.0 will attempt to connect as root with no password, and will not be able to connect to Cloud SQL
3) Users who are connecting to the database will need to know how to get the username/password and update their processes/apps.
4) Investigate if Forseti Visualizer will be effected.

**Possible Solutions**
- Resolve #1 by updating Forseti to default to root and no password if none provided
- Resolve #2 by moving the initialize_forseti_services script to Terraform repo so that regardless which version of Forseti is deployed, the Terraform module would be in control of the VM metadata startup script. By doing so, we can configure the Cloud SQL proxy to work with user/password regardless of the version of Forseti deployed.

[...]

Issue #2802, opened 2019-05-10
don't use root as the cloudsql db user


nbards:
- [ 2.13, 2.14, 2.14.1, 2.15 ] Which version of Forseti Security you're using (look in `/home/ubuntu/forseti-security/google/cloud/forseti/init.py` on the client).
 - [ installer ] Which module(s) (inventory, scanner, enforcer, explain) you're having trouble with.
 - [ n/a ] Include errors, log output, and host operating system, including installed packages.
 - [ deployment manager ] Note how you deployed (via Deployment Manager, Terraform, local system, etc.).

Following reasonable industry practices, the mysql root user should not be used by applications for operational access.  It would be great if forseti used a non-root user with minimal necessary permissions, revoking un-needed and un-used rights per https://cloud.google.com/sql/docs/mysql/users

cybermerc:
There is a reason for the current setup documented here  https://github.com/forseti-security/forseti-security/issues/921

That being said, its a bit dated so going to look and see if this can be addressed now

gkowalski-google:
Resolved by #3217

PR #3266, opened 2019-10-03
Added db user and password in the connection string.

joecheuk:
Resolves: #3217 , #2802
Relates: #3321 
Relates: https://github.com/forseti-security/terraform-google-forseti/pull/291


kevensen:
@gkowalski-google Yeah, let's tackle that in another issue (or 2).
1. Terraform - Pass the user and password variables to the on_gke submodule.
2. Helm - Update the templates to utilize a username and password variable.

gkowalski-google:
> @gkowalski-google Yeah, let's tackle that in another issue (or 2).
> 
> 1. Terraform - Pass the user and password variables to the on_gke submodule.
> 2. Helm - Update the templates to utilize a username and password variable.

New issue: https://github.com/forseti-security/forseti-security/issues/3294

joecheuk:
@kevensen updated docker entry point, PTAL
@gkowalski-google I have done some investigation, using KMS here doesn't seem to provide any security benefits as if users can ssh to the VM, they can run gcloud command with the VM's elevated privilege to decrypt the password

gkowalski-google:
> @kevensen updated docker entry point, PTAL
> @gkowalski-google I have done some investigation, using KMS here doesn't seem to provide any security benefits as if users can ssh to the VM, they can run gcloud command with the VM's elevated privilege to decrypt the password

Thanks for evaluating that!

kevensen:
Attempt to reopen

Issue #3145, opened 2019-08-28
Triage impact related to CAI field name changes for sourceRange/sourceRanges


dekuhn:
- [ ] Please determine and open issues if there are changes required for Forseti CAI Inventory Image
- [ ] Please determine and open issues if there are changes required for Forseti Firewall and other scanners
- [ ] Please determine and open issues if there are changes required in the Policy Library

Hello Cloud Asset Inventory Customer,  We are writing to let you know that Cloud Asset Inventory is releasing some asset field name updates starting on September 10, 2019. Initially, both updated field names and existing names will be returned together. The previous naming of the re-named asset fields will no longer be returned in about 6 months. We suggest you review whether these field names are compatible with your tools to avoid possible service disruption.  What do I need to know?  We are making these name changes because in a number of Cloud Asset Inventory resources, some field names in the asset returned by Cloud Asset Inventory differ from their corresponding public APIs. Take the compute.googleapis.com/Firewall asset for example:                   Cloud Asset Inventory export result   firewalls.get API result       "sourceRange":["10.128.0.0/9"]}   "sourceRanges":["10.128.0.0/9"]}       In order to be consistent with each service’s public API format, we are updating those inconsistent field names. You can review the mapping table for more information.  We are making this change in phases so you will have time to adjust to the updated field names:  Phase 1  From September 10, 2019, we will be adding the updated field names alongside with existing fields. The content of the fields is identical to the current ones, only field names will be modified. We expect this change to be backwards compatible for all our customers.  Phase 2  We will remove the fields with deprecated names in about 6 months, we will send you another notification to remind you of the deprecation.  What do I need to do?  Although we are expecting the asset field names change to be backwards compatible with all our customers, please review the list of updated asset field names.  If you have any current data processing with dependencies on any field names to be updated, you should update these fields as soon as possible to avoid possible service disruption.  If you have any questions or require assistance, please reply to this email to contact Google Cloud Support.  Thank you for being a valued Cloud Asset Inventory customer. 

Cloud Asset Inventory export result | firewalls.get API result | "sourceRange":["10.128.0.0/9"]} | "sourceRanges":["10.128.0.0/9"]}
  
Cloud Asset Inventory export result | firewalls.get API result
"sourceRange":["10.128.0.0/9"]} | "sourceRanges":["10.128.0.0/9"]}

gkowalski-google:
[...]

PR #3276, opened 2019-10-07
Remove custom mapping for CAI field name updates

gkowalski-google:
Removed the mapping for CAI resources from the old CAI field names to the API field names. Ran some tests to verify that the updated resources are showing up properly in the CAI export.

Addresses #3145


gkowalski-google:
> Sweet! Looks great overall. Can you confirm that you have manually tested this by running an inventory and check all the changed fields here are correct in Inventory?

Yes I verified the CAI export has the fields that the documentation mentions, and did ran a new inventory on our test org and verified the resources are being imported correctly.

PR #3277, opened 2019-10-08
Added Explain test for all members

red2k18:


integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -32,6 +32,49 @@
...
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_members").stdout).to match /projectviewer\/release-automate-silver/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_members").stdout).to match /projecteditor\/release-automate-silver/
+            end
+
+            it "should be visible from the command-line" do
+                expect(command("forseti explainer list_members").stdout).to match /serviceaccount\/84605163300-compute@developer.gserviceaccount.com"/
```
blueandgold: Is the service account here always in place for this test to pass?  i.e. if the env is purged and repopulated, will this service account still be present?

red2k18: Yes, below build verifies that statement.
https://travis-ci.org/forseti-security/forseti-security/jobs/596161769

PR #3278, opened 2019-10-08
Added Explain test for all roles

red2k18:


integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -32,6 +32,65 @@
             end
         end
 
+        describe "List all roles" do
```
blueandgold: qq: What is the test here testing?  The command will return all roles, but we are only asserting on a subset of the result.  How do we know all the roles are indeed returned correctly?

red2k18: There are 3 ways to test `forseti explainer list_roles`.
1. If `forseti explainer list_roles prefix rdevani` passes then `forseti explainer list_roles` should also pass and this test is not required.
2. Make assertions against each and every row returned by `forseti explainer list_roles`.
3. Make assertions against subset of rows returned by `forseti explainer list_roles`. 
This test implements #3.

Issue #3279, opened 2019-10-08
Website - Roles link on migrate page is broken


gkowalski-google:
The roles required by the service account points to the deleted 4.2.0 module. Update this to point to the latest version 4.3.0.

PR #3281, opened 2019-10-08
Fix roles link on website migrate page

gkowalski-google:
Update the roles link needed by the terraform service account to point to 4.3.0 instead of 4.2.0.

Addresses #3279


_docs/_latest/setup/manual.md
```diff
@@ -56,7 +56,7 @@ service account, see
 
 * For information about the specifications needed for a Forseti server
   VM instance, see the
-  [Compute Engine deployment template](https://github.com/forseti-security/forseti-security/blob/master/deployment-templates/compute-engine/server/forseti-instance-server.py)
+  [Compute Engine deployment template](https://github.com/forseti-security/forseti-security/blob/master-old/deployment-templates/compute-engine/server/forseti-instance-server.py)
```
gkowalski-google: The deployment templates were removed as part of removing support for the Python installer/DM. I have updated these broken links to use the `master-old` branch which points to the commit that `master` was pointing to before it was updated.

PR #3299, opened 2019-10-10
Updated CFT Docker image to add support for Terraform G Suite provider

red2k18:


.travis.yml
```diff
@@ -37,19 +37,19 @@ jobs:
     name: Integration test - docker
     # Check on cron ensures build for integration tests is triggered only when
     # cron job runs on dev branch. This cron job is currently set to run daily.
-    if: type = cron
+    # if: type = cron
```
blueandgold: Why is this commented out?

red2k18: It was commented out to ensure build passes with new docker image, and it has been verified that it works as expected. 

Build passed: https://travis-ci.org/forseti-security/forseti-security/builds/596207609?utm_source=github_status&utm_medium=notification

Uncommented the cron job check.

PR #3300, opened 2019-10-10
Created G Suite group for integration testing

red2k18:
G Suite group has been created using the CFT framework for integration testing to prove that G Suite resources can also be created using the chosen framework.


red2k18:
This PR should be merged only after merging #3299.

PR #3301, opened 2019-10-10
Created bucket with allUsers and allAuthenticatedUsers permission for integration testing

red2k18:


red2k18:
Build passed: https://travis-ci.org/forseti-security/forseti-security/builds/595906300?utm_source=github_status&utm_medium=notification

gkowalski-google:
> This PR should contain not just the resource creation, but also the test. This way, I can see how the buckets are used and tested.

Agreed. Which test case is this for?

red2k18:
@blueandgold @gkowalski-google Test for this resource and can pushed after #3332 is merged in.

blueandgold:
@red2k18 #3332 is merged in.  Please see gregg's question.  Thanks.

red2k18:
@gkowalski-google This is for scanner module, more specifically for IAM scanner.

red2k18:
Added test for IAM scanner. Build passed: https://travis-ci.org/forseti-security/forseti-security/jobs/603971128

integration_tests/tests/forseti/controls/explain.rb
```diff
@@ -60,31 +60,6 @@
             end
         end
 
-            it "should be visible from the command-line" do
-                expect(command("forseti explainer list_members").stdout).to match /serviceaccount\/84605163300-compute@developer.gserviceaccount.com"/
```
blueandgold: Why are these list member results removed in this PR?  Do they have anything to do with the adding the buckets?

integration_tests/tests/forseti/controls/scanner.rb
```diff
@@ -40,6 +40,10 @@
                expect(command("forseti scanner run").stdout).to match /Scan completed/
            end
 
+           it "should be visible in the database" do
+               expect(command("mysql -u root --host 127.0.0.1 --database forseti_security --execute \"select count(*) from violations where violation_type = 'IAM_POLICY_VIOLATION';\"").stdout).to match /2/
```
blueandgold: Is there a way to tune the db query so that it exactly find the violation from the bucket that's added here?  Otherwise, this test will break, when other resources with IAM violations are added.

integration_tests/tests/forseti/controls/explain.rb
red2k18: I have removed all the service accounts that do not need to be preserved for the tests to run for clean-up purpose.

integration_tests/tests/forseti/controls/scanner.rb
red2k18: Bucket names are randomized, so a check on the name of the bucket cannot be added, but a check to see if the resource is `bucket` can be added. This will prevent breakage when resources other than`bucket` are added.

integration_tests/tests/forseti/controls/explain.rb
blueandgold: Then it would be best for this clean-up to be in a separate PR, and not be part of this PR.  So, please remove this change here, and submit separately in a new PR.

integration_tests/tests/forseti/controls/scanner.rb
blueandgold: Can you please add the check to for the resource is bucket into the query?

integration_tests/tests/forseti/controls/explain.rb
red2k18: Some of this service accounts have been deleted as they were in pending deletion phase. So, it needs to be removed in this PR for tests to pass.

blueandgold: Can you still break this out into a separate PR, and submit that first to fix the brokenness?  Then, you can submit this later.  The separation helps with later discoverability of the work that's being done.

integration_tests/tests/forseti/controls/scanner.rb
blueandgold: My point is that the query is currently very generic, and needs to be more scoped to the bucket violations that you are adding here.  Otherwise, this test will break again, when new IAM violations are added in the future.
```
\"select count(*) from violations where violation_type = 'IAM_POLICY_VIOLATION';\"
```

blueandgold: Will merge this in for now, and improve this when more violations are added.

PR #3302, opened 2019-10-10
Update stale.yml

dekuhn:
Remove onlyLabels setting,  reduce # of updates per hour and updated default stale label.


dekuhn:
@blueandgold I suspect the onlyLabels: [] setting has the Stale bot disabled. I am removing this config and tweaking a few other minor details. I am using stale label vs wontfix as I think it is more descriptive of the state of the issues.

Issue #3207, opened 2019-09-16
Publish how-to documentation to use Cloud Profiler with Forseti


dekuhn:
- [ ] Add to Forseti docs how to enable Cloud Profiler
- [ ] Add to Forseti docs how to export and share Cloud Profiler data with Forseti team.

PR #3303, opened 2019-10-10
[WEBSITE] Add Cloud Profiler Documentation

hshin-g:
Addresses Issue #3207.

### DIFF
--- a/_docs/_latest/configure/cloud-profiler/index.md
+++ b/_docs/_latest/configure/cloud-profiler/index.md
@@ -52,6 +52,8 @@ For an in-depth instruction on how to use the interface, refer to the official C
 
 ## Disabling Cloud Profiler
 
+*Important:* Once you have completed an analysis of your Forseti application, we recommend that you disable the Cloud Profiler.
+
 In your `main.tf` file, set the `cloud_profiler_enabled` variable in the Forseti Terraform module to `false`:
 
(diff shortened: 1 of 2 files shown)


_docs/_latest/configure/cloud-profiler/index.md
```diff
@@ -0,0 +1,52 @@
...
+}
+```
+
+Apply the Terraform module.
+
+```
+terraform apply
+```
+
+## Permissions & Installations
```
blueandgold: I think it would be awesome to be clear that when people are done with troubleshooting, they should disable it.  Otherwise, there will be processing overhead that will slow down Forseti.

dekuhn:
@hshin-g If we have a good profile graphic can we include it in the docs so people can get an understanding of what the results look like?

hshin-g:
@dekuhn They can navigate to the link that is included in the docs for a better understanding of what they will get as an output. I want to make sure that the docs are future proof so would like to avoid adding in any screenshots.

dekuhn:
> @dekuhn They can navigate to the link that is included in the docs for a better understanding of what they will get as an output. I want to make sure that the docs are future proof so would like to avoid adding in any screenshots.

@hshin-g Good point on keeping it future proof. Thanks for considering it!

PR #3309, opened 2019-10-12
Updated config validator integration with parallel policy evaluation.

joecheuk:


joecheuk:
Will create another one against master

PR #3310, opened 2019-10-14
Updated CV to the latest version w/ parallel policy evaluation.

joecheuk:


google/cloud/forseti/scanner/scanners/config_validator_util/validator_client.py
```diff
@@ -105,37 +106,23 @@ def paged_audit(self, assets):
...
-                data_loaded += current_page_size
-                self.add_data(paged_assets)
-                paged_assets = []
-                current_page_size = 0
-            if data_loaded >= self.max_audit_size:
-                LOGGER.debug('Auditing data, size: %s', data_loaded)
-                violations = self.audit()
-                self.reset()
-                data_loaded = 0
+            if current_page_size + asset_size >= self.max_audit_size:
```
hshin-g: Is this max_audit_size or max_page_size?

google/cloud/forseti/scanner/scanners/config_validator_util/validator_client.py
```diff
@@ -105,37 +106,23 @@ def paged_audit(self, assets):
...
+                violations = self.review(paged_assets)
                 if violations:
                     yield violations
+                paged_assets = []
```
hshin-g: Should paged_assets = [] and current_page_size = 0 be unindented once?

google/cloud/forseti/scanner/scanners/config_validator_util/validator_client.py
joecheuk: max_page_size variable is not being used, removed the variable. max_audit_size is correct in this case.

google/cloud/forseti/scanner/scanners/config_validator_util/validator_client.py
joecheuk: That looks like the right indentation to me, per audit we want to remove all the audited data and reset the page size hence it's inside the if statement.

hshin-g: Ok! Thanks for the clarification.

PR #3326, opened 2019-10-17
feat: send mailjet email with HTML content

jf-marquis-Adeo:
Small changes due to malformed HTML content


blueandgold:
@jf-marquis-Adeo I am cherrypicking this into the 2.23 release, that we are preparing.  So, please watch out for it.

PR #3328, opened 2019-10-17
Modified `forseti_server_conf.yaml` and added integration test for Scanner

red2k18:


.travis.yml
```diff
@@ -37,19 +37,19 @@ jobs:
...
+    # if: type = cron
     services:
     - docker
     before_install:
     - openssl aes-256-cbc -K $encrypted_5c0ae2aa862c_key -iv $encrypted_5c0ae2aa862c_iv
       -in secret-key.tar.enc -out secret-key.tar -d
     - tar xvf secret-key.tar
     install:
-    - docker image pull gcr.io/cloud-foundation-cicd/cft/developer-tools:0.4.1
+    - docker image pull gcr.io/cloud-foundation-cicd/cft/developer-tools:0.2.0
```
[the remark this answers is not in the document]
red2k18: Updated.

.travis.yml
```diff
@@ -37,19 +37,19 @@ jobs:
...
     # Environment setup, tests execution and clean up is happening in
     # /usr/local/bin/test_integration.sh
     script:
     - docker container run -it -e KITCHEN_TEST_BASE_PATH="integration_tests/tests" -e SERVICE_ACCOUNT_JSON="$(< release-silver.json)"
-      --env-file travis-integration.env -v $(pwd):/workspace gcr.io/cloud-foundation-cicd/cft/developer-tools:0.4.1
+      --env-file travis-integration.env -v $(pwd):/workspace gcr.io/cloud-foundation-cicd/cft/developer-tools:0.2.0
```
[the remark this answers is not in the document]
red2k18: Updated.

integration_tests/tests/forseti/controls/scanner.rb
```diff
@@ -0,0 +1,50 @@
...
+
+require 'json'
+control 'Scanner' do
+
+
+   describe "Run scanner" do
+
+       before :context do
+           command("forseti inventory purge 0").result
+           command("forseti inventory create").result
```
blueandgold: Why is the inventory purged and created again?  Can the existing one be re-used?

integration_tests/tests/forseti/controls/scanner.rb
```diff
@@ -0,0 +1,50 @@
...
+           @inventory_id = JSON.parse(command("forseti inventory list").stdout).fetch("id")
+           command("forseti model create --inventory_index_id #{@inventory_id} model_scanner").result
+           command("forseti model use model_scanner").result
+           command("sudo cp /home/ubuntu/forseti-security/configs/forseti_conf_server.yaml /home/ubuntu/forseti-security/configs/forseti_conf_server_backup.yaml").result
```
blueandgold: Why is a backup copy needed?  I don't think the audit logging scanner needs to be disabled.  Even if necessary, why not just re-reset the the yaml parameter instead of mv the file back?

integration_tests/tests/forseti/controls/scanner.rb
```diff
@@ -0,0 +1,50 @@
...
+           @modified_yaml = yaml('/home/ubuntu/forseti-security/configs/forseti_conf_server.yaml').params
+           @modified_yaml["scanner"]["scanners"][0]["enabled"] = "true"
```
blueandgold: The reference to `[0]` is brittle.  What if the configuration change in the future?  If something is added in front of it, then this test will stop working.  Should do something more flexible: either determine the actual position of `audit_logging` or the first disabled scanner.

integration_tests/tests/forseti/controls/scanner.rb
red2k18: Backup copy is needed to restore the original state after executing the test. This makes tests independent of each other. If multiple changes are made in the config file, it wouldn't be feasible to revert changes one-by-one.

integration_tests/tests/forseti/controls/scanner.rb
red2k18: This is the first time inventory and model are created in `Scanner` control. It is needed for scanner to run successfully. If you're referring to re-using this from other controls, it is doable, but a new library needs to be created/module needs to be added. Do you think it's worth the effort?

integration_tests/tests/forseti/controls/scanner.rb
red2k18: Updated. PTAL

Issue #3338, opened 2019-10-21
Incorrect project_id reported in a cloudsql violation data in CSCC


dlinares-linux:
# Overview
This is a very small cosmetic issue where incorrect data is populated in a CloudSQL violation.
- Forseti Security version: v2.23.0
- Affected module: Scanner
- Deployment: Terraform on VMs with CSCC integration enabled

# Actual output:
```
violation_data 
{
  "authorized_networks": ["1.2.3.4/32"], 
  "full_name": "organization/123456789/folder/1234/project/forseti-security/cloudsqlinstance/1122334455/", 
  "instance_name": "forseti-server-db-ffffffff",
  "project_id": "forseti-server-db-ffffffff", 
  "require_ssl": true
}
```

# Expected output:
```
violation_data 
{
  "authorized_networks": ["1.2.3.4/32"], 
  "full_name": "organization/123456789/folder/1234/project/forseti-security/cloudsqlinstance/1122334455/", 
  "instance_name": "forseti-server-db-ffffffff",
  "project_id": "forseti-security", 
  "require_ssl": true
}
```

# Code to correct
The problem comes from the cloudsql_rules_scanner.py, line 70 where we are reporting the resource_id instead of the project_id as seen in the last line below:
```
            violation_data = {'instance_name': violation.instance_name,
                              'full_name': violation.full_name,
                              'authorized_networks': (
                                  violation.authorized_networks),
                              'require_ssl': violation.require_ssl,
                              'project_id': violation.resource_id}
```

red2k18:
Hi @dlinares-linux, I see that the fix has been merged into master. Thanks again for your contribution. Is there anything else we can help you with?

PR #3346, opened 2019-10-22
Report project_id in CloudSQL violations (#3338)

dlinares-linux:
Correcting a minor issue where the resource_id was reported
in CloudSQL violation data instead of the project_id.


blueandgold:
Thanks for fixing this!

gkowalski-google:
Fixes #3338

PR #3348, opened 2019-10-22
[WEBSITE] Add installation instructions for Terraform in setup page

hshin-g:
Add in manual Terraform installation instructions in website setup page


_docs/_latest/setup/install.md
```diff
@@ -24,11 +24,189 @@ organization Cloud IAM policy.
...
+### IAM Roles
+For this module to work, you need the following roles enabled on the Service Account.
+
+{% include docs/latest/forseti-terraform-sa-roles.md %}
+
+## Inputs
+Forseti provides many optional settings for users to customize for their environment and security requirements. 
+View the list of inputs below to see all of the available options.
+
+{% include docs/latest/forseti-terraform-inputs.md %}
```
gkowalski-google: Do we want to include the TF inputs/outputs in the website? I think it is better to link to the TF repo here where we already have a script to update these inputs as new variables are changed. Since this list is changing a lot, I think it's better to keep that documented in one place in the TF repo. Same for the outputs.

_docs/_latest/setup/install.md
```diff
@@ -24,11 +24,189 @@ organization Cloud IAM policy.
 
 ## Setting up Forseti Security
 
-Users can install Forseti using Terraform by clicking the button below.
+The Forseti Terraform module is the only supported method of installing Forseti Security The default infrastructure for
```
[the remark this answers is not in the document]
hshin-g: Good catch- thanks!

_docs/_latest/setup/install.md
hshin-g: I agree that it's good to have the inputs/outputs in one place. @blueandgold do you want to jump in here?

_docs/_latest/setup/install.md
```diff
@@ -24,11 +24,196 @@ organization Cloud IAM policy.
...
+gcloud auth login
+```
+
+### Install Terraform
+Terraform version 0.12 is required for this module, which can be downloaded from the 
+[Terraform website](https://www.terraform.io/downloads.html).
+
+### Create the Service Account and enable required APIs
+
+#### Service Account
```
blueandgold: We have mentioned elsewhere that if someone is org admin, they don't need to create a separate service account.  Can we do mention that here too?

hshin-g: @blueandgold added in- please take a look!

PR #3370, opened 2019-10-24
Update Integration tests to restrict bastion host to specific netblocks

dekuhn:
Adding specific network blocks for firewall rule


integration_tests/fixtures/bastion/main.tf
```diff
@@ -76,13 +76,7 @@ resource "google_compute_firewall" "main" {
...
   direction     = "INGRESS"
   priority      = "100"
-
-  # It is okay to leave this open to everyone as SSH key is also required to
-  # SSH into the server VM and SSH key is generated dynamically. Also, the host
-  # is ephemeral and the environment itself is deconstructed soon after tests
-  # are executed. Narrowing it down to few IP addresses makes it difficult for
-  # the developers to test on their local dev environment.
-  source_ranges = ["0.0.0.0/0"]
+  source_ranges = ["${bastion_firewall_netblocks}"]
```
hshin-g: The bastion_firewall_netblocks variable should be prepended by `var` (`"${var.bastion_firewall_netblocks}"`) as it tells Terraform what that value is (i.e. it's a variable, so should be found in `variables.tf`)

Since this is already a list, should be ok to just have it on its own, i.e. `source_ranges = "${var.bastion_firewall_netblocks}"`

dekuhn: Fixed. Thanks

dekuhn:
fixes #3374

dekuhn:
@gkowalski-google You were concerned if this change will prevent people from running the tests locally?

This net blocks are just default variable, this I **think** can be overwritten when testing locally. This value is a list so having a [terraform.tfvars](https://github.com/hashicorp/terraform/blob/master/website/intro/getting-started/variables.html.md#lists) is one way to overwrite this value. Does this address your concern with testing locally?

gkowalski-google:
> @gkowalski-google You were concerned if this change will prevent people from running the tests locally?
> 
> The network cider ranges are just default variable, this I **think** can be overwritten when testing locally. This value is a list so having a [terraform.tfvars](https://github.com/hashicorp/terraform/blob/master/website/intro/getting-started/variables.html.md#lists) is one way to overwrite this value. Does this address your concern with testing locally?

Thanks for following up on that. I think you are right, that should work. I think this is a good change and I can play around with getting this to work normally and add any steps to our dev workflow doc.

PR #3387, opened 2019-10-29
Support tests being run on any org/domain

gkowalski-google:
Updated syntax to move to HCL2. Updated main.tf to remove any mention of silver or to use the Terraform domain variable.

### DIFF
--- a/integration_tests/fixtures/forseti/main.tf
+++ b/integration_tests/fixtures/forseti/main.tf
@@ -27,7 +27,7 @@ resource "tls_private_key" "main" {
 }
 
 resource "local_file" "gce-keypair-pk" {
-  content  = "${tls_private_key.main.private_key_pem}"
+  content  = tls_private_key.main.private_key_pem
   filename = "${path.module}/sshkey"
 }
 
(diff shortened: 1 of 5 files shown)


integration_tests/fixtures/forseti/main.tf
```diff
@@ -124,25 +124,24 @@ resource "random_pet" "random_name_generator" {
...
+  key_ring        = google_kms_key_ring.test-keyring.self_link
   rotation_period = "100000s"
 }
 
 provider "gsuite" {
-  version     = "~> 0.1"
-  credentials = "/workspace/release-silver.json"
-  impersonated_user_email = "admin@silver.forsetisecurity.dev"
+  version                 = "~> 0.1"
+  impersonated_user_email = var.gsuite_admin_email
```
gkowalski-google: Main functional change is to remove the credentials variable here because it is not needed, and use a variable for the impersonated user email. Same thing with the test group that is created, this will be created in the org being tested.

blueandgold: Nice!

PR #3388, opened 2019-10-29
Adding platform as an env var in the user-agent

kevensen:
This PR exposes an environment variable, **FORSETI_PLATFORM**, in order to set the platofrm on which Forseti is running.  The default is _gce_.  This will be set to _gke_ inside the Forseti Server's pod definition.


tests/common/util/http_helpers_test.py
```diff
@@ -47,22 +49,24 @@ class HttpHelpersTest(ForsetiTestCase):
     """Test http_helpers utility functions."""
 
     def test_without_suffix(self):
+        http_helpers._PLATFORM = "gke"
```
gkowalski-google: I think we should leave existing unit tests as is and add new ones for the platform changes. Instead of `http_helpers._PLATFORM = "gke"` you can set the env variable in the test setup like this `os.environ['FORSETI_PLATFORM'] = 'gke'`.

kevensen: @gkowalski-google Please remind me, if I set this in the test setup, won't it be set for each unit test?

gkowalski-google: Yes, sorry for the confusing words. I just meant in the test setup/arrangement inside of the unit test you want to test this. I'd say add one test and verify platform is gce in the UA, and then one test where you set the env var to 'gke' and then verify that.

kevensen: Got it.  Will do.

kevensen: I remember why I set the attribute.  If I set `os.environ['FORSETI_PLATFORM'] = 'gke'` in the test, the http_helpers won't honor it.

kevensen: But, I think I can mock the environment variable.

kevensen:
@googlebot

tests/common/util/http_helpers_test.py
```diff
@@ -53,16 +55,36 @@ def test_without_suffix(self):
         response, content = http.request(DUMMY_URL)
         self.assertIn(UA_KEY, mock_http.headers)
         self.assertRegex(
-            mock_http.headers[UA_KEY], r'forseti-security/[0-9.]+$')
+            mock_http.headers[UA_KEY], r'forseti-security/[0-9.]+\s+gce$')
 
-    def test_with_suffix(self):
+    def test_with_suffix_set_platform(self):
```
gkowalski-google: Can you rename this method back to `test_with_suffix`? As it stands there are 2 methods with the same name (see L79).

kevensen: d'oh

gkowalski-google:
@kevensen Any other changes for this PR? If not, can you remove [WIP] so that we know it's ready for final review?

kevensen:
@gkowalski-google I am still doing functional testing.  It may not make it in to 2.24.

kevensen:
@gkowalski-google Thanks for the recommendation.

PR #3408, opened 2019-11-06
Iam policy investigation

kevensen:
Changed the exception logic to drop the member with an invalid type instead of the whole role binding when only one member has an invalid type.


google/cloud/forseti/common/gcp_type/iam_policy.py
```diff
@@ -182,13 +190,16 @@ def create_from(cls, binding):
         """
         if isinstance(binding, type(cls)):
             return binding
-        try:
-            return cls(binding.get('role'), binding.get('members'))
-        except errors.InvalidIamPolicyMemberError:
-            LOGGER.debug(
-                'Invalid IAM policy member: %s.', binding.get('members'))
+
+        iam_policy_binding = cls(binding.get('role'), binding.get('members'))
```
blueandgold: I am wondering if it's better to keep the `try/except` here, so to avoid every caller from having to handle the exception?  This also means the code in line 39 would not need to be changed.  Would something like this work?

```
try:
    iam_policy_binding = cls(binding.get('role'), binding.get('members'))
except errors.InvalidIamPolicyMemberError:	
    LOGGER.debug(
        'Invalid IAM policy member: %s.', binding.get('members'))
    return None

# The IamPolicyBinding.members attribute will be [] if the member types
# for all the members are invalid.
if not iam_policy_binding.members:
    return None

return iam_policy_binding
```

kevensen: Unfortunately, with this logic (and restoring line 39), we would still end up dropping the whole binding and not just the user with the invalid type from the binding.

PR #3449, opened 2019-11-15
Test/update scanner tests

### LINKED ISSUE #3409: Update Scanner test
## Story
The existing [scanner test](https://github.com/forseti-security/forseti-security/blob/master/integration_tests/tests/forseti/controls/scanner.rb) does a full inventory purge before/after the test. This can effect other tests and isn't necessary if the test tracks which inventory is created as part of the test.

## Proposed Solution
Separate out this control into multiple files/controls:
- Test updating config to enable scanner
- Test IAM Policy Scanner

Both controls should:

- Update the test to create the inventory/model like as done in [this test](https://github.com/forseti-security/forseti-security/blob/master/integration_tests/tests/forseti/controls/scanner-external_project_access.rb#L21).
- Update the control/file name to reflect what is being tested. In this case this test is updating the server config to enable a scanner and test that the enabled scanner is actually run.
- Assert exit code is 0 for scanner run
- Assert `scanner_status` is `SUCCESS` in the database
- Look into possibility of reusing inventory create command that can be shared amongst tests. This would create inventory/model with a random model name, verify exit code is 0, and verify `inventory_status` is `SUCCESS` in the database

The IAM Policy Scanner test should:
- Create bucket/object within test with gcloud and delete at the end

## Acceptance Criteria
The updated test completes successfully.

gkowalski-google:
Resolves #3409


gkowalski-google:
The failed Cloud Build check can be ignored; meant to disable that trigger that I am testing with.

goofiva:
@googlebot I consent.

PR #3464, opened 2019-11-19
Created an environment variable for PORT (server port)

### LINKED ISSUE #3462: Support configurable PORT environment variable in Forseti container
## Story
The [docker_entrypoint.sh](https://github.com/forseti-security/forseti-security/blob/master/install/scripts/docker_entrypoint.sh) script sets the default port to 50051 if not specified via the `--server_port` option.

https://github.com/forseti-security/forseti-security/blob/c302a45f7c9398232936f28ab1b0caca2728bb91/install/scripts/docker_entrypoint.sh#L51

Cloud Run configures this port via the `PORT` environment variable which is not configurable by the user and is set to 8080.

## Proposal
- [x] Change the line above to `SERVER_PORT=$PORT`.
- [x] Add an environment variable in the [Dockerfile](https://github.com/forseti-security/forseti-security/blob/master/Dockerfile) `PORT=50051`

## Validation
GIVEN the Forseti server running in a container
WHEN no `PORT` variable is specified
THEN the Forseti server starts listening on port 50051

GIVEN the Forseti server running in a container
WHEN the `PORT` variable is defined
THEN the Forseti server starts listening on the port with the value set in `PORT`.

kevensen:
Resolves #3462 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,6 +69,8 @@ FROM base AS runtime
 
 USER forseti
 
+ENV PORT 50051
+
 COPY --from=build --chown=forseti:forseti \
     /home/forseti/.local \
     /home/forseti/.local
--- a/install/scripts/docker_entrypoint.sh
+++ b/install/scripts/docker_entrypoint.sh
@@ -48,7 +48,7 @@ SQL_HOST=127.0.0.1
 SQL_PORT=3306
 
 SERVER_HOST=127.0.0.1
-SERVER_PORT=50051
+SERVER_PORT=${PORT:=50051}
 
 # Check if the k8s cloud sql proxy environment variables have been set, if so
 # overwrite our cloud sql variables with their contents


install/scripts/docker_entrypoint.sh
```diff
@@ -48,7 +48,7 @@ SQL_HOST=127.0.0.1
 SQL_PORT=3306
 
 SERVER_HOST=127.0.0.1
-SERVER_PORT=50051
+SERVER_PORT=${PORT:=50051}
```
hshin-g: Can you explain what this syntax is doing?

kevensen: This is setting `SERVER_PORT` to the value of `PORT`.  If `PORT` is unset or null, then it will set to the default value of 50051.

Dockerfile
```diff
@@ -69,6 +69,8 @@ FROM base AS runtime
 
 USER forseti
 
+ENV PORT 50051
```
hshin-g: Does this mean that PORT variable is 50051? Or that default is 50051?

kevensen: What this does is expose the environment variable to be set by the container run time.  If it is not set there (e.g. not set via Kubernetes or Cloud Run), then it defaults to 50051.

I set a default in two places because I am being overly cautious.

hshin-g: Thanks for the explanation!

install/scripts/docker_entrypoint.sh
hshin-g: Thanks for the explanation!

PR #3502, opened 2019-11-26
Excluded files that shouldn't be included in the output

red2k18:
Excluding these files cuts down the build time by ~1.25 minutes.


joecheuk:
Not sure if I understand the context of this PR, do you have an issue describing the problem we are trying to address here?

red2k18:
This is one of the optimization to speedup the website build.

_config.yml
```diff
@@ -28,7 +28,11 @@ exclude: ["Gemfile",
           "scripts/",
           ".gitignore",
           ".sass-cache/",
-          ".idea/"]
+          ".idea/",
```
blueandgold: This `.idea` file is not checked in. So, why should this be excluded?

_config.yml
```diff
@@ -28,7 +28,11 @@ exclude: ["Gemfile",
...
+          "README.md",
```
blueandgold: This is already excluded at above.  So no need to repeat here.

_config.yml
```diff
@@ -28,7 +28,11 @@ exclude: ["Gemfile",
...
+          ".ruby-version",
+          "_includes/site",
```
blueandgold: This has the html files.  Can you explain what they are used for, and why they should be excluded?

_config.yml
```diff
@@ -28,7 +28,11 @@ exclude: ["Gemfile",
...
+          "_sass"]
```
blueandgold: These `.scss` files look like they might be used?

red2k18: `scss` is generally excluded as this need not be included in the generated `_site`. Also, I have tested the changes on my local and there are no changes in the layout or the content display format after excluding these files.

_config.yml
red2k18: I will remove this one.

_config.yml
red2k18: I will remove this one.

_config.yml
red2k18: When the website is built, static site is generated and sent to a directory called `_site`. If `_site` is deleted it can be regenerated by building website as this is generated every time website is built. Hence, `_site` is not used and it need not be stored on GitHub. Tested changes on local and it works just fine.

blueandgold: The excluded resource is `_includes/site`, instead of `_site`.  Are we discussing the same thing?

Also, there are some references like this... is this referring to  `_includes/site`?  And if so, does the page that use this include still work?  See the screenshot below.
'{% include site/news/news_item.html post=post %}`

image

red2k18: By default, all the files in your in your Jekyll project are included in the output. In `_config.yml` file we can exclude files and directories that shouldn't be included in the output which Jekyll will then ignore. [Here](https://jekyllrb.com/docs/configuration/options/) is the documentation on `exclude` straight from the Jekyll site.

PR #3503, opened 2019-11-26
Removed reference folder

red2k18:
Removing reference folder that is manually generated while building website during the release. This folder is not really referred to by developers and users. 

image


blueandgold:
This will remove over 6000 files which are too large for me to view.  Can you confirm that these are the API docs that are generated?  And which versions (2.18, 2.19, etc) are you removing?  Are these really generated manually or automatically?  If we limit our build versions in the future to 3 versions, would the old reference docs be automatically deleted?

red2k18:
In this PR I have removed the API docs that were manually generated at the time of release testing for all the versions except latest version. I will go ahead and comment out the step that generates API docs in the release testing doc so that it is not generated in the future. Limiting the number of future releases will not delete the already generated docs; it needs to be explicitly deleted.

blueandgold:
Can you please break up this PR into smaller ones, so that I can review all the files?  i.e. one PR for each version, and also another PR for the change to prevent this from being generated in the future.

red2k18:
Closing this PR as I broke down the changes in chunks so that it is easier to review. Below are the PRs I opened.
1. [Removes API docs from v2.23.0](https://github.com/forseti-security/forseti-security/pull/3527)
2. [Removes API docs from v2.24.0](https://github.com/forseti-security/forseti-security/pull/3528)
3. [Removes API docs from the latest version](https://github.com/forseti-security/forseti-security/pull/3529)

PR #3505, opened 2019-11-27
Create a simplified container for the forseti-server

kevensen:
The resulting container image does not rely on install/scripts/docker_entrypoint.sh.

This PR neither removes nor alters existing functionality.


build/cloudbuild-master.yaml
```diff
@@ -24,9 +24,13 @@ steps:
...
   id: Build forseti-security master container image
-  args: ['build', '-t', 'gcr.io/$PROJECT_ID/forseti:master', '.']
+  args: ['build',  '--target', 'runtime', '-t', 'gcr.io/$PROJECT_ID/forseti:master', '.']
+- name: 'gcr.io/cloud-builders/docker'
+  id: Build forseti-server master container image
+  args: ['build', '--target', 'forseti-server', '-t', 'gcr.io/$PROJECT_ID/forseti-server:master', '.']
 
 images:
+- gcr.io/$PROJECT_ID/forseti-server:master
 - gcr.io/$PROJECT_ID/forseti:master
```
gkowalski-google: Will the forseti image refer to the client only now? If so, it might be better to rename this image as forseti-client.

build/cloudbuild-master.yaml
```diff
@@ -24,9 +24,13 @@ steps:
...
 - gcr.io/$PROJECT_ID/forseti-base:master
 - gcr.io/$PROJECT_ID/forseti-prebuild:master
```
gkowalski-google: Nothing specific to this PR, but I am curious if we really need to publish the base and prebuild images?

kevensen: Honestly, not sure.  I thought it might speed up builds in Cloud Build, but it isn't.  It's worth considering removing them.

build/cloudbuild-master.yaml
kevensen: @gkowalski-google I'd like to hold off on renaming it here.  I don't think I'll be able to update the helm-charts in time for the next release to include the `forseti-server` container image anyway.  This way, the helm-charts and Terraform module will still be able to use the image named `forseti`.

kevensen:
@gkowalski-google Thanks for the review!

build/cloudbuild-master.yaml
gkowalski-google: OK makes sense, should we create a ticket to document what else needs to be completed for this change?

kevensen: @gkowalski-google I agree.  I'll do that now.

kevensen: https://github.com/forseti-security/forseti-security/issues/3511

PR #3507, opened 2019-11-28
Enable caching for all responsive images

red2k18:
Enabled caching for all the images and disabled saving the files to the source folder to speed up website build. `responsive_image` tag is already being used. 

This change cuts down the website build time by one minute.


_config.yml
```diff
@@ -79,6 +79,8 @@ responsive_image:
     - width: 1400
   default_quality: 95
   strip: true
+  save_to_source: false
+  cache: true
```
blueandgold: Can you point me to the documentation for these 2 configurations?

red2k18: Here's[https://github.com/wildlyinaccurate/jekyll-responsive-image#all-configuration-options](url) all of the available configuration options for the responsive images straight from the [Jekyll plugin official page](https://github.com/wildlyinaccurate/jekyll-responsive-image) that we're already using.

blueandgold: Looks good overall, please add In the PR description, what is the time improvement from this change.

red2k18: Done

PR #3508, opened 2019-11-28
Added sassc gem and jekyll-minifier

red2k18:
`jekyll-minifier` minifies HTML, XML, CSS, JSON and Javascript.
`sassc` gem compiles SASS files more efficiently.


red2k18:
Closing this as `jekyll-minifier` only works if `JEKYLL_ENV` is set. Opened #3509 instead.

PR #3509, opened 2019-11-28
Added sassc gem and updated config to compress sassc output. 

red2k18:
'sassc` gem compiles SASS files more efficiently. 

This cuts down the build time by about 15 seconds.


_config.yml
```diff
@@ -80,6 +80,10 @@ responsive_image:
   default_quality: 95
   strip: true
 
+# Compress Sass
+sass:
+  style: compressed
```
blueandgold: Can you please give me a link that explains the benefit of this gem and using compressed?

red2k18: 1. https://jekyllrb.com/docs/assets/
2. https://github.com/sass/sassc-ruby
3. Please see `sass` section on [this](https://cloudcannon.com/jekyll/2019/09/03/whats-new-jekyll-4/) page.

blueandgold:
Can you please add the speed improvement to the PR description?

PR #3516, opened 2019-12-03
Added guard-jekyll-plus plugin for smarter Jekyll builds

red2k18:
Below are the features of Guard Jekyll Plus plugin:

1. Changing static files won't trigger a jekyll build
2. Batched processing
3. Reads options from your YAML config file(s)
4. Supports multiple config files
5. Serve with Jekyll or Rack
6. Clear and colorized output.


red2k18:
Here [https://github.com/imathis/guard-jekyll-plus](url) is the README.md of the guard-jekyll-plus plugin itself. PTAL.

blueandgold:
Can you please add the speed improvement to the PR description?  I am interested to know if we got the same improvement as in the article.

red2k18:
Build time is about the same after adding this plugin as in our case it is proportional to the copies of documentation we maintain on the website.

Guardfile
```diff
@@ -0,0 +1,8 @@
+# The Guardfile
+# Taken directly from https://github.com/imathis/guard-jekyll-plus
+
+ignore /^_site/ # NOTE: this can interfere with Guard::LiveReload
+
+guard "jekyll-plus", :serve => true do
+  watch /.*/
+end
```
blueandgold: add a newline here

_config.yml
```diff
@@ -26,9 +26,10 @@ exclude: ["Gemfile",
           ".travis.yml",
           "travis/",
           "scripts/",
-          ".gitignore",
+          ".git:ignore",
```
blueandgold: this is correct?

Guardfile
red2k18: Done

_config.yml
red2k18: Fixed it

red2k18:
Closing this. Open #3546 instead.

PR #3529, opened 2019-12-11
Removed references for API docs for latest version

red2k18:
This PR removes the API docs from the latest version. 

API docs are not among the top 100 pages visited by the users in the last 90 days.


red2k18:
I believe API docs are generated automatically for the `latest` version. Need to look for the script that has the logic to generate them. I believe removing that logic from that script will also delete API docs for the `latest` version. This PR is removing references to the API docs.

PR #3544, opened 2019-12-16
Add an endpoint to run the server processes end-to-end.

### LINKED ISSUE #3342: Server Side Run Logic
As a user of Forseti, I would like to invoke the inventory creation, model creation, scan and notify in one CLI command in order to simplify the invocation of a Forseti run.

As a maintainer of Forseti, I would like to remove the two run scripts (one for [GCE](https://github.com/forseti-security/forseti-security/blob/master/install/gcp/scripts/run_forseti.sh) and one for [GKE](https://github.com/forseti-security/forseti-security/blob/master/install/scripts/docker_entrypoint.sh) in order to simplify maintenance.

As a developer of Forseti, I would like to have a single server side API endpoint to invoke inventory creation, model creation, scan and notify in order to simplify eventing inputs (grpc and those not yet designed) and set Forseti up for future architectures.

kevensen:
Resolves #3342 


google/cloud/forseti/services/server_config/service.py
```diff
@@ -26,6 +28,64 @@
...
+    """Runs Forseti.
+
+    Args:
+        client (ClientComposition): Forseti client object.
+
+    Yields:
+        ServerRunReply: the returned proto message.
+    """
+
+    model_name = 'run_model'
```
gkowalski-google: I think the model name should be dynamic. Maybe prefix it with RunModel and append the timestamp to the end?

kevensen: Yes.  That is a very good suggestion. Will do.

kevensen:
@gkowalski-google just FYI, I am going to close this PR in favor of another which I will open this morning.

kevensen:
@gkowalski-google Please disregard my last.  I want to complete this PR.  I'll make the fix you suggested.

google/cloud/forseti/services/server_config/service.py
```diff
@@ -26,6 +29,48 @@
...
+    Returns:
+        ServerRunReply: the returned proto message.
+    """
+    today = datetime.now()
+    model_name = today.strftime('run_model_%Y%m%d_%H%M')
+    message = 'Forseti server run complete'
+
+    inventory_id = 0
+
+    for progress in client.inventory.create():
```
blueandgold: Should you check for the status of the inventory, so that if there are errors then don't proceed to build a model or the rest of the e2e run?

google/cloud/forseti/services/server_config/service.py
```diff
@@ -26,6 +29,48 @@
...
+        False)
+
+    model_handle = model_reply.model.handle
+    model_status = model_reply.model.status
+
+    client.switch_model(model_handle)
+
+    if model_status in ['SUCCESS', 'PARTIAL_SUCCESS']:
+        for progress in client.scanner.run(scanner_name=None):
+            pass
```
blueandgold: If we are just passing here, can we just do below, instead of looping?
```
client.scanner.run(scanner_name=None)
client.notifier.run(inventory_id, 0)
```

google/cloud/forseti/services/server_config/service.py
```diff
@@ -26,6 +29,48 @@
...
+        model_name,
+        inventory_id,
+        False)
+
+    model_handle = model_reply.model.handle
+    model_status = model_reply.model.status
+
+    client.switch_model(model_handle)
+
+    if model_status in ['SUCCESS', 'PARTIAL_SUCCESS']:
```
blueandgold: The code would be easier to read if you go like this:
```
if model_status not in ['SUCCESS', 'PARTIAL_SUCCESS']:
    message = 'ERROR: Model status is {}'.format(model_status)
    return server_pb2.ServerRunReply(message=message)
```

google/cloud/forseti/services/server_config/service.py
```diff
@@ -26,6 +29,48 @@
...
+
+    Args:
+        client (ClientComposition): Forseti client object.
+
+    Returns:
+        ServerRunReply: the returned proto message.
+    """
+    today = datetime.now()
+    model_name = today.strftime('run_model_%Y%m%d_%H%M')
+    message = 'Forseti server run complete'
```
blueandgold: Move this message to be closer to the `return`, where it it used. around line 71.

google/cloud/forseti/services/server_config/service.py
```diff
@@ -26,6 +29,48 @@
...
+    """Runs Forseti.
+
+    Args:
+        client (ClientComposition): Forseti client object.
+
+    Returns:
+        ServerRunReply: the returned proto message.
+    """
+    today = datetime.now()
+    model_name = today.strftime('run_model_%Y%m%d_%H%M')
```
blueandgold: Can you move these 2 lines to be closer to where they are actually being used?  i.e. where the model is created?

google/cloud/forseti/services/cli.py
```diff
@@ -233,6 +233,11 @@ def define_server_parser(parent):
               'the server vm or a gcs path starts with gs://).')
     )
 
+    action_subparser.add_parser(
+        'run',
+        help='Run the Forseti process, end-to-end.'
```
blueandgold: Please also change the title of this PR to be more descriptive of the work done here, so that when folks are scanning through the PRs, they can easily find the one that they are looking for.  So, something like:
```
Add an endpoint to run the server processes end-to-end.
```

google/cloud/forseti/services/server_config/service.py
kevensen: That's probably a good idea.

google/cloud/forseti/services/cli.py
kevensen: I like it.

kevensen: Done

google/cloud/forseti/services/server_config/service.py
kevensen: Done

google/cloud/forseti/services/server_config/service.py
kevensen: Done

google/cloud/forseti/services/server_config/service.py
kevensen: Done

google/cloud/forseti/services/server_config/service.py
kevensen: That would certainly be cleaner. Because the gRPC reply to a run request is a stream, we need to consume the entire stream before proceeding.  Without the loop, the execution would continue without waiting for the scan to complete, for example.

google/cloud/forseti/services/server_config/service.py
kevensen: Yes

google/cloud/forseti/services/server_config/service.py
```diff
@@ -26,6 +29,54 @@
...
+        inventory_id,
+        False)
+
+    model_handle = model_reply.model.handle
+    model_status = model_reply.model.status
+
+    client.switch_model(model_handle)
+
+    if model_status not in ['SUCCESS', 'PARTIAL_SUCCESS']:
+        message = 'ERROR: Model status is {}'.format(model_status)
```
blueandgold: I think that you need to add a step to delete model here.  And also a return the message step.

Issue #3354, opened 2019-10-22
Verify Explain resource count matches GCP resource count


gkowalski-google:
## Story
Forseti Explain uses the database model to provide helpful info about the GCP organization. We want to verify that the resources created by the model match what is in GCP. A good test of this is the projects.

## Proposed Solution
When the design for preventing violations for deleted resources is complete and implemented, compare the counts of the following commands and ensure they match.
```
forseti explainer list_resources | grep -c \"project
```
vs
```
gcloud projects list
```

## Acceptance Criteria
GIVEN: An inventory is created
WHEN: Explainer lists the resources
THEN: The number of projects match the output from gcloud CLI

goofiva:
https://github.com/forseti-security/forseti-security/tree/feature/issues-3354

goofiva:
This ticket should be ready for review the moment the https://github.com/forseti-security/forseti-security/issues/2993 ticket is complete.

goofiva:
@gkowalski-google bump.

Issue #2993, opened 2019-07-24
Prevent violations being sent for resources that are pending deletion


rich-boyce:
- Forseti version: 2.17.0
 - Module: notifier or scanner
 - No relevant logs

It seems like Forseti should _not_ send violations for resources that are in the state 'pending_delete'.

Example: A violation is raised for a Cloud SQL instance that is accessible by 0.0.0.0/0. This particular instance was created for testing purposes and is no longer required, so it is deleted. The resource is put into state 'pending_delete', and immediately becomes invisible from the GCP user interfaces, but is still visible to Forseti. Violations continue to be raised by Forseti until the resource is actually removed, approximately a day later. This can cause confusion.

This bug raised as requested by @blueandgold

blueandgold:
Thanks for filing this!  I would think that making the change in the scanner would make sense, as we don't want the violation to be created in the first place.

joecheuk:
Hi @rich-boyce , thanks for reporting the issue, I was doing some investigation but can't seem to find `pending_delete` state  (or any state related to delete) for cloud sql instance.

ref: https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/instances#resource

state | string | The current serving state of the Cloud SQL instance. This can be one of the following. **RUNNABLE**: The instance is running, or is ready to run when accessed. **SUSPENDED**: The instance is not available, for example due to problems with billing. ### PENDING_CREATE: The instance is being created. `MAINTENANCE`: The instance is down for maintenance. **FAILED**: The instance creation failed. **UNKNOWN_STATE**: The state of the instance is unknown.
-- | -- | --

Do you mind providing more information on where to find the field? We consume data directly from cloud asset inventory API so there is a chance they still have that data cached on their end but without a field clearing stating that the instance is gone, it's hard for us to decide when to remove a field during inventory/scanning time.

hshin-g:
This issue needs more research / verification that the CloudSQL instance being deleted is generating violations.

kevensen:
The field for checking whether a CloudSQL instance is pending deletion is actually found in the the [operations response](https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/instances/delete#response) after a delete action.

- status: PENDING
- operationType: DELETE

blueandgold:
[...]

PR #3556, opened 2019-12-18
Verify Explain resource count matches GCP resource count

goofiva:
Adding test. Will still require #2993 to be fixed to run consistently. However issues-2993 is not a blocker. Rather, this test can help make sure issues-2993 has been fixed.


gkowalski-google:
Resolves #3354

PR #3581, opened 2020-01-13
Change the violation type for Config Validator Scanner

### LINKED ISSUE #2734: Update Config Validator violation category
## Story
All the config validator violations has violation type 'CONFIG_VALIDATOR_VIOLATION' but we should separate them on a more granular level, maybe per rule name instead since the FCV instance can contain multiple rules and the resulting violation file can be huge and hard to read for users.

## Proposed Solution
Update the violation category to be `CV_{RULE_NAME}`.

## Acceptance Criteria
- Existing findings in CSCC do not get duplicated
- New CV findings in CSCC have the new category
- [Update CV end-to-end test assertions](https://github.com/forseti-security/forseti-security/blob/master/integration_tests/tests/forseti/controls/scanner-config_validator_gcs_policy.rb#L63)

red2k18:
Resolves #2734 

Today, Config Validator violations are labeled as CONFIG_VALIDATOR_VIOLATION. Making the violation type specific (generic today) by using the constraint name with CV as prefix to help identify Config Validator violations as well as the constraint that is violated.

### DIFF
--- a/google/cloud/forseti/scanner/scanners/config_validator_scanner.py
+++ b/google/cloud/forseti/scanner/scanners/config_validator_scanner.py
@@ -83,7 +83,7 @@ class ConfigValidatorScanner(base_scanner.BaseScanner):
                 'full_name': full_name,
                 'rule_index': 0,
                 'rule_name': violation.constraint,
-                'violation_type': 'CV '+violation.constraint,
+                'violation_type': 'CV ' + violation.constraint,
                 'violation_data': json_format.MessageToDict(
                     violation.metadata, including_default_value_fields=True),
                 'resource_data': resource_data,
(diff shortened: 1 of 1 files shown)


google/cloud/forseti/scanner/scanners/config_validator_scanner.py
```diff
@@ -85,7 +83,7 @@ def _flatten_violations(self, violations):
                 'full_name': full_name,
                 'rule_index': 0,
                 'rule_name': violation.constraint,
-                'violation_type': self.VIOLATION_TYPE,
+                'violation_type': 'CV ' + violation.constraint,
```
hshin-g: Slight nit- would suggest putting 'CV: ' + violation.constraint instead to make it more clear that it is from CV.

red2k18: Thanks for the suggestion! Violation type/category of finding in CSCC usually do not have a colon(:) in it. They have underscore(_) instead.

PR #3628, opened 2020-02-05
Reorganized steps to install Forseti using Terraform

red2k18:
Reorganized steps to install Forseti using Terraform in an effort to make the external documentation easy to understand and easy to follow.

### DIFF
--- a/_docs/_latest/setup/install.md
+++ b/_docs/_latest/setup/install.md
@@ -41,13 +41,6 @@ can be used to provision the project with the required APIs enabled, along with 
 3. The Service Account used to execute this module has the right permissions.
 4. Enable the required GCP APIs to allow the Terraform module to deploy Forseti.
 
-## **Software Dependencies**
-
-### Terraform and Plugins
-- [Terraform](https://www.terraform.io/downloads.html) 0.12
-- [Terraform Provider for GCP](https://github.com/terraform-providers/terraform-provider-google) 2.11.0
-- [Terraform Provider Templates](https://www.terraform.io/docs/providers/template/index.html) 2.0
-
 ## **Google Cloud Shell Walkthrough**
 A Google Cloud Shell walkthrough has been setup to make it easy for users who are new to Forseti and Terraform. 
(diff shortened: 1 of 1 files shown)


_docs/_latest/setup/install.md
```diff
@@ -22,14 +22,33 @@ organization Cloud IAM policy.
...
+1. Install Terraform.
+2. A GCP project to deploy Forseti into. The 
+[Google Project Factory Terraform](https://github.com/terraform-google-modules/terraform-google-project-factory) module 
+can be used to provision the project with the required APIs enabled, along with a Shared VPC connection.
+3. The Service Account used to execute this module has the right permissions.
+4. Enable the required GCP APIs to allow the Terraform module to deploy Forseti.
+
+## **Software Dependencies**
+
+### Terraform and Plugins
```
gkowalski-google: (nit) I know this PR is just about reorganizing the docs, but I don't think we need to list the provider versions on the website. The user don't need to do anything to get those provider versions and as we update provider versions, we will need to update the website.

_docs/_latest/setup/install.md
```diff
@@ -79,7 +105,7 @@ Create a file named `main.tf` in an empty directory and copy the contents below
 ```hcl
 module "forseti" {
   source  = "terraform-google-modules/forseti/google"
-  version = "~> 5.0.0"
+  version = "~> 5.2.0"
```
gkowalski-google: 5.2 is not out yet, should this be part of the 2.25 release? Omit the micro version from this syntax, otherwise we will need to update this version for every release. This can be `~> 5.0` or `~> 5.1` which will use the latest release up until 6.x.

red2k18: Updated version to `5.1`.

_docs/_latest/setup/install.md
red2k18: I too think `Software Dependencies` section is not required. I went ahead and removed it in this PR itself.

PR #3632, opened 2020-02-05
[WEBSITE] Update Slack signup links on website

### LINKED ISSUE #3604: Update Slack Link on website
## Story
The Slack link on the website has expired.

## Solution
Create a new Slack link and update everywhere on the website.

hshin-g:
Resolves #3604.

### DIFF
--- a/_docs/v2.23/setup/migrate.md
+++ b/_docs/v2.23/setup/migrate.md
@@ -11,7 +11,7 @@ deprecated Python Installer to the new Terraform module.
 If you have any
 questions about this process, please contact us by
 [email](mailto:discuss@forsetisecurity.org) or on
-[Slack](https://forsetisecurity.slack.com/join/shared_invite/enQtNDIyMzg4Nzg1NjcxLTM1NTUzZmM2ODVmNzE5MWEwYzAwNjUxMjVkZjhmYWZiOGZjMjY3ZjllNDlkYjk1OGU4MTVhZGM4NzgyZjZhNTE).
+[Slack](https://join.slack.com/t/forsetisecurity/shared_invite/enQtOTM4NTkwMDcwMDA1LTk1ZDExYTExZTJlNjY3NjIwZmVhZmJkMjk3YzVhZmYwNGRmYmU0N2UzZDc2Njg4NmEwYWU4ODc3MWI3NjJkZTE).
 
 ---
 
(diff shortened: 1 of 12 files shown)


_docs/v2.23/setup/migrate.md
```diff
@@ -11,7 +11,7 @@ deprecated Python Installer to the new Terraform module.
 If you have any
 questions about this process, please contact us by
 [email](mailto:discuss@forsetisecurity.org) or on
-[Slack](https://forsetisecurity.slack.com/join/shared_invite/enQtNDIyMzg4Nzg1NjcxLTM1NTUzZmM2ODVmNzE5MWEwYzAwNjUxMjVkZjhmYWZiOGZjMjY3ZjllNDlkYjk1OGU4MTVhZGM4NzgyZjZhNTE).
+[Slack](https://join.slack.com/t/forsetisecurity/shared_invite/enQtOTM4NTkwMDcwMDA1LTk1ZDExYTExZTJlNjY3NjIwZmVhZmJkMjk3YzVhZmYwNGRmYmU0N2UzZDc2Njg4NmEwYWU4ODc3MWI3NjJkZTE).
```
blueandgold: Instead of joining, how about a more generic link that doesn't expire?  like `http://forsetisecurity.slack.com`?

_includes/docs/v2.23/resource-coverage.md
```diff
@@ -44,7 +44,7 @@
 | [Compute Engine Instance Templates](https://cloud.google.com/compute/docs/reference/latest/instanceTemplates) |
 | [Compute Engine Interconnects](https://cloud.google.com/compute/docs/reference/rest/v1/interconnects) |
 | [Compute Engine Interconnect Attachments](https://cloud.google.com/compute/docs/reference/rest/v1/interconnectAttachments) |
-| [Compute Engine Licenses](https://cloud.google.com/compute/docs/reference/rest/v1/licenses/get) |
+| Compute Engine Licenses |
```
blueandgold: I do see this 404 publicly.  Is this a temporary problem on their end, but will come back?  I still see it in the search result.

_docs/v2.23/setup/migrate.md
hshin-g: This invite is set to never expire, so we technically should not run into this issue in the future. I think you need to be invited to the slack workspace in order to join (can only sign in), so not sure that link would work.

_includes/docs/v2.23/resource-coverage.md
hshin-g: According to the Compute Engine team, the API is not changing, but they are removing the public documentation because it is not useful for end users and has caused confusion.

blueandgold: Sounds good.

_docs/v2.23/setup/migrate.md
blueandgold: Sounds good, glad it won't expire.

PR #3640, opened 2020-02-10
Added instructions to run Forseti from a non-org root using Terraform

red2k18:
Added instructions to run Forseti from a non-org root using Terraform

### DIFF
--- a/_docs/_latest/configure/general/non-org-root.md
+++ b/_docs/_latest/configure/general/non-org-root.md
@@ -28,57 +28,41 @@ manually assign the correct roles later.
 
 ## Configure Forseti to Run on a Folder
 
-1. Edit `forseti_conf_server.yaml` and point the `root_resource_id`
-to the target folder:
-`folders/<foo_folder_id>`.
+1. Edit `main.tf` and set `composite_root_resources` variable to the target 
+folder: `["folders/<foo_folder_id>"]`.
 
-1. **NEW for version 2.12.0+**: You can use the `composite_root_resources`
-   configuration to include multiple resources in a single Forseti installation.
-   See [Configure Inventory]({% link _docs/latest/configure/inventory/index.md %})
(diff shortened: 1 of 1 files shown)


_docs/_latest/configure/general/non-org-root.md
```diff
@@ -28,57 +28,38 @@ manually assign the correct roles later.
 
 ## Configure Forseti to Run on a Folder
 
-1. Edit `forseti_conf_server.yaml` and point the `root_resource_id`
-to the target folder:
-`folders/<foo_folder_id>`.
+1. Edit `main.tf` and set `composite_root_resources` variable to the target
```
blueandgold: nit: `...point to the target folder`

_docs/_latest/configure/general/non-org-root.md
```diff
@@ -28,57 +28,38 @@ manually assign the correct roles later.
...
+include one or more resources from GCP resource hierarchy in a single Forseti 
+installation.
 See [Configure Inventory]({% link _docs/latest/configure/inventory/index.md %})
 for more details.
 
-1. This assumes that Forseti is not installed with Org Admin credential, and
-you want Forseti to run on projects that you own. If Forseti is installed
-with Org Admin credential, then all the resources in the organization
-will be returned.
+For example: `composite_root_resources:` `["folders/<foo_folder1_id>", "projects/<foo_project1_id"]`
```
blueandgold: nit: we should not use folder as example here

_docs/_latest/configure/general/non-org-root.md
red2k18: Done

_docs/_latest/configure/general/non-org-root.md
red2k18: Updated

PR #3644, opened 2020-02-11
Updated command to restart Forseti in v2.23.1

red2k18:
Updated command to restart Forseti in v2.23.1


red2k18:
Verified on GCP instance and checked the database.

PR #3648, opened 2020-02-14
dedupe violations

fathom-ragib:

- [ ] All of the [tests](https://forsetisecurity.org/docs/latest/develop/dev/testing.html) pass.
- [ ] I have submitted a corresponding PR for documentation in the `forsetisecurity.org-dev branch` and included this on this PR (if applicable).
- [ ] My documentation has been functionally tested and used (if applicable).


red2k18:
Hi @fathom-ragib ,  `find_violations()` method finds all the violations and stores it in a `set` which does not hold duplicate items or violations in this case.

fathom-ragib:
Hi @red2k18 , I see that the violations are fed into a set here (I am assuming for deduplication purposes): https://github.com/forseti-security/forseti-security/blob/v2.4.0/google/cloud/forseti/scanner/audit/iam_rules_engine.py#L154

However, when I use the following yaml for iam_rules, I receive duplicated violations:

`
rules:
  - name: Prevent public users from having any role bindings in projects
    mode: blacklist
    resource:
      - type: folder
        applies_to: self_and_children
        resource_ids: ['*']
    inherit_from_parents: true
    bindings:
      - role: '*'
        members:
          - user:*
          - group:*
          - domain:*

  - name: Rule duplicated from above
    mode: blacklist
    resource:
      - type: folder
        applies_to: self_and_children
        resource_ids: ['*']
    inherit_from_parents: true
    bindings:
      - role: '*'
        members:
          - user:*
          - group:*
          - domain:*
 `

Looks like the `RuleViolation` namedtuple contains the `rule_name` and `rule_index` columns which would explain why the violations are not being deduplicated: https://github.com/forseti-security/forseti-security/blob/v2.4.0/google/cloud/forseti/scanner/audit/rules.py#L149-L160. I think the `resource_id`, `role` and `members` fields are enough to uniquely identify a violation. 

Imho, a cleaner option (instead of using a `registry` as I do in my PR) would be to change the `RuleViolation` namedtuple to a class-style `NamedTuple` and override the __hash__ function in the class (as I do below), although I haven't really tested this option.

`
from typing import NamedTuple, Tuple

from google.cloud.forseti.common.gcp_type.iam_policy import IamPolicyMember


class RuleViolation(NamedTuple):
    resource_type: str
    resource_id: str
    full_name: str
    rule_name: str
    rule_index: int
    violation_type: str
    role: str
    members: Tuple[IamPolicyMember]
    resource_data: str
    
    def __hash__(self):
        return hash((self.resource_id, self.role, self.members))
`

Also, I am using v2.4.0, so if this issue was fixed in later versions please let me know.

red2k18:
Hi @fathom-ragib , thanks for the detailed information! @gkowalski-google is on interrupts and will be investigating this issue. A new issue will be filed after investigation if required.

gkowalski-google:
@fathom-ragib Thanks for the info. I understand how these violations are being duplicated. Although I'm not sure this is an issue or that we would want to de-dupe violations. If a user does have multiple rules that check for similar policies, and they produce violations for the same resource, then maybe that is OK. Perhaps it would be better to adjust the rules so that they are unique and don't produce duplicates? Do you have another example of some rules that you would like to have, which would produce duplicates?

blueandgold:
@fathom-ragib I do agree with @gkowalski-google that we should not have multiple rules that would produce the same violations, i.e. would be better to de-dupe the rules.

blueandgold:
@fathom-ragib I'm going to go ahead and close this PR, per the comments above.  If you have different perspective, please let us know and we can revisit this.

fathom-ragib:
Hi @blueandgold, @gkowalski-google. Thank you for taking the time to look at this issue and sorry for the late reply. The example I used was just a toy yaml. A more realistic scenario would be the following.

```
rules:
  - name: Prevent users from having any role bindings
    mode: blacklist
    resource:
      - type: folder
        applies_to: self_and_children
        resource_ids: ['*']
    inherit_from_parents: true
    bindings:
      - role: '*'
        members:
          - user:*
          - group:*
          - domain:*

  - name: Project and bucket level role bindings can only be for service accounts
    mode: whitelist
    resource:
      - type: folder
        applies_to: self_and_children
        resource_ids: ['*']
    inherit_from_parents: true
    bindings:
      - role: 'roles/deploymentmanager.editor'
        members:
          - <enter Google service account used for Google Deployment Manager>
```

In the example above, we are 

1. blacklisting every account
2. whitelisting specific service accounts for the specific roles that they require

We do this to explicitly keep stock of which (service_account, role) bindings are whitelisted. However, the way the Forseti IAM Engine currently works, the whitelist rules implicitly blacklist other accounts that are not included in the rule. For the example given above, the IAM Engine would first blacklist me for the `roles/deploymentmanager.editor` role due to the first blacklist rule in the yaml. And then the engine would blacklist me a second time due to the implicit blacklisting caused by the whitelist rule, resulting in duplicated violations for the same binding.

Please let me know if any further clarification is required.

blueandgold:
Thanks @fathom-ragib for the explanation.  It makes sense to me.  Let's go through the code review process, and get this merged in.

blueandgold:
@fathom-ragib one more question here, it might be beneficial to know all the violations produced by a rule.  That's why the rule reference is provided as part of the violation output.  If we de-dupe this, then we may not see every violation produced by a rule.  We only end up seeing violations from some rules.  Is this acceptable to you?  Overall, let's make this configurable for people to enable or disable based on their preference (dedupe default is false, i.e. preserve the current behavior)

google/cloud/forseti/scanner/audit/iam_rules_engine.py
```diff
@@ -152,7 +152,14 @@ def find_violations(
         violations = self.rule_book.find_violations(
             resource, policy, policy_bindings)
 
-        return set(violations)
+        # dedupe violations
```
blueandgold: As mentioned in the PR comment, would be great to make this de-dupe feature to be an option, and there is a condition here to check that option before de-duping.  Are you familiar with how to make this optional?  There are examples in the [forseti_conf_server.yaml.in](https://github.com/forseti-security/forseti-security/blob/master/configs/server/forseti_conf_server.yaml.in) file, where the values are [populated by terraform like here](https://github.com/forseti-security/terraform-google-forseti/blob/master/modules/server_config/variables.tf).  If you need more exact instructions, please let us know.

blueandgold:
HI @fathom-ragib, are you able to add unit test, and make this an optional feature as mentioned before?  I will close this for now, but please feel free to re-open if you do the above.

Issue #3601, opened 2020-01-24
Add Website Documentation for Scanners


blueandgold:
## Story
Resource Scanner and Role Scanner are not documented on our website.  Please add these and look through other scanners that also need to be added.

https://forsetisecurity.org/docs/latest/configure/scanner/descriptions.html

## Solution
- Descriptions are added on the website for resource scanner and role scanner.
- Look through [all the scanners that we have](https://github.com/forseti-security/forseti-security/tree/master/google/cloud/forseti/scanner/scanners), and ensure there are corresponding descriptions on the website.

PR #3668, opened 2020-02-25
Enabling default scanning for the scanners enabled by default

red2k18:
Resolves part of #3601 
Enabling default scanning for the scanners enabled by default


red2k18:
Closing this as we want users to uncomment the default rules as they need.

PR #3680, opened 2020-03-04
Update stale.yml

dekuhn:
Updating probot-stale configuration to be a bit more aggressive, we started with really relaxed settings initially.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -20,10 +20,10 @@ staleLabel: stale
 limitPerRun: 1
 
 # Configuration settings specific to pull requests
-# Details: pull requests older than 30 days will become stale, after 37 days of no activity the issue is closed.
-# Stale notice message will be posted to the pull request at 30 days.
+# Details: pull requests older than 14 days will be marked stale. After 21 days of no activity the issue is closed.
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -20,23 +20,23 @@ staleLabel: stale
...
     daysUntilClose: 7
     markComment: >
       This pull request has been automatically marked as stale because it has not had
       recent activity. It will be closed in 7 days if no further activity occurs. Thank you
       for your contributions.
 
 # Configuration settings specific to issues
-# Details: Issues older than 5 months will become stale at 152 days. After 182 days of no activity the issue is closed.
-# Stale notice message will be posted to the issue at 152 days.
+# Details: Issues older than 45 days will be marked stale. After 60 days of no activity the issue is closed.
```
blueandgold: I feel we should not be so aggressive here with issues, but okay if we are in a good position to tackle the issues as they come in.

dekuhn: Keep in mind the clock is reset if the issue is touched; so stale bot is only going to touch issues that have not comments / activity. It also gives the team and/or customer 15 days to respond once marked stale before it gets closed. I do feel this is still very reasonable.

PR #3728, opened 2020-04-04
Test/pytest model use

### LINKED ISSUE #3726: Migrate model use test to pytest
As part of migration from inspec to pytest, replace the model use test with a pytest equivalent. Setup the test so that other client tests can be gradually added.

gkowalski-google:
Resolves #3726


red2k18:
Looks good. Thanks for adding this! Can we run GCB build with these changes before merging it in? If you have already done so, can you add the screen shot of the status please?

red2k18:
Accidentally clicked `Close and comment` instead of `Comment`.

red2k18:
I see that integration tests run as part of this PR. LGTM.

PR #3733, opened 2020-04-14
Updated Bucket ACL Scanner documentation

### LINKED ISSUE #3518: Update Bucket ACL Scanner documentation
## Story
The Bucket ACL Scanner only works for legacy ACL on buckets, it does not work with IAM policy on buckets.

## Solution
Update the [Bucket ACL Scanner](https://forsetisecurity.org/docs/latest/configure/scanner/descriptions.html#bucket-acl-scanner) documentation to clearly note that this scanner does not work IAM policy on the buckets.

red2k18:
Resolves #3518 

Updated Bucket ACL Scanner documentation to include the edge case where this scanner doesn't work.

### DIFF
--- a/_docs/_latest/configure/scanner/descriptions.md
+++ b/_docs/_latest/configure/scanner/descriptions.md
@@ -36,15 +36,15 @@ For examples of how to define scanner rules, see the
 
 ## Bucket ACL scanner
 
-Access Control Lists (ACLs) are enabled on the Cloud Storage buckets that don't 
-have uniform bucket-level access. ACLs allow you to grant public access to your 
-Cloud Storage bucket and objects. The bucket ACL scanner supports a blacklist 
-mode, to ensure unauthorized users don't gain access to your Cloud Storage bucket.
-
-**Note:** When uniform bucket-level access is enabled on a Cloud Storage bucket,
-only bucket-level Cloud IAM permissions grant access to that bucket and the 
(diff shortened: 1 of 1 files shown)


_docs/_latest/configure/scanner/descriptions.md
```diff
@@ -36,9 +36,15 @@ For examples of how to define scanner rules, see the
 
 ## Bucket ACL scanner
 
-Cloud Storage buckets have ACLs that can grant public access to your
-Cloud Storage bucket and objects. The bucket scanner supports a blacklist mode,
-to ensure unauthorized users don't gain access to your Cloud Storage bucket.
+Access Control Lists (ACLs) are enabled on the Cloud Storage buckets that don't 
+have uniform bucket-level access. ACLs allow you to grant public access to your 
+Cloud Storage bucket and objects. The bucket ACL scanner supports a blacklist 
+mode, to ensure unauthorized users don't gain access to your Cloud Storage bucket.
```
gkowalski-google, line 42: It's good to mention the info about uniform bucket-level access. However, the intent was to specify that this scanner ONLY operates on the ACLs of a bucket. It does not scan for IAM policies on the buckets. Please keep the uniform bucket-level access though. The 2nd sentence is a little misleading; I think the previous verbiage about public access is better.

red2k18: Thanks for the comment! This was captured in the `Note` below. However, I rephrased and moved it up!

PR #3775, opened 2020-07-17
Fix model creation and e2e tests

### LINKED ISSUE #3773: Test dependencies break e2e tests
## Story
Recent updates to test dependencies have been released which break the pytest e2e tests.

Versions of setuptools > `47.2.0` and importlib-metadata (installed with pytest) are not compatible. The following error occurs when running any Forseti command:
```
ModuleNotFoundError: No module named 'google.cloud.forseti'
```

The [pip requirements file](https://github.com/forseti-security/forseti-security/blob/master/requirements-test.txt) for the e2e tests did not pin `protobuf==3.9.0` which Forseti requires. Google-cloud-storage recently started to install protobuf==3.12.2, and even though the gcs version has been pinned to 1.25.0, recently protobuf==3.12.2 started to get installed (still not quite sure why).
```
pkg_resources.DistributionNotFound: The 'protobuf==3.9.0' distribution was not found and is required by forseti-security
```

## Proposed Solution
- Pin setuptools to `47.2.0` in the main requirements
- Pin protobuf to `3.9.0` in the test requirements

### LINKED ISSUE #3774: Duplicated permissions listed in some IAM roles break Forseti model
## Story
Multiple users have reported that Forseti is unable to create a model: 

- #3767
- #3766 
- #3770 

Recently some IAM roles started to provide duplicate permissions. For example, the `roles/compute.networkAdmin` role now includes `networkservices.httpfilters.create` and `networkservices.httpFilters.create`. During the model creation, a dictionary is used to cache the permissions that have been collected; however the duplicate permission names (ignoring case) break [this code](https://github.com/forseti-security/forseti-security/blob/master/google/cloud/forseti/services/model/importer/importer.py#L1189). Sample error message:

```
sqlalchemy.exc.IntegrityError: (pymysql.err.IntegrityError) (1062, "Duplicate entry 'networkservices.httpfilters.create' for key 'PRIMARY'") [SQL: 'INSERT INTO `{{ MODEL }}_permissions` (name) VALUES (%(name)s)'] [parameters: ({'name': 'accessapproval.requests.approve'}, {'name': 'accessapproval.requests.dismiss'}, {'name': 'accessapproval.requests.get'}, {'name': 'accessapproval.requests.list'}, {'name': 'accessapproval.settings.get'}, {'name': 'resourcemanager.projects.get'}, {'name': 'resourcemanager.projects.list'}, {'name': 'accessapproval.settings.delete'}  ... displaying 10 of 3019 total bound parameter sets ...  {'name': 'resourcemanager.folders.update'}, {'name': 'resourcemanager.projects.create'})] (Background on this error at: http://sqlalche.me/e/gkpj)
```

The duplicate permission can be seen with gcloud:
```
gcloud iam roles describe roles/compute.networkAdmin | grep -i networkservices.httpFilters.create

- networkservices.httpFilters.create
- networkservices.httpfilters.create
```

## Proposed Solution
Update model creation to ignore the case of the permission.

## Acceptance Criteria
All tests pass, add unit/e2e test for some more model scenarios.

gkowalski-google:
Fix model for duplicate permissions, add e2e test to check role table is populated.

Pin setuptools to version that doesn't cause issues with pytest. Pin protobuf==3.9.0 for e2e tests.

Resolves #3774 
Resolves #3773

### DIFF
--- a/google/cloud/forseti/services/dao.py
+++ b/google/cloud/forseti/services/dao.py
@@ -305,8 +305,8 @@ def define_model(model_name, dbengine, model_seed):
         This is used to make MySQL column case sensitive by adding
         an encoding type.
         Args:
-            db_dialect (String): The db dialect.
-            column_size (Integer): The size of the column.
+            db_dialect (str): The db dialect.
+            column_size (int): The size of the column.
 
         Returns:
             String: Sqlalchemy String.
@@ -478,7 +478,8 @@ def define_model(model_name, dbengine, model_seed):
         """Row entry for an IAM permission."""
 
         __tablename__ = permissions_tablename
-        name = Column(String(256), primary_key=True)
+        name = Column(get_string_by_dialect(dbengine.dialect.name, 256),
+                      primary_key=True)
         roles = relationship('Role',
                              secondary=role_permissions,
                              back_populates='permissions')
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,3 +1,4 @@
 # End-to-end test requirements
+protobuf==3.9.0
 google-cloud-storage==1.25.0
 pytest==5.3.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,3 +14,6 @@ codecov==2.0.15
 pylint==1.9.4
 flake8==3.6.0
 future==0.17.1
+
+# Latest versions of setuptools cause issues when pytest is installed
+setuptools==47.2.0
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -67,11 +67,11 @@ suites:
             - notifier-inventory-summary-email
             - notifier-temp-file-deletion
             - scanner-bucket-acl-scanner
-            - scanner-config-validator
             - scanner-enabled-apis-scanner
             - scanner-enable-audit-logging-scanner
             - scanner-firewall-scanner
             - scanner-run
+            - server-pytest
           hosts_output: forseti-server-vm-ip
           user: ubuntu
           key_files:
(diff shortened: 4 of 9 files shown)


gkowalski-google:
@EricBeach I made some changes here. Instead of converting the permission names to lowercase to avoid the duplicates, I found that other columns in the model table are using a different collation which allows for the duplicate names to be added to the database without hitting the unique constraint. I found that converting to lowercase all permission names might cause issues in other parts of the application. Moving the logic to the DB seems like the better option. All tests are passing now, I think we are good here. Would like your opinion 😄

google/cloud/forseti/services/dao.py
```diff
@@ -478,7 +478,8 @@ class Permission(base):
         """Row entry for an IAM permission."""
 
         __tablename__ = permissions_tablename
-        name = Column(String(256), primary_key=True)
+        name = Column(get_string_by_dialect(dbengine.dialect.name, 256),
```
gkowalski-google, line 481: Main change to fix the duplicate permission. Other columns in the model tables do this as well.

requirements-test.txt
```diff
@@ -1,3 +1,4 @@
 # End-to-end test requirements
+protobuf==3.9.0
```
gkowalski-google, line 2: Fix for e2e tests.

requirements.txt
```diff
@@ -14,3 +14,6 @@ codecov==2.0.15
 pylint==1.9.4
 flake8==3.6.0
 future==0.17.1
+
+# Latest versions of setuptools cause issues when pytest is installed
+setuptools==47.2.0
```
gkowalski-google, line 19: Fix for e2e tests.

EricBeach:
I looked again, and it looks fine to me. But, I don't really know the code base very well. If you feel uncertain, I'd ask Henry to review.

PR #3792, opened 2020-07-31
Update violation hash to include rule name

### LINKED ISSUE #3769: Config Validator violations produce duplicate findings for CSCC
## Story
Forseti uses a violation hash to use as the finding id in CSCC. Config Validator violation data typically just contains the resource name, so if a resource has multiple violations, only one of those violations will make it into CSCC.

## Proposed Solution
Add an option to include the constraint name in the violation data. By making it an option that is disabled by default, it would not affect any current users.

gkowalski-google:
Update violation hash to include rule name so that violations are correctly reported to CSCC. Without the rule name, some resources and violations were getting the same violation hash which is used as the CSCC finding id.

Update unit and e2e tests. Update protobuf and grpcio packages within setup.py to allow tests to install newer versions which are required by google-cloud-securitycenter. Update cloudbuild and terraform configuration for e2e tests to use pre-created service account and CSCC source ID. Update expected role count for Forseti SA.

Resolves #3769

### DIFF
--- a/endtoend_tests/forseti/notifier/cscc_test.py
+++ b/endtoend_tests/forseti/notifier/cscc_test.py
@@ -1,0 +1,71 @@
+# Copyright 2020 The Forseti Security Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""CSCC test"""
+
+import pytest
+from google.cloud import securitycenter
+from sqlalchemy.sql import text
+
+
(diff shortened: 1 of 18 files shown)


endtoend_tests/forseti/notifier/cscc_test.py
```diff
@@ -0,0 +1,71 @@
...
+        violations = (TestNotifierCloudSecurityCommandCenter.get_violations(
+                cloudsql_connection, scanner_id))
+
+        # Act
+        client = securitycenter.SecurityCenterClient()
+        finding_result_iterator = client.list_findings(
+            cscc_source_id, filter_='state="ACTIVE"')
+
+        # Assert
+        for _, finding_result in enumerate(finding_result_iterator):
```
gkowalski-google, line 67: This test will assert that each of the violations is found in CSCC.

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -233,6 +233,7 @@ def create(self, violations, scanner_index_id):
                 violation.get('full_name', ''),
                 violation.get('resource_data', ''),
                 violation.get('violation_data', ''),
+                violation.get('rule_name', '')
```
gkowalski-google, line 236: This file contains the functional changes, the rest is for testing.

gkowalski-google:
Forgot to update the GCB variables for the PR trigger. Updated and running tests again.

PR #3800, opened 2020-08-07
Include ancestor list in call to Config Validator

### LINKED ISSUE #3798: Config Validator errors with Kubernetes resources
## Story
An issue has been found scanning some resources with the main branch of Forseti and Config Validator. The following error is seen by the Config Validator scanner:
```
"serverMessage": "Error running scanner: ConfigValidatorScanner: 'Traceback
(most recent call last):\n  File
\"/usr/local/lib/python3.6/dist-packages/forseti_security-2.25.1-py3.6.egg/google/cloud/forseti/scanner/scanners/config_validator_util/validator_client.py\",
line 176, in review\n    return self.stub.Review(review_request).violations\n
File \"/usr/local/lib/python3.6/dist-packages/grpc/_channel.py\", line 565, in
__call__\n    return _end_unary_response_blocking(state, call, False, None)\n
File \"/usr/local/lib/python3.6/dist-packages/grpc/_channel.py\", line 467, in
_end_unary_response_blocking\n    raise _Rendezvous(state, None, None,
deadline)\ngrpc._channel._Rendezvous: <_Rendezvous of RPC that terminated
with:\n\tstatus = StatusCode.INTERNAL\n\tdetails = \"Received RST_STREAM with
error code 2\"
```

It has been found that certain Kubernetes resources are triggering this error and is linked back to [this PR on Config Validator](https://github.com/forseti-security/config-validator/pull/106/files). The following CAI assets can trigger this error (possibly other k8s assets as well):

- k8s.io/Namespace
- k8s.io/Node
- k8s.io/Pod
- k8s.io/Service
- rbac.authorization.k8s.io/ClusterRole
- rbac.authorization.k8s.io/ClusterRoleBinding
- rbac.authorization.k8s.io/Role
- rbac.authorization.k8s.io/RoleBinding

The issue is that CV is expecting Forseti to provide a [list of ancestors](https://github.com/forseti-security/config-validator/pull/106/files#diff-1c3aabca4a0bc3e4fd2da170d9fbe1c3R136) which is not provided by [Forseti in the call to CV](https://github.com/forseti-security/forseti-security/blob/master/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py#L125).

## Alternatives
1) Update Forseti TF module to use a CV release that does not have this change. Small change.
2) Update Forseti to provide ancestors by [parsing the resource full name](https://github.com/forseti-security/forseti-security/blob/master/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py#L41). This is a small change, but would provide the Project ID instead of Number. Org and Folder ancestors would be sent correctly. This could cause issues down the line with Forseti. Small change, [PR has been submitted](https://github.com/forseti-security/forseti-security/pull/3800).. need to look into tests.
3) Update CV to allow for null/empty ancestors. Small change.
4) Update Forseti to store and provide the asset ancestors from CAI. This a large change.

## Acceptance Criteria
- Perform a scan on a model that includes the above k8s resources and not have any errors.
- Perform some testing to check that kubernetes constraints are identifying violations correctly.

gkowalski-google:
Update Asset data sent to CV to include a list of ancestors generated from the full name of the resource.

Resolves #3798

### DIFF
--- a/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py
+++ b/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py
@@ -70,7 +70,7 @@ def generate_ancestors(full_name):
     """
     ancestors = []
     full_name_items = full_name.split('/')
-    for i in range(0, len(full_name_items) - 1):
+    for i in reversed(range(0, len(full_name_items) - 1)):
         if full_name_items[i] in SUPPORTED_ANCESTORS:
             ancestors.append(f'{full_name_items[i]}s/{full_name_items[i + 1]}')
     return ancestors
--- a/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py
+++ b/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py
@@ -72,7 +72,7 @@ def generate_ancestors(full_name):
     full_name_items = full_name.split('/')
     for i in range(0, len(full_name_items) - 1):
         if full_name_items[i] in SUPPORTED_ANCESTORS:
-            ancestors.append(f'{full_name_items[i]}/{full_name_items[i + 1]}')
+            ancestors.append(f'{full_name_items[i]}s/{full_name_items[i + 1]}')
     return ancestors
 
 
--- a/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py
+++ b/google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py
@@ -1,4 +1,4 @@
-# Copyright 2019 The Forseti Security Authors. All rights reserved.
+# Copyright 2020 The Forseti Security Authors. All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -30,6 +30,7 @@ LOGGER = logger.get_logger(__name__)
 _IAM_POLICY = 'iam_policy'
 _RESOURCE = 'resource'
 
+SUPPORTED_ANCESTORS = ['organization', 'folder', 'project']
 SUPPORTED_DATA_TYPE = frozenset([_IAM_POLICY, _RESOURCE])
 
 MOCK_CAI_RESOURCE_TYPE_MAPPING = {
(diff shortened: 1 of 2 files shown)


google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py
```diff
@@ -30,6 +30,7 @@
 _IAM_POLICY = 'iam_policy'
 _RESOURCE = 'resource'
 
+SUPPORTED_ANCESTORS = ['organization', 'folder', 'project']
```
briantkennedy, line 33: The CAI ancestors list uses plural (organizations, folders, projects), can you have generate_ancestors append the plural version when constructing the list to be more CAI-like?

tests/scanner/scanners/config_validator_util/cv_data_converter_test.py
```diff
@@ -90,6 +90,18 @@ def test_generate_ancestry_path_for_folder(self):
...
         self.assertEqual(expected_path, actual_path)
 
+    def test_generate_ancestors_for_org(self):
+        full_name = 'organization/1234567890/project/test-project-123/firewall/1234567890123456789/'
+        expected_ancestors = ['organization/1234567890', 'project/test-project-123']
+        actual_ancestors = cv_data_converter.generate_ancestors(full_name)
+        self.assertListEqual(expected_ancestors, actual_ancestors)
+
+    def test_generate_ancestors_for_folder(self):
+        full_name = 'folder/folder-1/project/project-2/lien/p123/'
```
briantkennedy, line 100: nit: folders are always a numeric ID, projects as given by CAI use project number for ancestry

gkowalski-google: That's a good catch, thanks. Perhaps instead of hacking this together I should just make the change to the Forseti inventory/model tables to store the CAI ancestors. Since Forseti was already using some custom code to determine the ancestry path, I thought it wouldn't be terrible to just do the same for the ancestors. Let me try going that route instead; unless you have other ideas.

briantkennedy: This is only test code, so updating the values here to numeric would be more of a best practice.  Overall this change makes sense until things get fixed up in the Forseti inventory model/tables (DB changes tend to be pretty heavyweight).

gkowalski-google: I have updated the folders to use the numeric ID and left the projects to use the ID instead of the numeric ID since that is what Forseti will be using. Org and folders will pass the numeric ID to CV, and the project ID.

google/cloud/forseti/scanner/scanners/config_validator_util/cv_data_converter.py
gkowalski-google: Updated PTAL. https://github.com/forseti-security/forseti-security/pull/3800/files#diff-e90bf6871d0fd206933197d0495ed8a6R75

gkowalski-google:
While looking into the previous build failure, I was able to reproduce the issue of CV not detecting the CloudSQL violations. Pushing up the change for the plural resource types as requested, and will see if the build is successful this time.

gkowalski-google:
The build failed because the no CV violations were found for the [SQL Location constraint](https://github.com/forseti-security/forseti-security/blob/master/integration_tests/fixtures/forseti/modules/policy_library/templates/policy-library/policies/constraints/cloudsql_location.yaml). The nightly builds have been reliably finding this violation, so there is an issue with this change. Running Forseti (this branch) and CV (any tag) locally, I can repro that no violation is found. If I switch to Forseti main branch or comment out the ancestors, then the violation is found. Nothing different is seen from the CV logs.

gkowalski-google:
Looks like the ancestors have to be placed in the correct order. Since the ancestry_path and ancestors are passed, the provided ancestry path is ignored and constructed from the ancestors. The highest ancestors are expected to be last in the list. While troubleshooting I have noticed that the CV protos are not correct for the version that is specified as the default in the Terraform module, and think these should be updated. Updating the protos requires `protobuf==3.12.4`; might do this in a separate PR or just do it as part of [this issue](https://github.com/forseti-security/terraform-google-forseti/issues/603). Will do some more testing on this and update this PR.

@briantkennedy Is it correct that since Forseti is passing Project ID (not number), if users wanted to specify a target matching the project, then they would need to set the constraint target to the ID? This change wouldn't be changing that, but that is a good thing to know for Forseti users since the behavior is different from other CV clients.

PR #3817, opened 2020-08-28
Update CSCC Notifier to send resource name in supported format

hazyd:
1. Send resource name in supported format
2. Change the logic of creating violation hash. This can help users to have a updated resource_name sending to CSCC. Currently resource_name for existing findings is immutable. 

Resolves #3817


google/cloud/forseti/notifier/notifiers/cscc_notifier.py
```diff
@@ -282,3 +283,58 @@ def run(self, violations, source_id=None):
...
+    def _cscc_resource_name_transformer(self, violation):
+        """Generate resource_name that CSCC accepts.
+
+        There are certain assets where CSCC's resource name
+        format differs from CAI's.
+        There are 4 cases where CSCC differ:
+        for "Cluster" assets
+        for "docker image" assets
+        for ALL compute type assets
+        for DNS managed zone and managed policy assets
```
aimjwizards, line 296: Please comment this case is not handled yet due to lack of data.

gkowalski-google:
I looked into the test logs. Some of the resource_name fields from the violations table can be empty. It looks like this will throw a 500 error from the CSCC API; if this was a required field I'd expect to see a 400 code, so it looks like this field cannot be empty. Perhaps use the transformed resource_name that you have added and then default to the `full_name` like was being used previously if resource_name is empty?

google/cloud/forseti/services/scanner/dao.py
```diff
@@ -233,7 +233,8 @@ def create(self, violations, scanner_index_id):
                 violation.get('full_name', ''),
                 violation.get('resource_data', ''),
                 violation.get('violation_data', ''),
-                violation.get('rule_name', '')
+                violation.get('rule_name', ''),
+                violation.get('resource_name', '')
```
gkowalski-google, line 237: I know we discussed if we should add in the resource_name to the hash, and I actually don't think it's required. The violation hash was recently updated (not released) to include the violated rule, so when the next release goes out, user's findings will get updated. The resource name doesn't hurt to have here, but not required.

gkowalski-google:
Tests are good now. We noticed that some assets were not showing up correctly in CSCC, such as LogSinks, BigQuery, and some Compute resources. Looking into that..

PeterYou:
Very good @gkowalski-google , 
Can not wait for the v2.26 release :)

PR #3820, opened 2020-09-02
feature/migrate-model-tests-to-pytest

### LINKED ISSUE #3819: Migrate model tests to pytest
Migrate the model tests to pytest as part of the migration from kitchen/inspec to pytest. This will help the tests to run faster and be more reliable.

gkowalski-google:
Migrate the model get, create, list, and delete tests to pytest. This will improve the performance reliability of the integration tests.

Resolves #3819


endtoend_tests/forseti/model/model_test.py
```diff
@@ -26,16 +26,90 @@ class TestModel:
...
+
+        # Assert model in database
+        query = text('SELECT '
+                     'state '
+                     'FROM forseti_security.model '
+                     'WHERE '
+                     f"name = '{model_name}'")
+        model = cloudsql_connection.execute(query).fetchone()
+        assert model
+        assert model[0] == 'PARTIAL_SUCCESS'
```
aimjwizards, line 52: Why partial success?

gkowalski-google: Good question.

If there are any warnings/errors during the inventory process, then these will be [recorded as warnings for the model](https://github.com/forseti-security/forseti-security/blob/master/google/cloud/forseti/services/model/importer/importer.py#L389). Right now the tests do not have G Suite integration and the G Suite API polling is not disabled, so these get recorded as warnings. I'd like to get the G Suite integration added to these tests at some point.

There are also model warnings when a [role does not have any permissions](https://github.com/forseti-security/forseti-security/blob/master/google/cloud/forseti/services/model/importer/importer.py#L1183). I'm not sure this should really be considered a warning because it appears to be a common thing, and doesn't necessarily mean that anything is broken.

There are a couple of other things that can cause a model warning, but these are the most common.

aimjwizards (approved):

PR #3825, opened 2020-09-30
Started updating the sqlite queries so testing works

scottsavarese:
Took the next few steps in moving this forward. Tests look like they work though I see unrelated (I think) errors.


google/cloud/forseti/services/inventory/cai_temporary_storage.py
```diff
@@ -326,7 +341,9 @@ def populate_cai_data(data, engine):
                 if row:
                     num_rows += 1
                     rows.append(row)
-                    rows_total_length += sum(len(v) for v in row.values())
+                    # Datetime doesn't have length so assume 21 bytes for now (length of the string)
```
scottsavarese, line 344: I don't like my own change here. The issue was that you can't call len() on a datetime object. The way around that is to covert it to string first: len(str(v)). Much easier this way.

scottsavarese: Changed my repo to fix this

gkowalski-google (approved):
This still needs some work to verify, and fix lint/unit tests, but will merge for now so that I can continue testing.

PR #3844, opened 2020-12-22
Enabling enforcer updates between denied <-> allowed FW rules

dragon1672:
Currently enforcer will fail to update a rule from 'allowed' <-> 'denied' with the following error

```
Error changing firewall rule asdf-some-fw-rule-name-asdf for project bap-iap: GCP API Error: unable to get bap-iap from GCP:
<HttpError 400 when requesting https://compute.googleapis.com/compute/v1/projects/bap-iap/global/firewalls/asdf-some-fw-rule-name-asdf?requestId=aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee&alt=json returned "Invalid value for field 'resource.allowed[0]': ''. Should not specify both allowed and denied in the same time.">
{
  "error": {
    "code": 400,
    "message": "Invalid value for field 'resource.allowed[0]': ''. Should not specify both allowed and denied in the same time.",
    "errors": [
      {
        "message": "Invalid value for field 'resource.allowed[0]': ''. Should not specify both allowed and denied in the same time.",
        "domain": "global",
        "reason": "invalid",
        "debugInfo": "java.lang.Exception\n\tat com.google.cloud.control.common.publicerrors.PublicErrorProtoUtils.newErrorBuilder(PublicErrorProtoUtils.java:1617)\n\tat com.google.cloud.control.common.publicerrors.PublicErrorProtoUtils.createInvalidFieldValue(PublicErrorProtoUtils.java:764)\n\tat com.google.cloud.control.common.publicerrors.PublicErrorProtoUtils.createInvalidFieldValue(PublicErrorProtoUtils.java:737)\n\tat com.google.cloud.cluster.manager.networking.services.firewall.MixerFirewallInputValidator.createAllowAndDenyInOneRequestError(MixerFirewallInputValidator.java:543)\n\tat com.google.cloud.cluster.manager.networking.services.firewall.MixerFirewallInputValidator.verifyRules(MixerFirewallInputValidator.java:406)\n\tat com.google.cloud.cluster.manager.networking.services.firewall.MixerFirewallInputValidator.verifyFirewallInput(MixerFirewallInputValidator.java:239)\n\tat com.google.cloud.cluster.manager.networking.services.firewall.FirewallsServiceUpdateAction$Handler.prepareMutationFlow(FirewallsServiceUpdateAction.java:297)\n\tat com.google.cloud.cluster.manager.networking.services.firewall.FirewallsServiceUpdateAction$Handler.prepareMutationFlow(FirewallsServiceUpdateAction.java:160)\n\tat com.google.cloud.control.frontend.action.ResourceMixerMutationActionExecutor$InnerHandler.prepareMutationFlowForProvider(ResourceMixerMutationActionExecutor.java:483)\n\tat com.google.cloud.control.frontend.action.ResourceMixerMutationActionExecutor$InnerHandler.prepareMutationFlow(ResourceMixerMutationActionExecutor.java:408)\n\tat com.google.cloud.control.frontend.action.ResourceMixerMutationActionExecutor$InnerHandler.prepareMutationFlow(ResourceMixerMutationActionExecutor.java:365)\n\tat com.google.cloud.control.frontend.action.MixerMutationActionExecutor$InnerHandler.runAttempt(MixerMutationActionExecutor.java:809)\n\tat com.google.cloud.control.frontend.action.MixerMutationActionExecutor$InnerHandler.runAttempt(MixerMutationActionExecutor.java:489)\n\tat com.google.cloud.cluster.metastore.RetryingMetastoreTransactionExecutor$1.runAttempt(RetryingMetastoreTransactionExecutor.java:75)\n\tat com.google.cloud.cluster.metastore.MetastoreRetryLoop.runHandler(MetastoreRetryLoop.java:454)\n\t...Stack trace is shortened.\n"
      }
    ]
  }
}
```

These rules will share the same name, so enforcer attempts to use 'patch' to update these types of rule changes. I tinkered around with API/gcloud/webUI calls and it appears that patch/update will both fail for this type of update. The only option is to delete and insert the rule to make this change.

I tried to put this distinction close to the compute API calls, as the rules classified as 'insert/update/delete' is exported and can be depended on by other systems. This changes doesn't modify the format of the exported results.

The '_rule_update_can_patch' is intended to directly service the needs of this change, but also indicate an area that could be expanded if we discover more update patterns that will need this replacement style.

Considered failure conditions
 - general failure after insert. Final error result will be handled like other API errors, if the delete succeeded then the next iteration/retry of enforcer will classify this rule as a insert, not a replace and will fall under the existing feature set
 - timeout failures
   + delete/insert times out, but actually succeeds. The request_id is re-used for retries so the delete should just skip.

- [X] I signed Google's [Contributor License Agreement](https://opensource.google.com/docs/cla/)
- [X] My PR has been functionally tested.
- [X] All of the [tests](https://forsetisecurity.org/docs/latest/develop/dev/testing.html) pass.
- [X] I have submitted a corresponding PR for documentation in the `forsetisecurity.org-dev branch` and included this on this PR (if applicable).
- [X] My documentation has been functionally tested and used (if applicable).
- [X] I have submitted a corresponding PR in the [Forseti Terraform module](https://github.com/forseti-security/terraform-google-forseti) (if applicable).

### DIFF
--- a/google/cloud/forseti/enforcer/gce_firewall_enforcer.py
+++ b/google/cloud/forseti/enforcer/gce_firewall_enforcer.py
@@ -189,9 +189,9 @@ def _rule_update_can_patch(rule_from, rule_to):
     if not rule_to or not rule_from:
         raise ValueError('from and to must both exist for checking replace vs patch.')
     if 'allowed' in rule_from and 'denied' in rule_to:
-        return False # Patch fails to update allowed -> denied
+        return False  # Patch fails to update allowed -> denied
     if 'denied' in rule_from and 'allowed' in rule_to:
-        return False # Patch fails to update denied -> allowed
+        return False  # Patch fails to update denied -> allowed
     return True
 
 
@@ -956,7 +956,7 @@ class FirewallEnforcer(object):
             raise FirewallEnforcementUpdateFailedError(
                 'Firewall enforcement failed while deleting rules for '
                 'project {}. The following errors were encountered: {}'
-                    .format(self.project, change_errors))
+              .format(self.project, change_errors))
         return len(successes)
 
     def _replace_rules(self, rules):
@@ -969,10 +969,8 @@ class FirewallEnforcer(object):
             raise FirewallEnforcementUpdateFailedError(
                 'Firewall enforcement failed while deleting rules for '
                 'project {}. The following errors were encountered: {}'
-                    .format(self.project, change_errors))
+              .format(self.project, change_errors))
         return len(successes)
-
-        return change_count
 
     def _apply_change(self, firewall_function, rules):
         """Modify the firewall using the passed in function and rules.
--- a/google/cloud/forseti/enforcer/gce_firewall_enforcer.py
+++ b/google/cloud/forseti/enforcer/gce_firewall_enforcer.py
@@ -185,6 +185,16 @@ def filter_rules_by_network(rules, network):
     return filtered_rules
 
 
+def _rule_update_can_patch(rule_from, rule_to):
+    if not rule_to or not rule_from:
+        raise ValueError('from and to must both exist for checking replace vs patch.')
+    if 'allow' in rule_from and 'allow' not in rule_to:
+        return False # Patch fails to update denied -> denied
+    if 'denied' in rule_from and 'denied' not in rule_to:
+        return False # Patch fails to update denied -> allowed
+    return False
+
+
 class FirewallRules(object):
     """A collection of validated firewall rules."""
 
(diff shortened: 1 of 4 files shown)


google/cloud/forseti/enforcer/gce_firewall_enforcer.py
```diff
@@ -923,16 +933,44 @@ def _update_rules(self, network):
...
-            self._updated_rules.extend(successes)
-            change_count += len(successes)
-            if failures:
-                raise FirewallEnforcementUpdateFailedError(
-                    'Firewall enforcement failed while deleting rules for '
-                    'project {}. The following errors were encountered: {}'
+            rules_to_patch = []
+            rules_to_replace = []
+            for rule in rules:
+                if _rule_update_can_patch(self.current_rules.rules[rule['name']], rule):
```
gkowalski-google, line 939: This is checking if `allowed` or `denied is in the rule name. Instead should it be checking if the rule has a non-empty `allowed` or `denied` list ([firewall api ref](https://cloud.google.com/compute/docs/reference/rest/v1/firewalls))?

google/cloud/forseti/enforcer/gce_firewall_enforcer.py
```diff
@@ -185,6 +185,16 @@ def filter_rules_by_network(rules, network):
     return filtered_rules
 
 
+def _rule_update_can_patch(rule_from, rule_to):
```
gkowalski-google, line 188: It would be nice to see unit tests for this.

google/cloud/forseti/enforcer/gce_firewall_enforcer.py
```diff
@@ -923,16 +933,44 @@ def _update_rules(self, network):
...
+            for rule in rules:
+                if _rule_update_can_patch(self.current_rules.rules[rule['name']], rule):
+                    rules_to_patch.append(rule)
+                else:
+                    rules_to_replace.append(rule)
+            if rules_to_patch:
+                change_count += self._patch_rules(rules_to_patch)
+            if rules_to_replace:
+                change_count += self._replace_rules(rules_to_replace)
+        return change_count
```
gkowalski-google, line 947: Perhaps you can add a new unit test that would test the replace logic is working? Maybe something like [this](https://github.com/forseti-security/forseti-security/blob/c2ae170a4b9ce19208c054ba51f97e154dfba247/tests/enforcer/gce_firewall_enforcer_test.py#L995)?

gkowalski-google (commented):
I think there is an issue with the logic that is deciding to patch/replace.

Have you tried playing with the [setting that controls if rules are deleted first](https://github.com/forseti-security/forseti-security/blob/c2ae170a4b9ce19208c054ba51f97e154dfba247/google/cloud/forseti/enforcer/gce_firewall_enforcer.py#L647)? Perhaps we can just default to deleting before inserting? If you think that would possibly cause some issues for existing users, then we could make it a configuration setting controlled with Terraform.

google/cloud/forseti/enforcer/gce_firewall_enforcer.py
dragon1672: captured in test_apply_firewall_denied_to_allowed

Since the replacement rules are captured as 'updates' since they share the same name. replace vs update is a deeper implementation decision, so test_apply_firewall_denied_to_allowed captures the full stack to confirm these rules update.

google/cloud/forseti/enforcer/gce_firewall_enforcer.py
dragon1672: captured in test_apply_firewall_denied_to_allowed

google/cloud/forseti/enforcer/gce_firewall_enforcer.py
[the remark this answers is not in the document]
dragon1672: In local testing I found the API doesn't set this field so checking for an empty array isn't needed.

dragon1672:
> Have you tried playing with the [setting that controls if rules are deleted first](https://github.com/forseti-security/forseti-security/blob/c2ae170a4b9ce19208c054ba51f97e154dfba247/google/cloud/forseti/enforcer/gce_firewall_enforcer.py#L647)? Perhaps we can just default to deleting before inserting? If you think that would possibly cause some issues for existing users, then we could make it a configuration setting controlled with Terraform.

This was my first solution. (It used a similar matching logic and just overrode that boolean to always be true to process deletes first).
I was opting for this this solution to prioritize maintaining network connection up-time. Deletes for a rule updating between allowed <-> denied doesn't impact network traffic if we delete it first. If there are other updates bundled in this change then their changes could result in downtime.

dragon1672:
> Please review the travis build to fix the lint.

Thanks for calling these out as linter errors. that took longer than I care to admit for me to find the right errors in those logs.

gkowalski-google (approved):
LGTM. Will merge once int tests complete.

PR #3872, opened 2021-08-07
Migrate lint and unit tests from Travis to Google Cloud Build

gkowalski-google:
Add cloudbuild config for lint tests. Remove Travis configuration, and some of the minor references to Travis in this repo. There is still more work to do to modify some of the scripts to remove Travis references; although for now it doesn't hurt.


build/cloudbuild-lint.yaml
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+timeout: 1200s
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  id: Build forseti test image
+  args: ['build', '--target', 'forseti-test', '-t', 'forseti-security/forseti-test:test', '.']
```
aimjwizards, line 18: will multiple concurrent tests have race condition on the image target?

gkowalski-google: These will all build separately and build an image of the branch that is being pushed. They will not interfere with each other.

gkowalski-google:
The failed trigger above is for the e2e tests which are currently not working, I have disabled those.

